PR #2, opened 2026-05-31
fix(coven-bin): resolve coven binary against user's interactive PATH (finder-launched .app)

BunsDev:
## Symptom

A `coven run codex --stream-json …` invocation from a Finder-launched
cave .app surfaced this clap error mid-chat:

\`\`\`
error: unexpected argument '--stream-json' found
tip: to pass '--stream-json' as a value, use '-- --stream-json'
Usage: coven run <HARNESS> <PROMPT>...
\`\`\`

Followed by:

> _The "codex" harness completed but produced no output._

## Root cause

When the bundled cave .app is launched from Finder/Spotlight, macOS
launchd hands it a **minimal PATH** (`/usr/bin:/bin:/usr/sbin:/sbin`).
Interactive-only shell setup (nvm, fnm, asdf, mise, etc.) doesn't run.

On this dev box that means:

- ✅ `~/.nvm/versions/node/v24.13.0/bin/coven` — fresh, symlink into the
  npm `@opencoven/cli` v0.0.30 (has `--stream-json`, `--continue`, …)
- ❌ `~/.cargo/bin/coven` — stale `cargo install` from May 20, version
  0.0.17, doesn't know `--stream-json`

If `~/.cargo/bin` is in launchd's PATH (e.g. via `/etc/paths.d/cargo`)
but `~/.nvm/...` is not, the .app spawn picks the stale Rust binary and
chokes on the `--stream-json` flag.

## Fix

New `src/lib/coven-bin.ts` exposes two helpers every cave spawn site
of `coven` must use:

- **`covenBin()`** — probes a priority-ordered list of well-known coven
  install locations (nvm/fnm node bins, `~/Library/pnpm`, `~/.bun/bin`,
  `/opt/homebrew/bin`, `/usr/local/bin`, `~/.local/bin`, then
  `~/.cargo/bin` last so a stale Rust install never wins by default).
  Returns the literal `"coven"` if nothing is found, so spawn still
  surfaces a clean "not found" error.

- **`covenSpawnEnv()`** — builds a deduped PATH from the candidate dirs
  plus the user's login-shell PATH (resolved via
  `$SHELL -ilc 'echo $PATH'`) so any further nested tooling (codex,
  claude, git, gh) the harness reaches for is also reachable.

Both are memoized for the lifetime of the server process.

## Wired into

- `src/app/api/chat/send/route.ts` — the original failure path
- `src/app/api/daemon/start/route.ts`
- `src/app/api/coven/exec/route.ts`

## Test plan

- [ ] On a dev box that has both an outdated `~/.cargo/bin/coven` and a
      fresh `~/.nvm/.../bin/coven`: launch cave from Finder → start a chat
      with a `codex`-bound familiar → verify the response streams instead
      of erroring with the `--stream-json` clap message.
- [ ] Run `pnpm tsc --noEmit` clean (verified locally).
- [ ] Confirm `~/.cargo/bin/coven` being absent on a fresh box still works
      (falls through to nvm-installed binary).
- [ ] Confirm a box with no `coven` at all surfaces a clean
      "command not found" rather than crashing.

### DIFF
--- a/src/app/api/chat/send/route.ts
+++ b/src/app/api/chat/send/route.ts
@@ -3,6 +3,7 @@ import { stat } from "node:fs/promises";
 import { homedir } from "node:os";
 import { stripAnsi } from "@/lib/ansi";
 import { bindingFor, loadConfig, recordSessionFamiliar } from "@/lib/cave-config";
+import { covenBin, covenSpawnEnv } from "@/lib/coven-bin";
 import {
   type ChatTurn,
   loadConversation,
@@ -182,9 +183,10 @@ export async function POST(req: Request) {
       const STDOUT_ERR_KEEP = 10;
       const ERR_LINE_RE = /\b(error|failed|denied|unauthori[sz]ed|invalid|refused|missing|not found|401|403|500)\b/i;
 
-      const child = spawn("coven", args, {
+      const child = spawn(covenBin(), args, {
         cwd,
         stdio: ["ignore", "pipe", "pipe"],
+        env: covenSpawnEnv(),
       });
 
       // Client-side abort (cancel button) → kill the subprocess.
--- a/src/app/api/coven/exec/route.ts
+++ b/src/app/api/coven/exec/route.ts
@@ -1,6 +1,7 @@
 import { NextResponse } from "next/server";
 import { spawn } from "node:child_process";
 import { stripAnsi } from "@/lib/ansi";
+import { covenBin, covenSpawnEnv } from "@/lib/coven-bin";
 
 export const dynamic = "force-dynamic";
 
@@ -29,7 +30,10 @@ export async function POST(req: Request) {
   const args = [body.command, ...(SUBCOMMAND_ARGS[body.command] ?? [])];
 
   return new Promise<Response>((resolve) => {
-    const child = spawn("coven", args, { stdio: ["ignore", "pipe", "pipe"] });
+    const child = spawn(covenBin(), args, {
+      stdio: ["ignore", "pipe", "pipe"],
+      env: covenSpawnEnv(),
+    });
     let out = "";
     let err = "";
     child.stdout.on("data", (d) => (out += d.toString()));
--- a/src/app/api/daemon/start/route.ts
+++ b/src/app/api/daemon/start/route.ts
@@ -1,13 +1,15 @@
 import { NextResponse } from "next/server";
 import { spawn } from "node:child_process";
+import { covenBin, covenSpawnEnv } from "@/lib/coven-bin";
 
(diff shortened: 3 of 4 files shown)


BunsDev:
Self-review (can't formally approve own PR):

**Diagnosis is exactly right.** Finder-launched .app + minimal launchd PATH + stale `~/.cargo/bin/coven` is a textbook macOS desktop-app failure mode, and the priority ordering (nvm/fnm → pnpm → bun → brew → /usr/local → ~/.local → cargo last) puts the canonical npm install ahead of every common stale source. Both spawn sites that mattered (`/api/chat/send`, `/api/coven/exec`, `/api/daemon/start`) get the fix in one pass.

**Worth tightening before merging or in a follow-up:**

1. **`execFileSync($SHELL -ilc 'echo $PATH')` blocks the Node event loop for up to 4s on first call.** For most users that's fine (one slow API request after boot) but heavyweight zsh setups can stretch closer to the timeout. Two safer patterns: (a) trigger the resolution at module-load with a top-level `await execFileAsync(...)` so the slowness happens once during sidecar init, before the first request, or (b) keep sync but cache to disk (`~/Library/Caches/CovenCave/coven-path`) so restarts don't pay the cost again.

2. **`-ilc` can echo rc-banner output before `$PATH`.** Some users have `echo "welcome back"`-style lines in `.zshrc` that aren't gated by `$-` interactivity checks. The current `out.trim()` will swallow that into a busted PATH string. Cheap fix: wrap with delimiters — `$SHELL -ilc 'printf "__COVEN_PATH_START__%s__COVEN_PATH_END__" "$PATH"'`, then regex-extract.

3. **`statSync` follows symlinks**, so `st.isSymbolicLink()` is unreachable in the `coven-bin.ts:104` check. The `isFile()` branch already handles dereferenced symlinks correctly — so the `|| st.isSymbolicLink()` clause is dead code. Either drop it or switch to `lstatSync` if you specifically want to allow broken symlinks (probably not — broken symlinks should keep searching).

4. **fnm's actual default install dir is `~/.local/share/fnm/node-versions/`** when `$XDG_DATA_HOME` is unset; the `~/.fnm/node-versions/` path you're probing is the older layout. Worth probing both.

5. **No version arbitration.** If both nvm-coven and brew-coven exist and both are recent enough, the first wins. That's fine for the bug at hand (any non-stale install is fine), but if you want to be paranoid, a `coven --version` probe + `semver.gt` comparison could be added later.

The fix is correct and the priority ordering is sound. None of the above blocks merge — call them notes for a follow-up if the slow-first-request behavior shows up in telemetry.

Co-authored-by: OpenCoven <noreply@opencoven.ai>

PR #5, opened 2026-05-31
feat(chat): simplify chrome, widen transcript, surface tool use + reasoning

BunsDev:
## Summary

UX pass on the new Cave chat surface — strips chrome you said you didn't want, widens the transcript, and surfaces what the agent is actually doing.

> **Stacks on [#4](https://github.com/OpenCoven/coven-cave/pull/4) (iconify).** Branch is cut from `feat/iconify-icons`. Rebase onto `main` after #4 lands, or merge #4 first.

### Removed

| What | Where |
|---|---|
| Per-chat header row: back button, familiar name, harness, project name, daemon status, live/streaming pill | `ChatView` |
| Two-line footer under composer: `↵ send · ⇧↵ newline · / commands · ⌘K palette` + `harness · model` | `ChatView` |
| Bottom workspace footer: `CovenCave · v0` + `mode · Chats` | `Workspace` |
| Secondary nav row with mode tabs + `⌘K palette · ⌘B rail · ⇧⌘B inspector` shortcut text | `Workspace` |
| Header text: daemon status, pid, live/total session counts, `N reply` badge, start/refresh daemon button, setup button | `DaemonBar` |

### Migrated

- **Mode tabs** (Chats / Board / Inbox / Plugins) moved into the top header row, immediately right of the `CovenCave` brand. Inbox keeps the `· N` count when there are fired/response-needed items.
- **Daemon status polling** stays in `DaemonBar` but silent — `onRunningChange` still fires for downstream consumers.

### Added

- **Timestamps** under each message, `HH:MM` localized.
- **Reasoning collapsible** — any `<thinking>…</thinking>` or `<reasoning>…</reasoning>` block emitted by Claude/Codex harnesses is split out and rendered as a collapsed-by-default disclosure with a caret + line count. The visible reply text stays clean.
- **Tool use rendering** —
  - New SSE event kind on `/api/chat/send`: `{ kind: "tool_use", id, name, input?, output?, status, durationMs? }`.
  - Backend now parses `hook: tool_use|pre_tool_use|post_tool_use <Name> <rest>` lines out of the harness stdout (previously discarded by `AssistantFilter`) and emits structured events with timing.
  - Frontend renders each tool as a row above the assistant text: caret + 🔧 wrench icon + tool name + status dot (amber pulse / emerald / rose) + duration. Body collapses to show input + output when available; rows with no body are non-clickable.

### Widened

The transcript no longer constrains to `max-w-3xl`. It fills the chat panel width with consistent padding. Same for the composer.

## Test plan

- [x] `tsc --noEmit` clean across the four files.
- [ ] `next dev` — header shows only Brand + Tabs + 🔔; chat transcript fills panel width; messages show timestamps; long replies stay clean while reasoning collapses.
- [ ] Run a `coven` prompt that does tool use — verify the wrench row appears, stays amber-pulse during the call, turns emerald + shows duration on completion, and that the body expands to show input/output.
- [ ] Run a prompt with `<thinking>...</thinking>` content — verify it folds into the reasoning collapsible.
- [ ] Switch modes via the new header tabs.

## Notes

- `ChatView` still accepts `onBack` and `daemonRunning` props for backwards compat with `ChatRouter` but no longer reads them. Can be removed in a follow-up alongside the router signature.
- Tool-use parser is best-effort against the hook line format codex/claude currently emit. When the harnesses gain structured JSON tool events, the backend should switch to those instead of regex on hook lines.

### DIFF
--- a/src/app/api/chat/send/route.ts
+++ b/src/app/api/chat/send/route.ts
@@ -23,10 +23,25 @@ type StreamEvent =
   | { kind: "session"; sessionId: string }
   | { kind: "user"; text: string }
   | { kind: "assistant_chunk"; text: string }
+  | {
+      kind: "tool_use";
+      id?: string;
+      name: string;
+      input?: string;
+      output?: string;
+      status?: "running" | "ok" | "error";
+      durationMs?: number;
+    }
   | { kind: "done"; durationMs?: number; isError?: boolean; sessionId?: string }
   | { kind: "error"; message: string; code?: string };
 
 const HOOK_LINE_RE = /^hook:\s+/;
+// Hook-line shapes emitted by codex/claude harnesses while a tool runs.
+// Examples:
+//   hook: tool_use Bash {...}
+//   hook: pre_tool_use Edit { ... }
+//   hook: post_tool_use Bash {... exitCode: 0 ...}
+const TOOL_HOOK_RE = /^hook:\s+(?:pre_tool_use|post_tool_use|tool_use)\s+(\S+)(?:\s+(.*))?$/;
 const BANNER_LINE_RE = /^(?:--------|workdir:|model:|provider:|approval:|sandbox:|reasoning|session id:|tokens used|\d[\d,]*\s*$)/;
 const CODEX_START_LINE = "codex";
 const CLAUDE_ASSISTANT_RE = /^claude(?:\s+code)?$/i;
@@ -171,6 +186,18 @@ export async function POST(req: Request) {
       let assistantText = "";
       let jsonBuf = "";
       let result: { duration_ms?: number; is_error?: boolean } = {};
+      // Per-tool start times so post_tool_use can compute durationMs and
+      // associate output with the matching pre_tool_use event.
+      const toolStartTimes = new Map<string, { startedAt: number; id: string }>();
+      let toolSeq = 0;
+      const toolIdFor = (name: string): string => {
+        const existing = toolStartTimes.get(name);
+        if (existing) return existing.id;
+        toolSeq += 1;
+        const id = `tool-${toolSeq}-${name}`;
+        toolStartTimes.set(name, { startedAt: Date.now(), id });
+        return id;
+      };
       // Keep stderr off the assistant stream — surface it only on failure
       // or empty-success so users don't see raw 401 traces mid-bubble.
       const stderrTail: string[] = [];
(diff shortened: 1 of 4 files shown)


BunsDev:
Self-review (can't formally approve own PR):

Walked the diff end-to-end + `pnpm build` clean against `main` (with #4 already merged). Solid architectural move — the tool-use rendering and reasoning collapsible are real UX wins, and shrinking the chrome reads exactly like what you described wanting. Approving in spirit.

**Notes worth flagging, none blocking:**

1. **`toolIdFor` keys by tool name only.** Two parallel `Bash` calls would collide on id, and a post-event for one would clobber the other's tracking state. Serial harnesses (codex/claude today) make this fine, but when concurrent tool calls land, switch to keying by harness-supplied id and only fall back to seq when the hook line omits one.

2. **`isError` regex (`/error|fail|denied|exit\s*[1-9]/i`) misclassifies legitimate output.** A `grep -i error logs/` that succeeds will turn the dot rose. Acknowledged by the PR notes as best-effort against hook-line shape — fine for v0, replace when harnesses emit structured JSON tool events.

3. **`splitReasoning` runs on every `TurnRow` render** including each streamed chunk. For long assistant replies the regex re-scans the whole string on every state tick. `useMemo` keyed on `turn.text` would make streaming feel smoother on long replies. Trivial follow-up.

4. **Tool input/output `<pre>` has no max-height.** A long `Bash` output (build log, recursive ls) can blow up transcript height when expanded. Worth capping with `max-h-96 overflow-y-auto` on the body block.

5. **Daemon "start daemon" button is gone from `DaemonBar`.** PR description says start/stop now lives in `/doctor` or CLI — fine as an opinionated call, but the onboarding overlay is still the only first-launch path. Worth confirming the onboarding auto-opens reliably when daemon is down, otherwise first-launch users with a broken daemon need to know `/doctor` exists.

6. **`onBack` + `daemonRunning` props on `ChatView` are now unread** (acknowledged in PR notes). Fine to defer the prop-cleanup follow-up — just don't let it slide more than a sprint.

7. **`fmtTime` uses `toLocaleTimeString`** which is locale-dependent. Since `ChatView` is `"use client"`, no SSR mismatch — but if you ever pre-render a transcript snapshot on the server, this becomes a hydration hazard.

8. **Design-system compliance pass (OpenCoven Design System v1.0.0):** ✓ flat surfaces, no shadows added, accent stays on `purple-300` for the wrench icon and `purple-500/600` elsewhere, status colors use semantic `amber`/`emerald`/`rose`, focus states preserved on buttons. The transcript widening matches your "no wasted horizontal whitespace" preference. ✓

**Verified locally:** `pnpm build` clean against `feat/chat-experience-overhaul` after #4 merged to main.

Co-authored-by: OpenCoven <noreply@opencoven.ai>

PR #53, opened 2026-06-01
ci(release): build + notarize macOS DMG in CI

BunsDev:
## Summary
- Adds \`macos-latest\` to the release matrix so tag-push produces a signed + notarized DMG automatically. Reuses \`tauri-action\`'s built-in signing path via the standard \`APPLE_*\` env vars.
- Hard verification step: runs \`spctl\` against the produced DMG and fails the job if Gatekeeper does not accept it as a *"Notarized Developer ID"* build. This turns notarization from a convention (which broke for v0.0.13–v0.0.15) into a build-time invariant.
- \`scripts/release.sh\` stays around for local iteration but is no longer the canonical release path.

## Required secrets (need to be added in OpenCoven/coven-cave settings before this can run)
- \`APPLE_CERTIFICATE\` — base64 of the .p12 export of Developer ID Application cert
- \`APPLE_CERTIFICATE_PASSWORD\` — password set when exporting the .p12
- \`APPLE_SIGNING_IDENTITY\` — the identity name or SHA1 (e.g. \`EE732DF3F48D7535561AF54D3FFFC4B44DAF3E7F\`)
- \`APPLE_API_ISSUER\` — App Store Connect API issuer UUID
- \`APPLE_API_KEY\` — API key id (e.g. \`3822D8Z5XFI0\`)
- \`APPLE_API_KEY_BASE64\` — base64 of \`AuthKey_*.p8\`

## Test plan
- [x] YAML syntax validates (\`python3 -c \"import yaml; yaml.safe_load(open('.github/workflows/release.yml'))\"\`)
- [ ] Add the six secrets above to the repo
- [ ] Cut a test tag (e.g. \`v0.0.16-test\`) and confirm the macOS job produces a notarized DMG and the verification step passes
- [ ] Delete the test tag + release once green

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,22 +1,36 @@
 name: Release
 
-# Builds CovenCave Linux (AppImage) and Windows (unsigned MSI) artifacts
-# and uploads them to the GitHub Release matching the pushed tag.
+# Builds all three Cave artifacts and uploads them to the GitHub Release
+# matching the pushed tag:
 #
-# macOS is intentionally NOT built here — that flow stays local via
-# scripts/release.sh, which has the Apple Developer ID + notarization
-# credentials and produces a signed/notarized DMG. The local DMG is
-# uploaded into the same GitHub Release manually (or via `gh release
-# upload <tag> release/CovenCave-v<version>.dmg`).
+#   - macOS (DMG) — signed with the Soul Protocol LLC Developer ID and
+#     notarized via the App Store Connect API. The workflow refuses to
+#     upload an un-notarized DMG: a verification step runs `spctl` after
+#     the build and fails the job if Gatekeeper does not accept it.
+#   - Linux (AppImage) — unsigned.
+#   - Windows (MSI) — unsigned (users see SmartScreen on first run; we'll
+#     wire WINDOWS_CERTIFICATE secrets when we buy a code-signing cert).
 #
-# Trigger: push a tag matching `v*` (e.g. `git tag -s v0.0.12 && git push
-# origin v0.0.12`). The release.sh script already pushes a signed tag,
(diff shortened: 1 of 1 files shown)


BunsDev:
## Test-tag dry run: v0.0.16-rc1 (deleted)

All three platforms (mac/linux/win) failed at the same step in `Build with tauri-action`:

\`\`\`
resource path \`resources/server/.next/node_modules/shiki-1476e3c95a68a2b2\` doesn't exist
\`\`\`

This is **not** a signing/notarization issue — the macOS leg never got to the verification step. It's a Next.js standalone bundling quirk:

- The standalone output puts pnpm-style symlinks under `.next/node_modules/` (e.g. `shiki-<hash> -> ../../node_modules/.pnpm/shiki@4.1.0/node_modules/shiki`).
- `scripts/sidecar-bundle.sh` has a dangling-symlink resolver (lines 74-99) that's supposed to copy real directories over those symlinks before Tauri picks them up.
- On CI that resolver's guard `[ -d \"$DEST/.next/node_modules\" ]` evaluated false — the *"==> resolving dangling pnpm symlinks"* echo never appeared in the log.
- Net effect: the bundle ships with dangling symlinks → Tauri's build script aborts before the binary is produced.

**This PR's workflow is still correct in concept.** It just can't be greenlit until the sidecar bug is sorted. Two reasonable paths:

1. **Fix `sidecar-bundle.sh` first** so the bundle is CI-clean (likely a small change: handle the case where `.next/node_modules` exists but its sub-symlinks are dangling; or pre-resolve all symlinks via a separate find-and-rewrite pass).
2. Or change the Tauri resource glob to exclude `.next/node_modules/*` and rely on the top-level `node_modules` graft already done by the sidecar script.

Either fix is a separate PR. Leaving #53 open as the destination — the workflow change is good as-is.

PR #55, opened 2026-06-01
[codex] Fix Windows packaged launch with Node 24

qkal:
## Summary

This fixes the Windows packaged-app crash that happens when launching CovenCave from the installed Start Menu/Desktop shortcut.

The shortcut itself is valid and launches `C:\Program Files\CovenCave\app.exe`, but Tauri resolves the bundled Next.js server path with Windows' extended-length prefix and hands Node an entrypoint like:

```text
\\?\C:\Program Files\CovenCave\resources\server\server.js
```

Node.js v24.14.0 fails while resolving that main module and collapses the path to `C:`, producing:

```text
Error: EISDIR: illegal operation on a directory, lstat 'C:'
```

The app then times out waiting for the sidecar port and exits before the GUI can open.

## What changed

- Added a small `node_arg_path()` helper in the Tauri launcher.
- On Windows, it strips `\\?\` from local paths and converts `\\?\UNC\...` back to normal UNC form before passing paths to Node.
- On non-Windows platforms, it preserves paths unchanged.
- Applied that normalization only at the process boundary for:
  - the Node entrypoint argument
  - the sidecar working directory
- Added focused Windows regression coverage for both extended and normal paths.

## Why

Node can run the packaged server when it receives a normal `C:\Program Files\...\server.js` path, but fails before userland code runs when it receives the extended `\\?\C:\...` path. Normalizing at the launcher boundary keeps the existing Tauri resource lookup intact while avoiding Node's Windows entrypoint resolution bug.

## Validation

- `cargo test node_arg_path --locked` passes: 2 tests passed.
- `cargo check --locked` passes.
- `git diff --check HEAD~1..HEAD` passes.
- Manually reproduced the failing Node behavior with `\\?\C:\...\server.js` and confirmed the same packaged server starts successfully with a normal `C:\Program Files\...\server.js` path, returning HTTP 200.

## Notes

Full local MSI packaging was not rerun in this checkout because the existing `beforeBuildCommand` invokes `bash scripts/sidecar-bundle.sh`, and this Windows environment only has the WSL `bash.exe` stub without a working `/bin/bash`.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -1,5 +1,5 @@
 use std::net::TcpListener;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use std::process::{Child, Command, Stdio};
 use std::sync::Mutex;
 use std::thread;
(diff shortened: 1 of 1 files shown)


BunsDev:
Superseded by #59 which includes the UNC regression test.

PR #63, opened 2026-06-02
fix(chat): boot harness in familiar workspace so familiars know who they are

BunsDev:
## Problem

Familiars like Charm were introducing themselves as "Codex" (or "Claude Code") during chat because the harness was launched with no identity context — it booted in `~/` with no `AGENTS.md`, `SOUL.md`, or `IDENTITY.md`.

## Fix

When no explicit `projectRoot` is set, `/api/chat/send` now resolves `~/.openclaw/workspace/<familiarId>` and passes it as `--cwd` to `coven run`. This means the harness boots in the familiar's own workspace and reads their persona files automatically.

**Before:** `coven run codex --stream-json -- <prompt>`  
**After:** `coven run codex --stream-json --cwd ~/.openclaw/workspace/charm -- <prompt>`

Falls back gracefully — if the workspace dir doesn't exist, `--cwd` is omitted and behavior is unchanged.

## Why this works

OpenClaw familiars each have a workspace at `~/.openclaw/workspace/<id>/` containing:
- `AGENTS.md` — behavior and memory guidance
- `SOUL.md` — persona and tone  
- `IDENTITY.md` — name, creature, vibe

Codex and Claude Code both read `AGENTS.md` in their working directory on boot. Booting in the familiar workspace means they adopt the familiar's identity automatically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "coven-cave",
-  "version": "0.0.22",
+  "version": "0.0.23",
   "private": true,
   "scripts": {
     "dev": "next dev",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "coven-cave",
-  "version": "0.0.23",
+  "version": "0.0.24",
   "private": true,
   "scripts": {
     "dev": "next dev",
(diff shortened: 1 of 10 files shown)


BunsDev:
Already landed on main via separate squash commits — branch is now wholly redundant and would conflict with newer work (#64 sidebar accordion). Closing.

PR #97, opened 2026-06-03
feat(chat): Task 5 — system message terminal chrome

BunsDev:
- Replaces plain monospace box with cave-code-wrap-style terminal panel
- Header: $ sigil + optional command label (or dim 'system' fallback)
- Scrollable pre body, max-height 320px
- label prop added to MessageBubbleProps for future slash-command labelling

Part of Coven Cave chat presentation overhaul.


BunsDev:
System chrome included in #96 merge

PR #98, opened 2026-06-03
feat(chat): Task 4 — timestamp grouping + hover-reveal

BunsDev:
- Timestamps suppressed when consecutive same-role turns are <10 min apart
- Revealed on 600ms hover via CSS opacity transition
- Always shown when gap ≥10 min
- showTimestamp prop added to MessageBubble

Part of Coven Cave chat presentation overhaul.


BunsDev:
Timestamp + avatar changes included in #96 merge (resolved manually)

PR #120, opened 2026-06-06
fix(release): sign macOS DMG before notarization

BunsDev:
## Summary
- sign the generated macOS DMG with the Developer ID Application identity before notarization
- verify the DMG code signature before submit
- make release.sh fail on the same Gatekeeper DMG acceptance invariant as the workflow

## Root cause
The fallback release lane now gets notarization accepted and staples successfully, but the workflow gate still rejects the DMG with `source=no usable signature`. The .app is signed/notarized; the disk image envelope also needs a Developer ID signature before submission.

## Verification
- bash -n scripts/release.sh
- git diff --check

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -157,9 +157,8 @@ xcrun stapler staple "$DMG_PATH"
 xcrun stapler staple "$APP_PATH"
 
 echo "==> Verifying Gatekeeper acceptance"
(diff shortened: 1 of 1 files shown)


BunsDev:
Closing this duplicate follow-up: main already landed the DMG-container signing fix as b183f0e and the v0.0.38 release run passed with the macOS DMG uploaded.

PR #323, opened 2026-06-09
feat(voice): voice chat — WebRTC realtime overlay + OpenAI Realtime adapter

BunsDev:
## Summary

- Ship full-duplex voice calls between the user and any familiar via OpenAI Realtime (browser-direct WebRTC + server-minted ephemeral token); Gemini Live stubbed for v1.1 behind the same `VoiceProvider` adapter.
- Server resolves `OPENAI_API_KEY` from the existing vault (1Password `op://` ref), mints an ephemeral `client_secret`, never proxies audio. Long-lived key never leaves the server.
- Transcripts auto-append to the familiar's chat thread (`origin: "voice"`, grouped by `voiceCallId` under a "Voice call · MM:SS" header), so future text turns inherit voice context. Identity + last 12 conversation turns hydrate every call's system instructions.

## What's in the diff

- **New library** `src/lib/voice/` — `types.ts`, `registry.ts`, `openai-realtime.ts` (server `mintSession` + client WebRTC adapter), `gemini-live.ts` (stub), `hydrate-instructions.ts`, `append-voice-turn.ts`.
- **New routes** `POST /api/voice/session` (full error matrix: invalid_session, voice_not_configured, unknown_provider, vault_key_unresolved, provider_mint_failed) and `POST /api/voice/transcript` (writer-only, never invokes harness). Both registered in `api-contracts.test.ts`.
- **New UI** `VoiceCallButton` in chat-header (gated on `familiar.voiceProvider` + active `sessionId`), `VoiceCallOverlay` (pure reducer split into its own file so it's testable under `node --experimental-strip-types` without dragging in `@/` imports), and "Voice call · MM:SS" turn grouping in `chat-view.tsx`.
- **Familiar Studio → Brain** gains voice provider/model/voice pickers parallel to harness/model.
- Two vault entries (`OPENAI_API_KEY`, `GOOGLE_API_KEY`) documented with empty refs in `vault.yaml` — user supplies their own `op://`.

17 signed commits. 16 new unit tests across 8 files. Harness chat pipeline untouched.

## Correctness fixes applied during build

- `RTCPeerConnection` closed on SDP failure (was leaking on retry).
- WebRTC `disconnected` no longer fires `onDisconnect` (transient state; only `failed` is terminal — prevents teardown on minor network blips).
- `callId` captured as a local const inside the connecting branch (replaces `state.callId!` non-null assertion).
- `cleanup()` runs on the `error` state path so mic tracks + WebRTC session don't leak across RETRY.
- Call duration ticks every second while `live` (rendered duration was previously frozen at the moment of CONNECTED).

## Test plan

- [ ] `pnpm test:app` + `pnpm test:api` green (66 route contracts + 16 new voice tests)
- [ ] `pnpm typecheck` clean
- [ ] Add an `op://` ref for `OPENAI_API_KEY` in Vault settings (or `export OPENAI_API_KEY=...`)
- [ ] Familiar Studio → Brain: set Voice Provider = OpenAI on one familiar; defaults fill (`gpt-4o-realtime-preview` / `alloy`)
- [ ] Open that familiar's chat, click the phone icon in the linear header, grant mic permission
- [ ] Speak; hear the familiar's voice reply; click End call
- [ ] Confirm the transcript appears in the chat thread under a "Voice call · MM:SS" group header
- [ ] Confirm subsequent text turns can reference what was said on the call (assistant knows the voice transcript)

**Caveat:** SDP exchange / WebRTC handshake is intentionally NOT unit-tested in v1 (provider-specific, mocked at the adapter boundary). The manual smoke above is the only end-to-end signal. The implementer also flagged that OpenAI Realtime event type names have drifted historically (`response.audio_transcript.done` is current at time of writing); easy to patch in `openai-realtime.ts` if smoke reveals drift.

## Non-goals (deferred to v1.1+)

Gemini Live implementation, audio recording, dedicated Calls view, tool/function calling over voice, transcript retry queue, multi-call concurrency, memory retrieval into hydration.


BunsDev:
@copilot resolve the merge conflicts in this pull request

PR #339, opened 2026-06-09
fix(home, shell): mobile polish, symmetric panels, equidistant home

BunsDev:
## Summary

Three coordinated layout changes, all motivated by the user's recent feedback on mobile + centering:

### Mobile polish (phones 375–428 portrait)
- Top bar collapses the search button to **icon-only** at `< 640px` so the notification bell and settings cog stop getting clipped (was caused by `.top-bar__search { min-width: 420px }`).
- Suggestion chips cap at `min(100%, 360px)` and the inner `<span>` ellipsis-truncates so long titles like *"Continue: Task context: Title: Create Design Palettes…"* no longer spill past the viewport.
- `.hc-keyboard-hint` hides on `@media (pointer: coarse)` — no physical keyboard, no need for the desktop shortcut legend.
- `.home-composer-root` anchors to `flex-start` with `padding-top: clamp(72px, 18vh, 160px)` on phones so the title sits near the upper third instead of stranded mid-screen.
- Nav / list / agent panels now hide on phones by targeting **both** the className inner wrapper **and** `[data-panel="true"]#id` outer wrapper. The outer is what react-resizable-panels controls via inline `flex: N 1 0px` — without targeting it, the outer kept its 17% allotment and pushed the detail panel ~64px right of the viewport.

### Symmetric side panels
- Both `.shell-nav-panel` and `.shell-agent-panel` use `minSize="16%" maxSize="28%"`. Was asymmetric before (nav 14/25, agent 25/38), which guaranteed the layout could never be balanced even with both open.

### Equidistant Home centering (replaces PR #333 viewport-shift)
- Drop the `--shell-nav-px` / `--shell-agent-px` / `--hc-asymmetry` / `translateX(clamp(...))` machinery and the matching `.shell-detail-panel:has(...) { overflow: visible }` override.
- With symmetric panel min/max, simply centering the composer in `.shell-detail` gives **equal** gaps to both side-panel edges — which is what the user actually asked for.

Measured gaps (cardLeft − navRight  ===  agentLeft − cardRight):
| Viewport | leftGap | rightGap |
|---:|---:|---:|
| 1024 | 32 | 32 |
| 1440 | 148 | 148 |
| 1920 | 348 | 348 |

## Test plan

- [x] `node --experimental-strip-types src/components/home-composer-centering.test.ts` — now asserts symmetric panels, no transform, the four mobile CSS hooks, and the data-panel outer-wrapper hide.
- [x] `home-composer.test.ts` + `home-composer-polish.test.ts` still pass.
- [x] Visual: captured at 375 / 390 / 428 / 412 phone widths AND 1024 / 1440 / 1920 desktop — composer centered on viewport at mobile, equidistant from both side panels on desktop.

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -314,19 +314,6 @@ body {
   display: flex;
 }
 
-/* Home composer needs to visually center on the viewport, which means its
- * .home-composer-root translates left to compensate for the nav panel.
- * Allow the detail panel to overflow visible only while it contains a Home
- * composer; react-resizable-panels sets `overflow: auto` inline, so the
- * override needs `!important`. The composer's own .home-composer-root
- * keeps its own vertical scroll. */
-.shell-detail-panel:has(> .shell-detail > .cave-mode-fade > .home-composer-root) {
-  overflow: visible !important;
-}
-.shell-detail-panel:has(> .shell-detail > .cave-mode-fade > .home-composer-root) > .shell-detail {
-  overflow: visible;
-}
-
 /* Inner panel border — subtle dividing line on the right edge of side panels */
 .shell-nav-panel { border-right: 1px solid var(--border-hairline); }
 .shell-agent-panel { border-left: 1px solid var(--border-hairline); }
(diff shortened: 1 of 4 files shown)


BunsDev:
Closing — main has had 4 commits land in parallel (#334, #336, #337, #338) covering the bulk of this PR's scope (mobile readiness pass, wider 1200px composer, symmetric 14-25% panels). Opening a smaller follow-up with only the gaps that remain: suggestion-chip ellipsis, .hc-keyboard-hint hide on (pointer: coarse), and the [data-panel] outer-wrapper hide for the 64px ghost-margin.

PR #467, opened 2026-06-12
fix: unblock main CI — restore @xyflow/react + projects client/server boundary

BunsDev:
`main` CI has been red. Two independent breakages from the recent Projects/Workflows churn, both fixed here:

### 1. Missing dependency (was PR #464, folded in)
`@xyflow/react@12.11.0` was dropped from `package.json` (commit f811f20) while `workflow-canvas.tsx` still imports it. → `ERR_PNPM_OUTDATED_LOCKFILE` + `TS2307`. **Fix:** restore the one dep line.

### 2. node:fs in the client bundle
`f811f20` intended a client/server split (`cave-projects-types.ts`) so `"use client"` components import types + pure helpers, but the split file was **lost in a merge** and `projects-view.tsx` imported the *value* `normalizeProjectRoot` straight from `cave-projects.ts` (which imports `node:fs/promises`). → Turbopack: *"chunk item errored … does not support external modules (request: node:fs/promises)"* on `workspace.tsx`. **Fix:** recreate `cave-projects-types.ts`, re-export from `cave-projects.ts` for server importers, point `projects-view.tsx` at the safe module.

### Verification
`pnpm build` compiles clean and generates all 18 static pages locally; `pnpm typecheck` passes.

Supersedes #464 (its one-line fix is included here).

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "@xterm/addon-fit": "^0.11.0",
     "@xterm/addon-web-links": "^0.12.0",
     "@xterm/xterm": "^6.0.0",
+    "@xyflow/react": "12.11.0",
     "next": "16.2.6",
     "node-pty": "^1.1.0",
     "qrcode": "^1.5.4",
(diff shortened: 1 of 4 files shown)


BunsDev:
Redundant — main has self-healed. Both fixes already landed independently: @xyflow/react is restored in package.json, and another session shipped the equivalent cave-projects-types.ts client/server split. main CI is green (last 3 runs pass). Closing; no action needed.

PR #566, opened 2026-06-13
Potential fix for code scanning alert no. 45: Uncontrolled data used in path expression

BunsDev:
Potential fix for [https://github.com/OpenCoven/coven-cave/security/code-scanning/45](https://github.com/OpenCoven/coven-cave/security/code-scanning/45)

Use a canonical, root-contained path calculation in `saveWorkflowLayout` instead of `path.join(dir, file)`.

Best fix (without changing behavior): in `src/lib/workflow-source.ts`, inside `saveWorkflowLayout`, replace manual path joining with `layoutFilePath(id)` and reject if it returns `null`. This reuses the existing safe resolver that:
- resolves against an absolute root,
- normalizes the resulting path,
- verifies the target remains inside the root via `path.relative`.

This keeps current slug validation and storage resolution behavior while making the write/chmod sink use a defensively validated absolute path.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/src/lib/workflow-source.ts
+++ b/src/lib/workflow-source.ts
@@ -562,14 +562,19 @@ async function storageForWorkflowId(id: string): Promise<WorkflowStorageScope> {
 }
 
 /** Absolute path for a layout sidecar, constrained to its workflow storage dir. */
-async function layoutFilePath(id: string): Promise<string | null> {
+function layoutFilePathForStorage(id: string, storage: WorkflowStorageScope): string | null {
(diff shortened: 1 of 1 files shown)


BunsDev:
Closing as stale: code-scanning alert #45 is already `state: fixed` on main — `saveWorkflowLayout` already has the inline path.resolve + path.relative + '..' barrier (lines 613-617). This autofix is redundant and conflicting.

Issue #15, opened 2026-05-31
Health strip in daemon bar — surface silent failures


BunsDev:
Parent: #13

## Why

Val has already lost work to silent OpenClaw failures. Multica makes daemon/runtime health *visible* (daemons poll 3s, heartbeat 15s, marked offline at 45s). We need the same discipline.

## Scope

Expand the existing daemon bar into a **health strip** showing the state of each tracked surface as colored dots.

### Tracked surfaces
- \`coven\` daemon (local)
- Each familiar's session process (Sage, Cody, Kitty, Charm, Echo, Astra, Nova)
- OpenClaw gateway
- External integrations (Telegram bridge, etc.)

### States
- 🟢 green — healthy, recent heartbeat
- 🟡 yellow — degraded (slow, partial)
- 🔴 red — offline / failed
- ⚪ grey — not configured / disabled

### Interaction
- Hover/click a dot → popover with last heartbeat timestamp + recent errors (last 5)
- The accent color \`#9A8ECD\` may be used on the *healthy* dot as a presence pulse animation — Multica uses pure grayscale here; we use the ritual.

## Acceptance
- [ ] Health strip visible in daemon bar at all times
- [ ] Each tracked surface emits a heartbeat at a documented interval
- [ ] Stale heartbeat → state degrades within one configured window
- [ ] Click → popover with last-seen + recent errors
- [ ] No console errors when a surface is missing — degrades to grey

## Out of scope
- Webhook triggers for external alerting (separate concern)
- Historical health charts (separate issue)

PR #656, opened 2026-06-15
fix(library): resolve docs from trusted allowlist + anchor path guard

BunsDev:
Two-commit fix for library doc path traversal guard:\n- Anchors path resolution to prevent escape\n- Resolves library docs through a trusted allowlist\n\nAddresses code scanning alert #15/#16 more robustly.


BunsDev:
Superseded by #638, merged to main as 1daa3bb. Leaving this open would replay the older branch based at 42923ca and drop the #654 capability preview changes, so closing this stale duplicate.

PR #704, opened 2026-06-15
fix(terminal): harden pty websocket access

BunsDev:
### Motivation
- Close a critical unauthenticated remote command execution path exposed by the custom `/api/pty-ws` WebSocket upgrade handler which previously allowed any upgrade when `COVEN_CAVE_ACCESS_TOKEN` was unset.
- Preserve PTY functionality for legitimate local development and sidecar-authenticated clients while preventing network- or cross-origin attackers from reaching a spawned shell.

### Description
- Replace permissive token logic with a conservative gate that requires either a valid sidecar access token or a local loopback upgrade, and also enforces browser `Origin`/`Referer` checks by reusing the existing proxy source-origin helpers (`isAllowedRequestSource`/`isLoopbackHost`).
- Use the shared access cookie constant (`ACCESS_TOKEN_COOKIE`) instead of a hardcoded cookie name and keep bearer-token support for non-cookie clients.
- Restrict unauthenticated upgrades to genuine loopback connections by validating both the Host header and the peer address, and compute an expected origin for `Origin`/`Referer` checks.
- Default the server bind host to loopback via `HOST ?? "127.0.0.1"` (instead of `HOSTNAME`/`0.0.0.0`) to avoid exposing the PTY on the LAN by default.
- Small robustness fix: tighten session close / null checks in the WS close handler.
- Update `src/server-pty-ws.test.ts` assertions to pin the hardened behavior and to ensure the server code reuses the proxy origin guards.

### Testing
- Ran `node --experimental-strip-types src/server-pty-ws.test.ts` and `node --experimental-strip-types src/proxy-behavior.test.ts`, both completed successfully.
- Ran `git diff --check` with no issues reported.
- Attempted `pnpm exec tsc --noEmit`, which could not complete in the environment due to missing module/type resolutions for `ws` and `node-pty` (local dev dependency availability); this is an environment/install issue rather than a code semantics failure.
- `pnpm install --frozen-lockfile` was attempted but blocked by the environment registry (403 fetching `ws@8.21.0`).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a2a666c66548327b789909ffff9dbed)

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -6,11 +6,17 @@ import { parse } from "node:url";
 import next from "next";
 import { WebSocket, WebSocketServer, type RawData } from "ws";
 
+import {
+  ACCESS_TOKEN_COOKIE,
+  isAllowedRequestSource,
+  isLoopbackHost,
+} from "./src/proxy-helpers.ts";
+
 const require = createRequire(import.meta.url);
 const pty: typeof import("node-pty") = require("node-pty");
 
 const ACCESS_TOKEN = process.env.COVEN_CAVE_ACCESS_TOKEN ?? "";
-const ACCESS_COOKIE = "coven_access_token";
+const ACCESS_COOKIE = ACCESS_TOKEN_COOKIE;
 
 type PtySession = {
   pty: import("node-pty").IPty;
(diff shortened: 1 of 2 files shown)


BunsDev:
Closing as duplicate of #714 (codex/propose-fix-for-unauthenticated-websocket-access), which provides a more comprehensive fix with timing-safe token comparison, fail-closed auth, and proper query param support.

PR #837, opened 2026-06-16
feat(familiars): link avatars directly to the .coven file (Tauri asset protocol)

BunsDev:
## What

Familiar avatars now link **directly to the workspace file** in `~/.coven/workspaces/familiars/<id>/avatars/` via Tauri's asset protocol (`convertFileSrc`), instead of being proxied through the `/api/familiars/[id]/avatar` Next route.

## Changes
- `/api/familiars` returns `avatarPath` (absolute path) + `avatarVersion` (mtime) instead of `avatarUrl`; `Familiar` / `ResolvedFamiliar` types updated.
- `FamiliarAvatar` builds `` via `convertFileSrc(avatarPath)` after mount (SSR-safe), cache-busted with `?v=<mtime>`; falls back to a Cave-local upload, then the glyph.
- Enables the Tauri asset protocol scoped to `$HOME/.coven/workspaces/familiars/*/avatars/*`.
- Deletes the now-unused avatar route + its `api-contracts` entry.

## Verification (static)
- ✅ `tsc --noEmit` clean (confirms every consumer compiles with the contract change)
- ✅ unit/contract tests: `familiar-resolve`, `familiar-avatar`, `api-contracts` (95 routes), server `familiar-avatar`
- ✅ `check:tests-wired` (353 files)
- ✅ `next build` (exit 0; familiars routes built, avatar route gone)

## ⚠️ Before merging — please weigh / runtime-verify
1. **Runtime not verified.** This is the **first asset-protocol use in Cave**, and the frontend is served from a bundled localhost server. Whether `convertFileSrc` URLs load cross-origin needs confirming by running the **packaged Tauri app** (not just `next dev` in a browser).
2. **Custom workspace roots.** The static asset scope only covers the default `~/.coven/workspaces`. `COVEN_HOME` / declared per-familiar workspace paths fall outside it → avatar degrades to the glyph. (The old route handled any path.)
3. **No downscaling.** Full-resolution source images now load (the route made 256px WebP). Accepted trade-off.


BunsDev:
**Follow-up (`235984a4`): added a fallback chain — resolves caveats #1 and #2.**

The direct `.coven` asset link is still the primary, but avatars now render through an ordered candidate chain that advances on `` load error:

1. `convertFileSrc(avatarPath)` — direct file link (Tauri, in-scope paths)
2. `/api/familiars/<id>/avatar` — **restored** route (downscales, works everywhere)
3. Cave-local uploaded data URL
4. glyph

So **non-Tauri/browser** rendering (caveat #1) and **custom `COVEN_HOME`/declared-workspace paths outside the asset scope** (caveat #2) now degrade to a working downscaled image instead of the glyph. The SSR-safe route URL paints first, then the webview upgrades to the direct link.

Remaining note: caveat #3 (the direct-link primary path serves full-res, no downscaling) stands by design. Runtime confirmation in the packaged Tauri app is still worth a look, but worst case is now "downscaled route image" rather than "glyph."

Verified: `tsc`, unit/contract tests (api-contracts 95, avatar route re-included), `check:tests-wired` (348), `next build` (exit 0, avatar route rebuilt).

PR #843, opened 2026-06-16
refactor(familiars): extract avatar asset-url + tauri guard into tested helper

BunsDev:
## What

Pulls the inline avatar-resolution logic out of `FamiliarAvatar` (the multi-candidate component landed in #841) into a pure, unit-tested module `@/lib/familiar-avatar-src`:

- `avatarAssetUrl(base, version)` — appends the `?v=<mtime>` cache-buster, with a falsy-version guard so it never emits `?v=undefined`.
- `canResolveWorkspaceAvatar()` — the `__TAURI_INTERNALS__` runtime guard for `convertFileSrc`.
- `initialAvatarSrc(familiar)` — SSR-safe initial `` decision (workspace path resolves client-side only).

## Why

The avatar resolution path is high-impact (every familiar avatar, across Tauri / browser / SSR) but was previously only **grep**-tested at the component level — a behavioral regression in the cache-bust query or the runtime guard would slip through. Extracting the decisions into a pure `.ts` module makes them genuinely unit-testable per the repo's lib-pure / component-grep convention.

## Tests

- New `src/lib/familiar-avatar-src.test.ts`: cache-bust on/off (incl. falsy `0`), Tauri vs plain-browser vs node runtime guard, SSR-safe initial-src branches.
- Updated `familiar-avatar.test.ts` grep to assert the component routes through `canResolveWorkspaceAvatar` / `avatarAssetUrl`.
- Wired the new test into `test:app` (`check:tests-wired` passes — 349 files).
- `pnpm typecheck` clean.

> Note: `initialAvatarSrc` is exported and tested but not consumed by the current candidate-array component; kept for completeness — say the word if you'd prefer it pruned.


BunsDev:
Closing as obsolete. This PR extracted helpers for the direct Tauri asset-protocol avatar path, but #846 ("revert(familiars): always serve avatars via the downscaled route") removed that path in favor of the universal downscaled `/api/familiars/[id]/avatar` route — so `avatarAssetUrl` / `canResolveWorkspaceAvatar` / `initialAvatarSrc` no longer have a caller. Not worth rebasing.

PR #849, opened 2026-06-16
fix(shell): align the sidebar wordmark with the nav-card column

BunsDev:
## What

Shrinks the floating-toggle clearance on the sidebar wordmark — `.shell-body--floats .sidebar-header--static` `padding-left: 40px → 15px` — so the "Coven Cave" wordmark's left edge lines up with the nav-card content column below it (Salem / New session / Home, and the WORK/KNOWLEDGE/TOOLS section labels) instead of being indented ~25px past it. The 15px still clears the 28px left float (right edge 36px).

## Scope note

This PR originally also raised the left nav toggle onto the wordmark row, but that fix landed independently as **#852** (`--left { top: 20px }`). After merging latest `main`, the net change here is **only** the wordmark horizontal alignment.

## Verification

Measured live DOM on current `main` + this change (1280×800): wordmark left edge 65px → 40px, matching the card-icon column; left float (right edge 36px) clears it.

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -3157,9 +3157,13 @@ button:active > .edge-rail-chip {
 }
 
 /* Clear the floating top-left toggle so it doesn't sit on top of the sidebar
-   wordmark. Scoped to the desktop shell — the floats are desktop-only. */
+   wordmark, while keeping the wordmark's left edge aligned with the nav-card
+   content column below it (Salem / New session / Home). The left float is 28px
+   wide at left:8 (right edge 36); 15px lands the wordmark just clear of it, at
+   the same indent as the card icons. Scoped to the desktop shell — the floats
+   are desktop-only. */
 .shell-body--floats .sidebar-header--static {
-  padding-left: 40px;
+  padding-left: 15px;
 }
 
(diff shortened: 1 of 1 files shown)


BunsDev:
⚠️ **Owner directive — please don't decouple the toggle heights.**

The repo owner has explicitly asked that the left nav toggle sit at the **same height (level) as the right side-panel toggle** — "the left sidepanel trigger absolutely must match the same height as the other sidepanel triggers."

This PR's `b85d0c66` does the opposite on purpose:
> "the left and right toggles intentionally no longer sit level with each other — each now reads as part of the panel it controls"

by giving `.shell-panel-float--left` its own `top: 21px` (vs the right's `top: 50px`). That's the misalignment the owner flagged (left toggle floating above the rest).

Please **drop the per-side `top` override** so the left toggle inherits the same base `top` as the right (i.e. keep them level — which is what `main` currently does). The wordmark-clearance `padding-left` tweak in this PR is fine to keep; it's specifically the left-toggle `top: 21px` raise that conflicts with the owner's requirement.

(Flagged on the owner's behalf — not editing this branch directly to avoid clobbering your work.)

PR #854, opened 2026-06-16
fix(chat): track the live side-panel header so floating toggles don't flash misaligned

BunsDev:
## Problem

Follow-up to #845. The shell's floating toggles (left nav, expand, side-panel trigger) were pinned to a fixed `top:50px` to center on the side-panel header. That header's top can shift while the layout settles right after load (transient chrome briefly sits above it, ~+43px), so the floats **flashed out of alignment** before snapping into place. Verified earlier: header center read 107 during the transient vs 64 settled.

## Fix

`ChatSurface` measures the live `.right-panel-tabs` header and publishes its centered top as a `--shell-float-top` root CSS var that `.shell-panel-float` consumes (`top: var(--shell-float-top, 50px)`). A short `requestAnimationFrame` loop tracks it **only until the value holds steady** (~6 frames), with a `resize` re-arm and full cleanup. The floats now move *with* the header through the settle, so they stay aligned at every frame instead of sitting at a fixed offset. The 50px fallback preserves the settled position for when no panel is open.

## Verification (in-app, 1440×900)

Measured float-vs-header center at three settle points; the CSS var tracks the header the whole way:

| settle | header center | expand Δ | left Δ | trigger Δ | `--shell-float-top` |
|---|---|---|---|---|---|
| 120ms (transient) | 107 | 0 | 0 | 0 | `93px` |
| 400ms | 64 | 0 | 0 | 0 | `50px` |
| 1800ms (settled) | 64 | 0 | 0 | 0 | `50px` |

**Δ=0px at every point** — the flash is gone. `tsc` clean; `right-panel-expand`, `chat-surface`, `chat-surface-polish`, `shell-edge-rails` pass; added source-text coverage for the var + measurement.

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -3167,8 +3167,11 @@ button:active > .edge-rail-chip {
   /* Vertically centered on the side-panel header row — the 38px Inspector/Debug/
      Changes tab strip that sits directly below the ~45px chat scope-tabs header.
      The trigger + expand toggle read as part of that header rather than floating
-     above it; the left nav toggle is moved to the same row to stay level. */
-  top: 50px;
+     above it; the left nav toggle is moved to the same row to stay level.
+     ChatSurface measures the live header and publishes --shell-float-top so the
+     floats track it through the post-load settle (no fixed-offset flash); the
+     50px fallback matches the settled position when no panel is open. */
+  top: var(--shell-float-top, 50px);
   z-index: 40;
   display: grid;
   place-items: center;
--- a/src/components/chat-surface.tsx
+++ b/src/components/chat-surface.tsx
@@ -446,6 +446,55 @@ export function ChatSurface({
     return () => root.removeAttribute("data-right-panel-open");
   }, [rightPanel, isMobile, scope]);
 
+  // Keep the shell's floating toggles (left nav, expand, side-panel trigger)
+  // vertically centered on the LIVE side-panel header. Its top can shift while
+  // the layout settles after load (e.g. transient chrome above the panel), so a
+  // fixed CSS offset flashes out of alignment. Publish the header's centered top
+  // as a root CSS var the floats consume (--shell-float-top), and track it via a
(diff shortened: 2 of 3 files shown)


BunsDev:
Heads-up from a parallel session — this change has a side-effect on the open-chat header.

After #854 landed, the right floating panel toggle (`.shell-panel-float--right`, "Show side panel") sits **inside the open-chat header row and on top of the session overflow menu** — the `⋮` "Session options" kebab added in #844/#847.

**Measured on `main` @ b2e81b4c (viewport 1280):**
- Floating toggle occupies **x:1244–1272, y:50–78**.
- The chat header (`.cave-chat-linear-header`) right-edge actions (search + `⋮`) live in that same band.
- `document.elementFromPoint()` at the kebab's center returns the toggle's SVG, and Playwright `.click()` on the kebab **times out as obscured** → the whole session menu (Rename / Project / Voice / Debug / Delete) is unreachable while this overlaps.

It was clickable before this merge (verified during #847), so it's a positioning side-effect, not a redesign bug. Two clean fixes, your call since the toggle is your area:
1. Offset the floating toggle leftward (or hide it) when an open-chat header with right-edge actions is present, or
2. Reserve right-gutter space in `.cave-chat-linear-header` so its actions clear the toggle's corner zone.

I didn't touch the toggle or the header — left it for you. Same notice is dropped in your `.worktrees/float-align/` worktree. Happy to take fix #2 (header gutter, self-contained) if you'd rather not touch it — just say so.

BunsDev:
Took the header-side fix in **#859** — reserves a 44px right gutter on `.cave-chat-linear-header` guarded by `:root:not([data-right-panel-open])`, so the chat header's find + ⋮ actions clear the floating toggle when no panel is open. Does **not** touch `.shell-panel-float`, so it won't collide with your toggle work. Verified live (8px gap, kebab clickable, menu opens).

PR #878, opened 2026-06-16
Potential fix for code scanning alert no. 9: Uncontrolled data used in path expression

BunsDev:
Potential fix for [https://github.com/OpenCoven/coven-cave/security/code-scanning/9](https://github.com/OpenCoven/coven-cave/security/code-scanning/9)

General fix approach: canonicalize both the candidate path and the allowed root with `fs.realpathSync`, then enforce containment (`candidate === root || candidate.startsWith(root + path.sep)`) before any filesystem sink use. Avoid “resolve-only” fallback for security decisions.

Best targeted fix here (without changing behavior more than necessary):
1. In `src/app/api/changes/route.ts`, after computing `real`, add an explicit containment check that `real` remains inside the canonical `allowedRoot`. If not, return 403.
2. In `src/lib/server/session-project-roots.ts`, harden `resolveWithinSessionRoots` so it only returns a candidate when canonicalization via `fs.realpathSync` succeeds (return `null` if it fails), and canonicalize provided roots before comparing. This removes the insecure `path.resolve` fallback from authorization decisions.
3. Keep existing external behavior (same statuses/messages where possible), but ensure all accepted paths are truly canonical and within trusted roots.

_Suggested fixes powered by Copilot Autofix. Review carefully before merging._

### DIFF
--- a/src/app/api/changes/route.ts
+++ b/src/app/api/changes/route.ts
@@ -89,6 +89,10 @@ async function resolveRepoRoot(projectRoot: string): Promise<RootResolution> {
   let stat: fs.Stats;
   try {
     real = fs.realpathSync(path.resolve(allowedRoot));
+    const allowedCanonical = fs.realpathSync(path.resolve(allowedRoot));
+    if (!(real === allowedCanonical || real.startsWith(allowedCanonical + path.sep))) {
+      return { ok: false, status: 403, error: "path not allowed" };
+    }
     stat = fs.statSync(real);
   } catch {
     return { ok: false, status: 404, error: "projectRoot does not exist" };
--- a/src/lib/server/session-project-roots.ts
+++ b/src/lib/server/session-project-roots.ts
@@ -30,6 +30,14 @@ function realpathOrResolve(value: string): string {
   }
 }
 
+function realpathStrict(value: string): string | null {
+  try {
+    return fs.realpathSync(path.resolve(value));
+  } catch {
+    return null;
+  }
+}
+
 function isWithinRoot(candidate: string, root: string): boolean {
   return candidate === root || candidate.startsWith(root + path.sep);
 }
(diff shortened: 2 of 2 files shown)


BunsDev:
**Recommend closing — autofix is a no-op.**

The `route.ts` change adds:

```ts
real = fs.realpathSync(path.resolve(allowedRoot));
const allowedCanonical = fs.realpathSync(path.resolve(allowedRoot));  // same expression
if (!(real === allowedCanonical || real.startsWith(allowedCanonical + path.sep))) {
  return { ok: false, status: 403, error: "path not allowed" };
}
```

`real` and `allowedCanonical` are computed from the **same input** through the same `realpathSync(path.resolve(...))` call, so they are always equal — the containment check is tautological and never returns 403.

The CodeQL alert (CWE-22 on alert #9) appears to be a **false positive** for this site:
- `allowedRoot` is the value returned from `isAllowed(projectRoot)` (lines 77–85).
- `isAllowed` only returns a value if the candidate matches `resolveAllowedProjectPath` (static allow-list, returns realpath after containment) **or** `resolveWithinSessionRoots` (also returns realpath after containment via `isWithinRoot`).
- By the time we reach line 91, the path has already been canonicalized and contained.

The `session-project-roots.ts` portion of this autofix is reasonable on its own (rejects unresolvable candidates and re-canonicalizes roots), but it's not enough to justify keeping this PR open with the broken `route.ts` hunk.

Suggested next step: close this PR and dismiss the alert as a false positive (or, if we want defense-in-depth on `route.ts`, open a fresh PR that compares `real` against a separately-computed canonical of the **input** `projectRoot`, not against `allowedRoot`).

BunsDev:
Closing per analysis above — the `route.ts` autofix hunk is a no-op (compares `realpathSync(path.resolve(allowedRoot))` to itself), and the underlying alert is a false positive at this site since `isAllowed()` already canonicalizes + contains before line 91. The `session-project-roots.ts` hardening is reasonable on its own; if we want it, we can cherry-pick into a fresh PR.

PR #909, opened 2026-06-18
feat(chat): context meter + WYSIWYG model forwarding in composer

BunsDev:
## What

Two composer changes that came out of "show the context meter + make model selection actually change the runtime model":

### 1. Context meter (fully working)
A thin gauge in the composer settings row — fill bar + `37% · 74k/200k` — next to the model picker.

- New pure lib `src/lib/model-context-windows.ts`: model id → context-window tokens (exact catalog ids + pattern fallbacks). Unknown models return `null` so the meter **hides** instead of inventing a denominator and showing a fake percentage.
- `contextTokens(usage)` in `usage-format.ts` = `input + cacheRead + cacheCreation` — the real prompt size occupying the window (not input+output, not a running sum).
- Driven by the most recent settled turn's usage over its `responseMetadata.model` (falling back to the picker's effective model).

### 2. WYSIWYG model forwarding
`chat-view` now includes `modelOverride` + `modelOverrideScope: "session"` in the `/api/chat/send` body, sourced from the picker's effective model. So a turn runs on exactly what the UI shows rather than a stale persisted intent or the familiar/global default. Synthetic local pseudo-models are excluded.

## Known limitation — the runtime half is gated downstream

The actual model switch requires `coven run --model`, which the send route already forwards **but capability-gates** behind a `coven run --help` probe. The installed CLI (**v0.0.39**) does not advertise `--model`, so forwarding is a deliberate no-op today. This PR makes the Cave side fully correct so it activates automatically the moment the coven CLI ships the flag. That flag lives in the **coven repo** (outside this repo) and is the remaining work to make selection change the live model end-to-end.

## Tests
- New `model-context-windows.test.ts` (wired into `test:app`): exact/pattern/unknown windows, clamp, rounding, null-on-unknown, and the usage→meter end-to-end path.
- Full `test:app` + `test:api` green; `tsc --noEmit` clean.


BunsDev:
Closing in favour of #911 (chat-context-meter-model-wire) which shipped the context meter in the chat header with a cleaner implementation. This branch conflicts with main after #911 merged.

PR #1012, opened 2026-06-19
feat(library): replace reading-status select with 3-way segmented toggle

BunsDev:
Replaces the reading-status dropdown select with a compact 3-way segmented toggle (Want to Read / Reading / Read) for a cleaner library UX.


BunsDev:
Already merged to main as #985 — stale branch. Closing.

PR #1161, opened 2026-06-20
feat(code): render file-preview load errors as an error state, not fake code

BunsDev:
## What

When a file failed to load in the Code file preview, the error was stuffed into the preview as **fake source** — `// Error: <msg>` — rendered with syntax highlighting, line numbers, and an enabled **Edit** button, as if it were the file's real content.

Added a dedicated preview `kind: "error"` that renders a centered warning icon + the message + the filename (`role="alert"`). The Edit / Save / raw-toggle controls already gate on `kind === "text"`, so they correctly stay disabled for the error state.

## Tests / verification

- Full `pnpm build` (test:app + next build) green; `tsc --noEmit` clean (the new union variant didn't break the existing `preview?.content` accesses).
- Live-verified with `/api/project-file` mocked to fail: clicking a file shows a `role="alert"` with the warning icon + "Permission denied." + filename, and the old `// Error:` fake-code content is gone. Confirmed programmatically (`fakeCode: false`) + screenshot.

### DIFF
--- a/src/components/comux-view-terminal.test.ts
+++ b/src/components/comux-view-terminal.test.ts
@@ -279,7 +279,7 @@ assert.match(
 // unsupported-extension code block.
 assert.match(
   source,
-  /type ProjectFilePreview =[\s\S]*?\| \{ kind: "image"; dataUrl: string; mimeType: string; size\?: number \};/,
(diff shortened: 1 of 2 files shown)


BunsDev:
Superseded by #1163, which shipped the same fix (file-preview load errors no longer render as fake `// Error:` code) using the shared ErrorState/EmptyState — a cleaner approach than this PR's inline variant. Closing as redundant (duplicate-PR race; #1163 landed while this was in review).

PR #1652, opened 2026-06-23
fix(ios): un-red main — loosen updateTask test assertion after #1651

BunsDev:
## Why

#1651 added a `notes:` parameter to `CaveClient.updateTask`, but `scripts/ios-task-actions.test.mjs` pinned the **exact** old signature (`status:priority:steps:`), so `test:mobile` — and therefore the **Frontend build** required check — went red on `main`.

## Fix

Loosen the `updateTask` signature assertion to match the core params (`status`, `priority`, `steps`) without pinning the full list, so future optional fields don't re-break it.

## Verification

- `pnpm test:mobile` → **✓ 25 test file(s) passed** (the full suite, not just one test — which is how the original miss slipped through).

### DIFF
--- a/scripts/ios-task-actions.test.mjs
+++ b/scripts/ios-task-actions.test.mjs
@@ -18,11 +18,12 @@ const detail = await readFile(
   "utf8",
 );
 
-// Client speaks the board mutation contract.
+// Client speaks the board mutation contract. (Match the core params loosely so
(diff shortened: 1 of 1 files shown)


BunsDev:
Redundant — main was already fixed by another session's change that updated the updateTask assertion to include `notes:`. Closing.

PR #1899, opened 2026-06-24
fix(a11y): workflow run detail — live status region + hidden-pause poll + aria-controls

BunsDev:
## Workflow run detail audit (`workflow-run-progress.tsx` + `workflow-runs-panel.tsx`)

Solid foundation: the transcript poll is `alive`-guarded against stale/unmount, status is shown as **text** (not color-only), runs use `RelativeTime`, reduced-motion is global, and the spinner is `aria-hidden`. Four fixes:

1. **Live status is announced** — `WorkflowRunProgress`'s headline updates "Running: `<step>`" → "All N steps reported" live, but was silent to screen readers. Now `role="status"`/`aria-live="polite"`. The transcript-fetch error gains `role="alert"`.
2. **Transcript poll pauses when hidden** — the 2.5s poll kept hitting `/api/chat/conversation/<id>` in a backgrounded tab. Now skips the fetch while `document.hidden` and resumes (with an immediate refresh) on `visibilitychange`; a stale timer is cleared before each tick so resume can't double-poll.
3. **Disclosure toggles get `aria-controls`** — both the run row (→ its expansion region `id`) and the step-detail toggle (→ its `<pre>` `id`, per-instance via `useId`) had `aria-expanded` with no controlled-region link.

## Verification
- New `workflow-run-detail-a11y.test.ts` (wired; `check:tests-wired` ✓) + existing `workflow-studio.test.ts` pass · `tsc` clean · `pnpm build` exit 0
- No live Playwright pass: the runs panel renders deep inside the workflow **studio** (select a workflow → studio → runs), which isn't practically drivable headlessly. The changes are static markup (aria-live / aria-controls / role=alert) plus a hidden-pause poll that mirrors the same pattern shipped+verified on Calls/Schedules/Coven-floor — so source-text + type + build are the verification here.

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -140,6 +140,7 @@ export const SUITES = {
     "src/components/workflows-view.test.ts",
     "src/components/workflows/workflow-studio.test.ts",
     "src/components/workflows/workflow-step-list.test.ts",
+    "src/components/workflows/workflow-run-detail-a11y.test.ts",
     "src/components/workflows/workflow-capability-attachments.test.ts",
     "src/components/projects-view.test.ts",
     "src/components/dnd-context-stable-ids.test.ts",
--- a/src/components/workflows/workflow-run-detail-a11y.test.ts
+++ b/src/components/workflows/workflow-run-detail-a11y.test.ts
@@ -1,0 +1,45 @@
+// @ts-nocheck
+import assert from "node:assert/strict";
+import { readFileSync } from "node:fs";
+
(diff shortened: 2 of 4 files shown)


BunsDev:
Closing as obsolete: this audited the legacy Workflows run detail (workflow-run-progress.tsx + workflow-runs-panel.tsx), but those components were deleted on main by 6579c0d0 "Remove legacy Workflows page" (the legacy Workflows page is replaced by the new Flow surface). This is orphaned work — the surface no longer exists, so the a11y fixes are moot. No files resurrected.

PR #1931, opened 2026-06-25
chore(nav): rename the Board surface to "Tasks"

BunsDev:
Renames the kanban surface's display label from **Board** to **Tasks** — it's about tasks, and the iOS app already calls it Tasks.

- `sidebar-minimal.tsx`: nav label `Board` → `Tasks`.
- `workspace.tsx`: `WORKSPACE_MODE_TITLES.board` → `Tasks`.
- Updated the source-text test that pinned the old label.

The mode id stays `board`, so routes, deep links, the `board-*` components, badges, and the ⌘3 shortcut are all unchanged — display-label only. The command palette picks up the new label automatically (it reads `fm.label`).

## Verified
`tsc --noEmit` clean; app source-text suite green (421).

### DIFF
--- a/src/components/sidebar-minimal.test.ts
+++ b/src/components/sidebar-minimal.test.ts
@@ -109,8 +109,8 @@ assert.match(
 
 assert.match(
   source,
-  /\{ id: "board", label: "Board", iconName: "ph:kanban", group: "work", kbd: "⌘3", description:/,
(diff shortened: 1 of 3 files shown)


BunsDev:
Folding the Board→Tasks rename into #1929 — both touch the nav and their E2E fixes interact (they'd conflict on foundations.spec.ts otherwise). Consolidating into one coherent 'simplify nav' PR.

PR #1977, opened 2026-06-28
fix(group-chat): strip next-paths block from coven replies

BunsDev:
## What

The Group Chat ("coven") surface rendered raw assistant reply text, so the piggybacked `<coven:next-paths>` suggestions block (and its streaming partial) leaked into the bubble as **literal control markup** — visible in a Nova+Kitty broadcast.

The single-chat surface already strips this via `extractNextPaths`; the coven fan-out never did.

## Change

- `group-chat-view.tsx`: render `extractNextPaths(r.text).visible` instead of raw `r.text`, mirroring `chat-view`. Streaming-safe (hides the block while it streams).
- `group-chat-view.test.ts`: source-text assertion pinning the strip so it can't regress.

**Strip-only:** no chips surfaced yet, no fan-out/relay change. (`group-chat.ts` untouched.)

## Verify

- `pnpm test:app` — 418 files ✓
- `pnpm test:mobile` — 65 files ✓
- `pnpm typecheck` — exit 0 ✓

### DIFF
--- a/src/components/group-chat-view.test.ts
+++ b/src/components/group-chat-view.test.ts
@@ -20,6 +20,8 @@ test("GroupChatView broadcasts via /api/chat/send and reuses pure helpers", () =
   assert.match(view, /recordSession\(group\.id, reply\.familiarId/, "pins each familiar's session id");
   // A Stop control aborts the in-flight broadcast.
   assert.match(view, /abortRef\.current\?\.abort\(\)/, "Stop aborts the broadcast");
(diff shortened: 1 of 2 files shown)


BunsDev:
Superseded by #1976 — identical strip-only fix (same 2 files, +9/-1, same `extractNextPaths(r.text).visible`) landed on main first via a concurrent session. Closing as duplicate; the fix is already shipped.

PR #1991, opened 2026-06-28
fix(windows): launch coven via node, not the .cmd shim, so /api/harnesses stops 500ing

YogiSotho:
## Problem

On Windows, `/api/harnesses` returns **HTTP 500 (`spawn EINVAL`)**, so the onboarding *"Create your familiar"* step renders a **blank Option A / empty runtime list** even when runtimes (Claude Code, Hermes) are installed. The user gets stuck at onboarding with nothing selectable.

### Root cause

npm installs `coven` as a `.cmd` shim on Windows. Since the CVE-2024-27980 hardening (Node ≥18.20 / 20.12 / 21.7), `child_process.spawn()` / `execFile()` throw `EINVAL` when handed a `.cmd`/`.bat` unless `shell: true`. Several routes spawn `covenBin()` (which resolves to `coven.cmd`) directly:

```
⨯ Error: spawn EINVAL
    at covenSupportsAdapterList (src/app/api/harnesses/route.ts)
    at GET
  errno: -4071, code: 'EINVAL', syscall: 'spawn'
```

`covenSupportsAdapterList()` rejects → `GET` 500s → the client's `loadHarnesses` swallows the error → `harnesses` stays `[]` → blank grid.

The recent empty-state + PATH re-probe work (#1985 / #1988) improves the *message* but not the outcome on Windows: PATH was never the problem — the `.cmd` spawn is.

## Fix

Add `covenInvocation()` to `coven-bin.ts`: on a Windows `.cmd`/`.bat` shim it resolves the underlying Node script and returns `{ command: <node>, prefixArgs: [script] }`; it's the identity on macOS/Linux. Unlike `shell: true`, running `node <script>` directly stays **safe for the prompt-bearing `chat/send` argv** (no shell quoting/injection).

Routed the broken spawn/exec sites through it — `harnesses`, `onboarding/status`, `chat/send`, `coven-version`, `board/enrich-steps`, `coven/exec` — and made the harness version probe `EINVAL`-safe. `daemon/start` and `onboarding/install` already used `shell: process.platform === "win32"` for their fixed argv and are left unchanged.

## Verification

- Repro on Windows 11 / Node 24.12: `/api/harnesses` flipped **500 → 200**, now returns Claude Code + Hermes as `installed: true`; Option A renders them selectable.
- `coven-bin.test.ts` extended with assertions for the new path (kept alongside the `refreshCovenSpawnEnv` assertion from #1985); passes.
- Type-strip syntax check passes on all touched files.

### DIFF
--- a/src/app/api/board/enrich-steps/route.ts
+++ b/src/app/api/board/enrich-steps/route.ts
@@ -12,7 +12,7 @@ import {
 } from "@/lib/cave-board-types";
 import { normalizeTaskGitHubLinks } from "@/lib/task-github";
 import { bindingFor, loadConfig } from "@/lib/cave-config";
-import { covenBin, covenSpawnEnv } from "@/lib/coven-bin";
+import { covenInvocation, covenSpawnEnv } from "@/lib/coven-bin";
 import { familiarWorkspace } from "@/lib/coven-paths";
 import { isTrustedChatHarness } from "@/lib/harness-adapters";
 import { spawn } from "node:child_process";
@@ -221,7 +221,8 @@ function runCoven(
     try {
       let out = "";
       let settled = false;
-      const child = spawn(covenBin(), args, {
+      const { command, prefixArgs } = covenInvocation();
+      const child = spawn(command, [...prefixArgs, ...args], {
         cwd: familiarWorkspacePath ?? process.cwd(),
         stdio: ["ignore", "pipe", "pipe"],
         env: covenSpawnEnv(),
--- a/src/app/api/chat/send/route.ts
+++ b/src/app/api/chat/send/route.ts
@@ -31,7 +31,7 @@ import {
   toPersistedTools,
   ToolCallTracker,
 } from "@/lib/chat-tool-events";
-import { covenBin, covenSpawnEnv } from "@/lib/coven-bin";
+import { covenInvocation, covenSpawnEnv } from "@/lib/coven-bin";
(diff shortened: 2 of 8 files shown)


BunsDev:
Closing this per the current maintenance protocol: OpenCoven/coven-cave is not accepting external PRs until July 1.

The Windows .cmd shim launch failure is a valid issue. We are carrying the fix internally instead of merging an external contributor branch during the freeze. The internal patch keeps prompt-bearing argv out of shell mode and adds behavior coverage around Windows shim resolution.

No action is needed on this PR.

BunsDev:
Internal tracking is now in #1993, and the first-party fix is in #1992. This PR remains closed under the external PR freeze.

YogiSotho:
Makes sense given the freeze — glad the Windows `.cmd` fix landed in #1992. Looks like the same approach (launch the npm shim through `node <coven.js>`, same spawn/exec call sites), so the outcome's the same for Windows users either way. I'll hold PRs until July 1 and would be happy to pick up follow-ups after that.

BunsDev:
Belated note for the record: this PR was closed under the [external PR freeze policy](https://github.com/OpenCoven/coven-cave/blob/main/CONTRIBUTING.md) (no external PRs through July 1st, 2026).

The technical context has been preserved as **issue [#2011](https://github.com/OpenCoven/coven-cave/issues/2011)**. Internal implementation is in progress.

Thank you for the thorough investigation, @YogiSotho.

PR #1992, opened 2026-06-28
fix(windows): launch coven npm shim through node

### LINKED ISSUE #1993: fix(windows): coven npm shim breaks harness discovery and onboarding runtimes
## Problem

On Windows, npm installs `coven` as a `.cmd` shim. Modern Node hardening for `child_process.spawn()` / `execFile()` rejects `.cmd` / `.bat` targets unless the process is launched through a shell. Cave currently invokes `covenBin()` directly from several app routes, so a Windows `coven.cmd` can surface as `spawn EINVAL`.

This creates two user-visible failures:

- `/api/harnesses` can return HTTP 500 while probing Cave harness adapters.
- onboarding Option A / runtime discovery can render an empty or incomplete runtime list even when local runtimes are installed.

## Requirements

- Keep prompt-bearing argv out of `shell: true`.
- Resolve Windows npm `.cmd` shims to the underlying `coven.js` script and launch through `node <script>`.
- Preserve direct binary execution for macOS/Linux and non-shim Windows binaries.
- Cover both harness discovery and onboarding status paths.
- Keep existing fixed-argv installer/daemon Windows shell behavior unchanged unless separately required.

## Internal fix

Internal PR #1992 carries this as first-party work.

BunsDev:
## Summary
- close #1993 by routing Windows npm `.cmd` shims through `node <coven.js>` before appending dynamic args
- update Cave coven spawn/exec call sites used by harness discovery, onboarding status, chat send, coven exec/version, and board step enrichment
- add behavior coverage for direct POSIX launch, npm shim parsing, and fallback npm-global layout

## Problem Coverage
- `/api/harnesses`: avoids direct `spawn(coven.cmd, ...)` for adapter list probes and degrades version-probe spawn errors to a missing version instead of a route-level 500
- onboarding Option A/runtime list: `onboarding/status` uses the same safe launch path for adapter summaries, so installed runtimes are not hidden by Windows shim `EINVAL`

## Verification
- `node src/lib/coven-bin.test.ts`
- `pnpm check:tests-wired`
- `npx tsc --noEmit`
- `git diff --check`
- `node scripts/run-tests.mjs app`
- `CIRCLE_NODE_TOTAL=1 npx next build`

Closes #1993

### DIFF
--- a/src/app/api/board/enrich-steps/route.ts
+++ b/src/app/api/board/enrich-steps/route.ts
@@ -12,7 +12,7 @@ import {
 } from "@/lib/cave-board-types";
 import { normalizeTaskGitHubLinks } from "@/lib/task-github";
 import { bindingFor, loadConfig } from "@/lib/cave-config";
-import { covenBin, covenSpawnEnv } from "@/lib/coven-bin";
+import { covenLaunchCommand, covenSpawnEnv } from "@/lib/coven-bin";
 import { familiarWorkspace } from "@/lib/coven-paths";
 import { isTrustedChatHarness } from "@/lib/harness-adapters";
 import { spawn } from "node:child_process";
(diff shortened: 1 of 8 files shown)


YogiSotho:
Glad this shipped 🎉 — just flagging for the record that it matches my #1991 (opened 03:03Z, ~80 min before this PR at 04:25Z): same approach (launch the npm `.cmd` shim through `node <coven.js>`) and the same spawn/exec call sites. Totally understand carrying it as first-party during the freeze — would you be open to adding a co-author credit on the merge so the provenance is on the record? Happy to keep helping after July 1.

PR #2001, opened 2026-06-28
fix(sidecar): ship sharp so familiar avatars render in the packaged app

YogiSotho:
## Problem

Familiar (agent) avatars don't render in the packaged desktop app — most visibly the macOS DMG — even though they render fine in dev (`localhost:3000`). SVG / initial avatars show; uploaded raster avatars come back blank.

## Root cause

`src/app/api/familiars/[id]/avatar/route.ts` imports `sharp` at runtime to downscale + transcode workspace avatars (the seeded ones are ~30MB / 4096px PNGs under `~/.coven/workspaces/familiars/<id>/avatars/`). But `sharp` was dropped from the sidecar bundle **twice over**:

1. It was a **devDependency**, so the sidecar's `pnpm install --prod` (`scripts/sidecar-bundle.sh`) skipped it.
2. `prune_sidecar_nonruntime_files` explicitly `rm -rf`'d `node_modules/sharp` **and** `node_modules/@img`.

So `require('sharp')` threw in the packaged Node server and the route 404'd for every raster image. SVG avatars skip the sharp branch, so they kept working — which is why it looked like avatars were only *half* broken. `images.unoptimized` disables Next's own image optimizer but not this direct `sharp` use in the route.

## Fix

- Move `sharp` → `dependencies` so the production sidecar install includes it.
- Add `sharp` to `serverExternalPackages` so Next doesn't trace/strip its native `.node` binaries.
- Stop pruning `sharp` / `@img` in `sidecar-bundle.sh`; add a guard that fails the build if the `@img/sharp-<target>` native binary is missing.
- Flip `sidecar-bundle-deps.test.mjs` to assert `sharp` is **kept** (it was asserting removal — codifying the bug).

`prune_foreign_native_packages` still trims `@img` down to just the build target, so the bundle stays small, and `pnpm.supportedArchitectures` already fetches the per-OS binaries.

## Verification

- `dependency-policy`, `sidecar-bundle`, and `sidecar-bundle-deps` tests pass.
- Replicated the bundler's `pnpm install --prod --frozen-lockfile` staging step: `node_modules/sharp` is now present, and `@img/sharp-darwin-arm64` / `-x64` + libvips are fetched — the binaries the macOS DMG needs.

## Build note

The release DMG must be built on macOS — the prune keys off the build host's arch, same as `@next/swc` and `node-pty` already do.

### DIFF
--- a/next.config.ts
+++ b/next.config.ts
@@ -20,7 +20,12 @@ const nextConfig: NextConfig = {
   // Cave does not use Next's server-side image optimizer in the packaged app;
   // icon generation happens before build via scripts/generate-pwa-icons.mjs.
   images: { unoptimized: true },
-  serverExternalPackages: ["node-pty"],
+  // node-pty and sharp are native addons. Keep them external so Next doesn't try
+  // to trace/bundle their platform binaries into .next/ (which strips the .node
+  // files). sharp powers the familiar avatar route's downscale/transcode — see
+  // src/app/api/familiars/[id]/avatar/route.ts. It must resolve from node_modules
+  // at runtime, so it has to be a production dependency that the sidecar ships.
+  serverExternalPackages: ["node-pty", "sharp"],
   // Next.js file tracing otherwise sucks the entire src-tauri/target/
   // (multi-GB) into the standalone bundle. Exclude noisy siblings.
   outputFileTracingExcludes: {
--- a/package.json
+++ b/package.json
@@ -72,6 +72,7 @@
     "react-colorful": "5.7.0",
     "react-dom": "19.2.4",
     "react-resizable-panels": "4.11.2",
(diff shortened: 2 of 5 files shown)


BunsDev:
Thank you for the thorough investigation, @YogiSotho — the root-cause writeup is excellent and accurate.

Coven Cave is **not accepting external pull requests through July 1st, 2026**. See [CONTRIBUTING.md](https://github.com/OpenCoven/coven-cave/blob/main/CONTRIBUTING.md) for the full policy.

The technical context from this PR has been **preserved as issue [#2010](https://github.com/OpenCoven/coven-cave/issues/2010)** so the work isn't lost; the fix will land via the internal contribution path. Closing this PR per the freeze.

PR #2018, opened 2026-06-28
feat(library): inline GitHub/article reader + /research integration

BunsDev:
## Why

Refactors the Library into a **reader-centric** surface: instead of bouncing GitHub repos and saved articles out to an external browser or showing bare metadata cards, they now open as real reading surfaces inside the Cave — and a new `/research` flow lets a familiar research a topic and drop the cited brief straight into the Library.

### Custom inline viewers
- **GitHub repos** — `GET /api/library/github-repo` fetches repo metadata + README (`src/lib/github-repo.ts`, public API, optional `GITHUB_TOKEN`). The reader renders the README inline through the same markdown pipeline as research docs, with a repo header (★ stars, forks, language, license, topics). Issues/PRs keep the embedded link viewer.
- **Articles** — `GET /api/library/article` + `src/lib/article-extract.ts` do dependency-free readability extraction (title, byline, lead image, body → markdown). Saved bookmarks/reading articles open in a clean **Reader** view with a **Reader | Page** toggle; extraction failures fall back to the embedded page / structured details automatically.

### `/research` integration (headline)
- `POST /api/library/research` (SSE) runs a deep-research prompt through the active familiar's agent and writes the synthesized, cited brief into the familiar's `research/research/` collection so it appears in the Library next to hand-written docs.
- A **research composer** (rail **"New research"** button + the `/research <topic>` slash command) starts runs, streams status, and opens the resulting doc. `/research` is wired chat → workspace → Library.

## Safety
- Article fetch blocks loopback/private/link-local hosts (SSRF), caps the body at 2MB to `text/html`, times out at 12s.
- Research output path is slug-derived (no user-controlled path segments) under the familiar root, written atomically.
- GitHub repo slug validated against `[A-Za-z0-9._-]`.

## Tests
- Unit tests for all three pure libs (`github-repo`, `article-extract`, `research-run`), wired into `run-tests.mjs`.
- New routes added to the `api-contracts` allowlist.
- `pnpm typecheck` clean; `next build` green; existing Library source-text tests still pass.


BunsDev:
## ⚠️ CodeQL findings need review before merge

CI has 5 real security alerts on this PR:

### 🔴 Critical: SSRF
- **`src/app/api/library/article/route.ts:56`** — `fetch(target.toString())` where `target` comes from user-supplied `?url=` query param.

The route DOES have an `isPublicHost()` pre-check + per-redirect re-check (and only allows http/https + caps response size at 2MB + 12s timeout), so it's defensible — but CodeQL doesn't recognize `isPublicHost` as a taint sanitizer.

### 🟠 High × 4: Incomplete multi-character sanitization
- **`src/lib/article-extract.ts:45`** — `<script` tag stripping
- **`src/lib/article-extract.ts:91`** — `<script` tag stripping
- **`src/lib/article-extract.ts:137`** — `<iframe` tag stripping
- **`src/lib/article-extract.ts:152`** — `<script` tag stripping

The `stripHtmlTags` and `dropElements` helpers ALREADY use a fixpoint loop (`do/while (out !== prev)`) which defeats the typical `<<script>script>` bypass — but CodeQL is pattern-matching the bare regex and missing the loop semantics.

### Options

1. **Defer (cheap, dishonest)**: Add `// lgtm[js/...]` suppression comments with justification. CodeQL will pass.
2. **Add DOMPurify (correct, more deps)**: Replace the in-tree HTML scrubber in `article-extract.ts` with `isomorphic-dompurify` (or similar). Eliminates the high-severity findings cleanly.
3. **Tighten article reader**: Add an explicit user-facing allowlist of hosts (defeats the reader's purpose for the arbitrary-URL use case).
4. **Hold & redesign**: Park this PR; rethink whether the article reader should be a server-side proxy at all (vs. a sandboxed iframe in the renderer).

Recommend **option 2** if we want to keep the feature shape, since DOMPurify is well-vetted and the dep cost is small. But the CodeQL SSRF flag on the route may still need a suppression comment + justification regardless, since the underlying request-forgery surface is intentional for an article reader.

cc @BunsDev — your call on how to proceed.

Issue #2011, opened 2026-06-28
Windows packaged-app: /api/harnesses returns 500 (spawn EINVAL) because coven.cmd shim can't be spawned directly


BunsDev:
## Source

Original report: external PR [#1991](https://github.com/OpenCoven/coven-cave/pull/1991) by @YogiSotho (closed; the technical context preserved here as an issue per the [external PR freeze policy](https://github.com/OpenCoven/coven-cave/blob/main/CONTRIBUTING.md)).

## Problem

On Windows: `/api/harnesses` returns **HTTP 500 (`spawn EINVAL`)**, so the onboarding *"Create your familiar"* step renders a **blank Option A / empty runtime list** even when runtimes (Claude Code, Hermes) are installed. The user gets stuck at onboarding with nothing selectable.

## Root cause

npm installs `coven` as a `.cmd` shim on Windows. Since the CVE-2024-27980 hardening (Node ≥18.20 / 20.12 / 21.7), `child_process.spawn()` / `execFile()` throw `EINVAL` when handed a `.cmd` or `.bat` unless `shell: true`. Several routes spawn `covenBin()` (which resolves to `coven.cmd` on Windows) directly:

```
⨯ Error: spawn EINVAL
    at covenSupportsAdapterList (src/app/api/harnesses/route.ts)
    at GET
  errno: -4071, code: 'EINVAL', syscall: 'spawn'
```

`covenSupportsAdapterList()` rejects → `GET` 500s → the client's `loadHarnesses` swallows the error → `harnesses` stays `[]` → blank grid.

The earlier empty-state + PATH re-probe work (#1985 / #1988) improves the *message* but not the outcome on Windows: PATH was never the problem — the `.cmd` spawn is.

## Suggested fix shape

Add a `covenInvocation()` helper to `src/lib/coven-bin.ts`: on a Windows `.cmd` / `.bat` shim, resolve the underlying Node script and return `{ command: <node>, prefixArgs: [script] }`. Identity on macOS / Linux.

**Critical:** prefer `node <script>` over `shell: true` so the **prompt-bearing `chat/send` argv stays safe** from shell quoting / injection — `shell: true` is unsafe when any arg contains user-controlled text.

Route the broken spawn / exec sites through it. Likely sites (verify against current main before changing):
- `src/app/api/harnesses/route.ts` (`covenSupportsAdapterList`)
- `src/app/api/onboarding/status/route.ts`
- `src/app/api/chat/send/route.ts`
[...]

PR #2025, opened 2026-06-28
docs(plan): cross-OS conformance suite (Slice A of #1990)

### LINKED ISSUE #1990: Add cross-environment tests that verify the app actually works on every target OS
# Add cross-environment tests that verify the app actually works on every target OS

## Problem

We develop and ship across multiple environments (Windows / macOS / Linux), but
we don't yet have a test suite that actually *exercises and verifies* behavior
on each one. We need cross-environment tests that prove core functionality works
per-OS — not just that the code compiles, but that it runs correctly given each
platform's paths, shells, env-var handling, and packaging.

A hard requirement: these tests must genuinely run on each target. A test that
silently short-circuits to a no-op when a platform's preconditions aren't met
produces a misleading green check — it isn't verifying anything. Any such case
must be an *explicit, reasoned skip*, not a fake pass.

## What we want

1. **Cross-OS verification in a matrix.** Run the test suite on `ubuntu-latest`,
   `windows-latest`, and `macos-latest` from the *same* job spec, so the same
   definition of "works" is executed on every platform.

2. **A shared conformance/contract suite.** Define behavior against an
   abstraction (paths, line endings, env lookup, networking, process/terminal
   spawn); per-OS implementations swap underneath. The contract suite runs
   identically on every OS and is the source of truth for "it works here."

3. **Cover the things that actually differ per environment.** Filesystem paths,
   shell/env-var semantics (`$VAR` vs `$env:VAR` vs `%VAR%`), line endings,
   packaging/installer behavior, and any OS-specific binaries or sidecars.

4. **One neutral baseline job.** A single generic job (Linux) for the
   platform-agnostic checks — shared logic, formatting, lint — that every PR
   must pass. No OS-specific behavior lives here; that's what the matrix is for.

5. **Make skips explicit, not blank (quality bar).** When a platform genuinely
   can't satisfy a precondition, mark it skipped/pending with a reason
   (e.g. `skipped: requires Linux mDNS`) so the report shows the gap instead of
   a falsely green check.

6. **Pin and document the neutral defaults.** Default ports, config, and paths
   are the baseline; per-OS deltas are listed alongside them.

## Acceptance criteria

- [ ] The test suite runs on Windows, macOS, and Linux via a CI matrix.
- [ ] A shared conformance suite verifies core behavior identically on all three
      OSes.
- [ ] Per-environment differences (paths, shells/env vars, line endings,
      packaging) are exercised by real assertions.
- [ ] No test passes via a silent platform short-circuit; such cases are
      explicit skips with reasons.
- [ ] Neutral defaults (ports/paths/config) and their per-OS deltas are
      documented in the repo.

## Notes

Reference pattern: split a generic baseline job (one Linux job running the
shared, platform-agnostic checks) from the cross-OS verification matrix that
runs the suite on every target — neutral baseline + platform matrix.

BunsDev:
Plan-only PR for the Slice A scope of #1990. Implementation lands in follow-up PRs (subagent-driven, one PR per task or grouped — TBD per review checkpoints).

**Why a draft PR for a plan?** Makes the 9-task implementation plan reviewable in-place, gives a stable URL to reference from #1990 + commit messages, and avoids the "plan lives in a side branch nobody can find" failure mode.

## Scope

Adds [`docs/cross-os-conformance-plan.md`](./docs/cross-os-conformance-plan.md) (983 lines, 9 tasks) describing the **Slice A** approach: a single CI matrix job (`ubuntu/windows/macos`) running a shared conformance suite anchored on the exact bug classes that produced #2010 + #2011.

## What Slice A lands

| | |
|---|---|
| New | `scripts/sidecar-target.mjs` — pure single-source-of-truth for the `(platform, arch, libc) → native-package` mapping. Dual-use: simulated-args unit-testable AND callable from the conformance test with real `process.platform`. |
| New | `scripts/conformance-cross-os.test.mjs` — runs on every matrix OS. Asserts the real `covenLaunchCommandForBinary` branch + `sidecarTarget` → shipped `@img/sharp-*` dir + path/delimiter/EOL/env-casing + `child_process.spawn` smoke. |
| New | `scripts/expect-or-skip.mjs` — throws on empty reason so silent platform short-circuits are banned (criterion 5). |
| New | `docs/cross-environment.md` — neutral defaults + per-OS deltas table (criterion 6). |
| New | `Conformance (matrix)` job in `.github/workflows/ci.yml` (`fail-fast: false`). |
| Modified | `scripts/sidecar-bundle.sh` — shells out to `sidecar-target.mjs` instead of inlining a Bash `case` block (eliminates the bundler-vs-test divergence that let #2010 sit there). |
| Modified | `scripts/run-tests.mjs` + `package.json` — wire the new `conformance` suite. |

## What Slice B (NOT in this plan) covers

- Booting the packaged sidecar on each OS
- Real PTY spawn round-trip
- Installer / MSI / DMG / AppImage end-to-end

Slice B will be tracked separately; it depends on Slice A's scaffolding (`sidecar-target.mjs`, `expectOrSkip`, the matrix job shell) being in place first.

## Acceptance criteria mapping (from #1990)

| Criterion | Where in the plan |
|---|---|
| 1. Matrix across Windows / macOS / Linux | Task 7 |
| 2. Shared conformance/contract suite | Tasks 4–6 |
| 3. Cover per-OS differences (paths, env, line endings, spawn) | Tasks 4–6 |
| 4. One neutral baseline job | Existing `Frontend build` stays unchanged; documented in Task 7 + 8 |
| 5. Explicit skips, no silent short-circuits | Task 3 + usage in 4, 5 |
| 6. Neutral defaults + per-OS deltas documented | Task 8 |

## Disposition after Slice A ships

Recommendation in the plan's Self-Review: close #1990 as Slice A complete, open a fresh tracking issue for Slice B (different acceptance criteria — packaging/runtime, not pure-function conformance).

Refs #1990.

### DIFF
--- a/docs/cross-os-conformance-plan.md
+++ b/docs/cross-os-conformance-plan.md
@@ -1,0 +1,983 @@
+# Cross-OS Conformance Suite (Slice A) Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Stand up a CI matrix job (`ubuntu-latest` / `windows-latest` / `macos-latest`) that runs a shared **conformance suite** over the platform-divergent helpers we already have, so the OS-specific code paths get exercised on the real OS instead of via simulated `process.platform = "win32"` unit branches. The suite must catch the exact bug classes that produced #2010 (sidecar prune dropping a runtime native dep) and #2011 (Windows `.cmd` shim spawn EINVAL), with explicit reasoned skips (never silent short-circuits) when a platform genuinely can't run a check.
+
+**Architecture:**
+
+- Add one new GitHub Actions job (`Conformance (matrix)`) running on all three OS images via a job matrix. It's separate from the existing single-OS `Frontend build` job — that job stays the "platform-agnostic baseline" (criterion 4).
+- Add one new conformance test file `scripts/conformance-cross-os.test.mjs` that *only* contains assertions which (a) genuinely vary per OS or (b) prove a per-OS branch loads from the real OS. It calls into a tiny new pure-function module `scripts/sidecar-target.mjs` that extracts the platform→native-package mapping currently inlined in `scripts/sidecar-bundle.sh`, so the test can invoke the same logic the bundler does.
+- Add an explicit-skip helper (`expectOrSkip(reason)`) that fails the test if used without a stringly reason, so "skipped" runs always carry a documented justification.
+- Add a docs page `docs/cross-environment.md` pinning the neutral defaults (ports/paths/config) and per-OS deltas, linked from the conformance suite as the source of truth.
+- Wire the new test into `scripts/run-tests.mjs` (`conformance` suite) so the `check:tests-wired` guard accepts it and so the matrix job can invoke it with `pnpm test:conformance`.
+
+**Tech Stack:** Node 22, `node:test` style of `node --experimental-strip-types` runner already used by the repo (`assert/strict`, top-level await), GitHub Actions matrix syntax already used in `.github/workflows/release.yml`, plain Bash for `scripts/sidecar-bundle.sh` integration.
+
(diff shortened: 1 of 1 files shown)


BunsDev:
Closing — pivoting to a release stamp instead. Plan branch will be removed; we'll revisit Slice A when execution slot opens up.

PR #2085, opened 2026-06-28
[codex] Fix digest carousel on touch devices

BunsDev:
## Summary
- disables the auto-scrolling home digest marquee on touch/coarse-pointer devices
- falls back to manual horizontal scrolling with momentum scrolling on iOS
- adds a source assertion covering the touch/coarse-pointer fallback

## Why
Touch devices cannot reliably use hover to pause the marquee, so the digest could become an unpausable moving target. This keeps the desktop marquee behavior while matching the existing reduced-motion fallback on touch.

## Validation
- node --experimental-strip-types src/components/home-digest-carousel.test.ts
- pnpm test:app
- pnpm check:tests-wired
- pnpm typecheck
- git diff --check

### DIFF
--- a/src/components/home-digest-carousel.test.ts
+++ b/src/components/home-digest-carousel.test.ts
@@ -47,6 +47,11 @@ assert.match(
   /@media \(prefers-reduced-motion: reduce\)[\s\S]*?animation: none/,
   "reduced-motion disables the auto-scroll",
 );
+assert.match(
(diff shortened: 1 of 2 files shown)


BunsDev:
Closing as superseded by #2086, which landed the same touch/coarse-pointer digest-carousel fix on main.

PR #2140, opened 2026-06-30
Polish command center controls across Home, Chat, and Code

BunsDev:
## Summary
- adds a shared command-control model for runtime/model/thinking/speed behavior
- polishes Home as the primary command center and threads Home-started thinking/speed controls into Chat
- keeps Code on ChatSurface, adds compact Quick Chat controls, and locks parity with regression coverage
- fixes server hub status/auth so mobile-access tokens are sent and unauthorized responses are not reported as offline
- simplifies the Tauri desktop settings header row

## Tests
- pnpm check:tests-wired
- pnpm exec tsc --noEmit --pretty false
- pnpm test:app
- cargo check --manifest-path src-tauri/Cargo.toml
- focused guards for daemon status/settings shell/command-control handoff paths

### DIFF
--- a/docs/superpowers/plans/2026-06-30-home-chat-code-command-center.md
+++ b/docs/superpowers/plans/2026-06-30-home-chat-code-command-center.md
@@ -1,0 +1,1148 @@
+# Home Chat Code Command Center Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Build the shared command-control contract approved in `docs/superpowers/specs/2026-06-30-home-chat-code-command-center-design.md`, polish Home as the primary launch surface, and keep Chat, Code, and Quick Chat switching behavior consistent.
+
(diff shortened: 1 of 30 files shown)


BunsDev:
🔴 **Heads-up for whoever's iterating on the command-control chat handoff** (cc @BunsDev)

The `Frontend build` required check is red, blocking merge. Everything else (CodeQL, E2E, Rust, Cross-environment, Sidecar) is green.

**Failure:** `pnpm test:app` → `src/components/home-chat-handoff.test.ts`

The test expects the chat-handoff call to thread the new args in this shape:

```
newChat(
  pendingChatAction.projectRoot ?? undefined,
  pendingChatAction.initialPrompt ?? undefined,
  pendingChatAction.familiarId,
)
```

…but the current source (shell/workspace handoff path) doesn't match that signature, so the regex assertion fails.

Note: `home-chat-handoff.test.ts` is currently modified-but-uncommitted in a local working tree, so a fix looks like it's already in flight — this is just to flag that it hasn't landed on the PR head yet (`1341f30f`).

Unrelated context: the home-composer balanced-row layout work (`home-composer.css`, commit `1341f30f`) is fine — its own pinning tests pass; this red check is purely the chat-handoff assertion. A background watcher is re-checking #2140 and will confirm once Frontend build flips green.

PR #2177, opened 2026-06-30
fix(code): cache session-pins snapshot (stop useSyncExternalStore infinite loop)

BunsDev:
**Bugfix** for the Code sidebar shipped in #2162.

`getPinnedSessionIds()` re-parsed localStorage on every call, returning a fresh array reference each render. `useSyncExternalStore` requires a stable `getSnapshot`, so React threw **"The result of getSnapshot should be cached to avoid an infinite loop"** and the Code sidebar rendered a full error overlay on mount — for *every* visit (the no-pins path also returned a fresh `[]`).

**Fix:** cache the parsed array until a mutation (or cross-tab `storage` event) invalidates it — mirrors the proven `familiar-quick-switch` store. Regression test now asserts snapshot reference stability (`getPinnedSessionIds() === getPinnedSessionIds()` until a mutation).

**How it slipped through:** build, source-text, and CI all passed because none exercise the actual React render. Caught by visually verifying the Code surface in the running app (screenshot showed the overlay).

### DIFF
--- a/src/lib/session-pins.test.ts
+++ b/src/lib/session-pins.test.ts
@@ -32,6 +32,15 @@ assert.deepEqual(getPinnedSessionIds(), ["s1", "s2"], "second pin appended in or
 
 toggleSessionPin("s1");
 assert.deepEqual(getPinnedSessionIds(), ["s2"], "toggle removes existing id");
+
+// Snapshot stability: useSyncExternalStore requires getSnapshot to return the
+// SAME reference until a mutation, else React throws "getSnapshot should be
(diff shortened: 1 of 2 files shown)


BunsDev:
Superseded by #2175 ("Fix Code-surface crash: stable getPinnedSessionIds snapshot"), which already landed an equivalent snapshot-cache fix on main (memoizes on the raw localStorage string). Verified main's version satisfies the useSyncExternalStore cached-snapshot contract. Closing as duplicate.

PR #2187, opened 2026-07-01
Improve group chat first-pass replies

Aimplemented:
## Summary
- change default full-coven group chat from sequential peer relay to parallel independent first-pass replies
- add `renderCovenRoundtablePrompt` to give each familiar compact roster context plus explicit answer-as-yourself framing
- update group chat tests to lock in no default peer transcript relay, while keeping the existing relay helper available for future explicit debate/synthesis modes

## Why
Sequentially feeding earlier familiar replies into later familiar prompts makes group chat converge into a blended voice. The default roundtable pass should preserve distinct familiar perspectives first, then synthesize only in an explicit mode later.

## Tests
- `node --experimental-strip-types --test src/lib/group-chat.test.ts`
- `node --test src/components/group-chat-view.test.ts`
- `corepack pnpm exec tsc --noEmit`
- `corepack pnpm test:app` (510 app test files)

### DIFF
--- a/src/components/group-chat-view.test.ts
+++ b/src/components/group-chat-view.test.ts
@@ -21,11 +21,11 @@ test("GroupChatView broadcasts via /api/chat/send and reuses pure helpers", () =
   // A Stop control aborts the in-flight broadcast.
   assert.match(view, /abortRef\.current\?\.abort\(\)/, "Stop aborts the broadcast");
   // Injects the coven roster into each send so a familiar knows who else is present.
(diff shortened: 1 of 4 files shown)


BunsDev:
Thanks so much for this, @Aimplemented 🙏 This work shipped in #2206 (parallel first-pass replies + roundtable framing, closing #2188). Your original PR was re-landed through an internal branch during merge, so the commit trailer didn't link back to your profile — that's on us. We've added you to [CONTRIBUTORS.md](https://github.com/OpenCoven/coven-cave/blob/main/CONTRIBUTORS.md) (via #2436) with proper credit. Really appreciate the contribution.

PR #2245, opened 2026-07-02
test(board): stage both attachment fixtures in e2e

BunsDev:
## Summary
- stage both board attachment fixtures in the E2E composer flow
- assert both serialized attachment names in the board POST body

## Test Plan
- pnpm exec playwright test tests/board-attachments.spec.ts --project=desktop
- pnpm typecheck
- git diff --check


BunsDev:
Closing this duplicate branch-shape PR; reopening the same E2E review fix from current main so the diff only contains the follow-up change.

PR #2255, opened 2026-07-02
chore(deps): bump 15 patch/minor dependencies + pin enhanced-resolve

BunsDev:
Routine dependency maintenance — **patch/minor bumps only**. No major-version migrations.

### Updated (15)
| Package | From | To |
|---|---|---|
| next | 16.2.6 | 16.2.9 |
| react / react-dom | 19.2.4 | 19.2.7 |
| @playwright/test | 1.60.0 | 1.61.1 |
| shiki | 4.2.0 | 4.3.0 |
| mermaid | 11.15.0 | 11.16.0 |
| react-resizable-panels | 4.11.2 | 4.12.0 |
| @xyflow/react | 12.11.0 | 12.11.1 |
| tailwindcss (+ /browser, /postcss) | 4.3.0 | 4.3.1 |
| @tauri-apps/api | 2.11.0 | 2.11.1 |
| @tauri-apps/cli | 2.11.2 | 2.11.4 |
| @next/bundle-analyzer | 16.2.6 | 16.2.9 |
| esbuild | 0.28.0 | 0.28.1 |

### Held back / pinned
- **sharp** 0.34.5 → 0.35.2 is held back: the first attempt included it, but it broke the Windows sidecar runtime (`Failed to load external module sharp-*`, avatar endpoint HTTP 500). macOS/Linux sidecar were unaffected. The 0.35 native-binary packaging change needs a dedicated follow-up against the esbuilt sidecar bundle.
- **enhanced-resolve** is pinned to 5.23.0 through `pnpm.overrides` so the Tailwind 4.3.1 bump does not downgrade module resolution from the previous lockfile.
- **typescript** 5.9.3 → 6.0.3 and **@types/node** 20 → 26 are major bumps; dedicated migration.

Note: `minimumReleaseAge` (3 days) held a few packages one patch behind latest (e.g. next 16.2.10, tailwind 4.3.2) — expected.

### Verification
- ✅ `corepack pnpm install --frozen-lockfile`
- ✅ `corepack pnpm typecheck`
- ✅ `corepack pnpm test:supply-chain`
- Prior revision: `check:tests-wired` (685), `test:app` (512), `test:api` (111), `test:mobile` (65), `build` + `bundle-budget`

Regenerated PWA icons (sharp re-encoding) were excluded — `prebuild` regenerates them each build.

PR #2263, opened 2026-07-02
chore(deps): bump sharp 0.34.5 → 0.35.3 (fixes Windows sidecar)

BunsDev:
## What

Bumps `sharp` **0.34.5 → 0.35.3**, un-holding the major that was blocked since #2255.

## Why it was held

sharp 0.35.0–0.35.2 broke the required **`Sidecar runtime (windows-latest)`** check. In the esbuilt, spawned sidecar the avatar route 500'd with:

```
⨯ Error: Failed to load external module sharp-<hash>:
   TypeError: Cannot read properties of undefined (reading 'output')
```

**Root cause (upstream, not our bundle):** sharp 0.35.x rewrote native-binary loading for bundler support (static paths #4380, stub binaries #4543). On Windows the `.node` couldn't resolve its bundled `libvips-cpp-*.dll` in that context, so `sharp.format()` returned a table without `.heif` and `dist/utility.cjs` threw at **require-time** (`format.heif.output.alias = …`), which Next surfaces as "Failed to load external module sharp". macOS/Linux were unaffected — only the Windows leg caught it.

## Why 0.35.3 fixes it

0.35.3 upgrades to **sharp-libvips v1.3.2**, whose changelog entry is verbatim:

> Improve code bundler support by resolving path to libvips binary.

That's the exact failure class. **No change to our sidecar is needed** — `scripts/sidecar-bundle.sh` and `scripts/sidecar-target.mjs` are unchanged; the win32 mapping (keep `@img/sharp-win32-x64`, libvips compiled in) still holds.

## Release-age hold

0.35.3 is <72h old, inside the repo's `minimumReleaseAge: 4320`. The lockfile was regenerated with a **temporary, uncommitted** `minimumReleaseAgeExclude: [sharp, "@img/*"]`. CI installs with `--frozen-lockfile` (no re-resolution), so the hold does not apply here and the exclude is not part of the diff.

## Verification

- ✅ `tsc --noEmit` clean (with TS 6.0.3)
- ✅ `scripts/sidecar-bundle.sh` builds; the `require('sharp')` sanity check passes
- ✅ `pnpm test:sidecar-runtime` renders the avatar end-to-end on macOS
- 🎯 **The `Sidecar runtime (windows-latest)` leg on this PR is the real proof.**

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,7 +81,7 @@
     "react-colorful": "5.7.0",
     "react-dom": "19.2.7",
     "react-resizable-panels": "4.12.0",
-    "sharp": "0.34.5",
+    "sharp": "0.35.3",
     "shiki": "4.3.0",
     "sucrase": "3.35.1",
     "ws": "8.21.0",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -120,8 +120,8 @@ importers:
         specifier: 4.12.0
         version: 4.12.0(react-dom@19.2.7(react@19.2.7))(react@19.2.7)
       sharp:
-        specifier: 0.34.5
-        version: 0.34.5
+        specifier: 0.35.3
+        version: 0.35.3(@types/node@24.13.2)
       shiki:
         specifier: 4.3.0
         version: 4.3.0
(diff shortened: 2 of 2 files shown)


BunsDev:
**Closing — the bump was the wrong fix. Root cause is version *skew*, not a `0.35.3` defect, and `0.34.5` is not vulnerable.**

### Symptom
Bumping `sharp` `0.34.5` → `0.35.3` turned `Sidecar runtime (windows-latest)` red:
```
Failed to load external module sharp … Cannot read properties of undefined (reading 'output')
```
Thrown at module-load in `sharp/dist/utility.cjs` (`format.heif.output.alias = ['avif','heic']`) because the win32 native returned a `format()` table with **no `heif`**.

### Root cause — version skew with Next
`next@16.2.9` hard-depends on `sharp@0.34.5`. Bumping our **direct** dep to `0.35.3` made **two** sharp versions coexist in the lockfile:

```
$ pnpm why sharp
sharp@0.34.5  └─┬ next@16.2.9        ← Next's pinned copy
sharp@0.35.3  └── coven-cave         ← our direct bump
```

The sidecar's hoisted, dereferenced, pruned staging install (`scripts/sidecar-bundle.sh`, `--node-linker=hoisted`) landed **`0.34.5`'s** `@img/sharp-<target>` native at top-level while `sharp` JS resolved to **`0.35.3`** → native/JS ABI mismatch. macOS tolerated it (heif still present); Windows did not (mismatch drops `heif` from the format table), so **only the win32 leg** crashed.

### No security driver
The only `sharp` advisories are `< 0.32.6` (CVE-2023-4863, libwebp) and `< 0.30.5` (postinstall command injection). `0.34.5` is unaffected. And `sharp@0.35.3` was published ~24h ago — inside our 72h `minimumReleaseAge` supply-chain hold.

### Decision — stay on `0.34.5`
Align with Next's own pin. This is the pre-bump state (Windows sidecar was green), needs no hold override, ships no 24h-old release, and carries zero Next image-optimization compat risk. There is nothing to land, so this PR is closed.

### Guidance to avoid re-introducing this
Don't bump `sharp` above the version `next` pins **without** adding a repo-wide `pnpm.overrides.sharp` to force Next's transitive copy to the same version. Otherwise the hoisted sidecar bundle can pick the wrong `@img` native, and **only** the Windows sidecar leg will catch it.

PR #2323, opened 2026-07-03
Drag-to-split: dragging a pane wide collapses & promotes it

BunsDev:
The resizable split divider already closed the secondary when dragged past the near edge; this adds the mirror on the far edge — dragging a pane wide collapses the *other* pane. A page split promoted this way becomes the sole surface (via setMode); a companion (no primary mode) just sits at max width. The live guide shows a "Fill" chip in the far-edge zone, mirroring the "Close" chip near the near edge.

Two fixes were needed to make this (and the pre-existing snap/close) actually reachable at runtime:

1. `.ui-sep-handle` rendered at height 0 (a block child that never stretched), so the divider handle was unclickable — the component's promised hit area didn't exist. It now fills the separator.

2. react-resizable-panels' <Separator> owns the pointer drag and swallows the pointerdown, so the handle's onMouseDown never fired and the snap/close release logic was dead. Release is now detected from a global pointerup + RRP's live onResize ratio, independent of who owns the pointerdown.

Symmetric edges: panels clamp to 10%/90%; close fires 0.06 above the floor, collapse 0.06 below the ceiling.

## Changes

-

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -2300,10 +2300,15 @@ textarea.required-inputs-control {
 }
 .ui-sep-handle--col {
   width: 1px;
+  /* Fill the separator's height so the handle is actually clickable — a block
+     child otherwise collapses to 0 height, leaving the drag handler (and its
+     hover affordance / hit area) unreachable. */
+  height: 100%;
   cursor: col-resize;
 }
 .ui-sep-handle--row {
   height: 1px;
+  width: 100%;
   cursor: row-resize;
 }
 .ui-sep-handle::before {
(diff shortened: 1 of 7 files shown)


BunsDev:
Superseded by #2280 ("Drag-to-split: dragging a pane wide collapses & promotes it"), which already merged to main as `70dee2c1`. This PR's head commit `451dd454` is the same original work — rebasing it onto current main collapses to zero new commits (the feature is already live). Closing to keep the queue clean ahead of accepting external PRs.

PR #2363, opened 2026-07-04
feat(chat): show each thread's project icon in the Recent sidebar view

BunsDev:
## What
The chat sidebar's **Recent** view (the time-bucketed thread list) now shows each thread's **project icon** on its row — the same `ProjectAvatar` tile the *Projects* view already renders in its folder header.

The grouped **By project** view conveys a thread's project through its folder header, but **Recent** is a flat list that mixes threads from every project with no project cue. Now each Recent row leads with:
- the project's **`ProjectAvatar`** (the user-uploaded image, or the deterministic tinted monogram tile — identical to everywhere else the project appears) for a **registered** project, or
- the **dashed folder** glyph (`ph:folder-simple-dashed`) for an unregistered cwd / no-project thread, matching the app's existing "outside a project" visual language.

## How
- A `groupBySession` memo maps each session → its project group off the existing `deriveChatProjectGroups(...)` result (which is **override-applied**, so a thread's per-user project override is honored).
- `ThreadRow` gains an optional `project?: ChatProjectGroup | null` prop; only the Recent view passes it (the Projects view keeps its header-only treatment). Existing status dot, title, time, and hover actions are unchanged.
- New `.cnav__thread-project` CSS (a `flex-shrink:0` inline-flex slot in the existing `gap:8px` row).

## Verified visually
Built + drove the chat sidebar with mocked sessions/projects (Playwright). Recent view rendered: `Refactor auth flow`/`Write API docs` → **AL** tile (registered Alpha), `Fix eslint config` → **BE** tile (registered Beta), `Tune the ranker` (unregistered cwd) + `Scratch experiment` (no project) → dashed folder. Tiles align cleanly, status dots preserved.

## Test
- `pnpm test:app` — all app test files pass; new source-text pins in `chat-sidebar-wiring.test.ts` (the group lookup, the Recent-row `project` wiring, the `ProjectAvatar` render). `tsc --noEmit` clean.

## Notes / possible follow-up
The tile is decorative (`aria-hidden`, as everywhere `ProjectAvatar` is used) with a `title` tooltip; the thread title remains the row's accessible name. A visually-hidden project name per row could be added if we want AT to announce the project in the flat list.

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -1039,6 +1039,16 @@ tr[role="checkbox"]:focus-visible {
   flex-shrink: 0;
   color: var(--text-muted);
 }
+/* Per-row project tile in the flat Recent view (avatar for a registered project,
+   dashed folder for an unregistered cwd / no-project thread). */
+.cnav__thread-project {
+  flex-shrink: 0;
+  display: inline-flex;
+  align-items: center;
+}
+.cnav__thread-project--dir {
+  color: var(--text-muted);
+}
 .cnav__group-name {
   flex: 1;
   min-width: 0;
--- a/src/components/chat-sidebar-wiring.test.ts
+++ b/src/components/chat-sidebar-wiring.test.ts
@@ -108,5 +108,10 @@ assert.match(
   /project\.projectId \? \([\s\S]{0,160}<ProjectAvatar/,
   "a registered project renders a ProjectAvatar tile in the row",
 );
+assert.match(
+  chatSidebar,
+  /<span className="sr-only">, \{folderLabel\(project\)\}<\/span>/,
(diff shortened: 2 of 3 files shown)


BunsDev:
Superseded by #2360 (`feat(chat): project identity tiles on Recent sidebar rows`, merged as `9ed2d599`). Main already renders the per-row project tile in the Recent view via the same `ProjectAvatar` + `sessionProjectById` + `ChatProjectGroup`/`folderLabel` path — this branch's earlier `groupBySession`/`projectId` variant conflicts on `chat-sidebar.tsx` + `chat-sidebar-wiring.test.ts` and would only re-churn identical behavior. Closing to keep the branch count down; the feature is live on main.

PR #2417, opened 2026-07-05
fix(proxy): guard tokenless GET webhooks

BunsDev:
### Motivation
- Prevent cross-site GET requests from triggering state-changing production webhook flows when the server is running in tokenless Tailscale/native mode and both `Origin` and `Referer` headers are omitted.
- Close a high-severity CSRF hole introduced by allowing the same mutating handler to be exported for `GET` without a stricter source check in tokenless mode.

### Description
- Add a narrow helper `isProductionWebhookGet` to detect production webhook GET requests at `/api/flows/webhook` and `/api/flows/webhook/*`.
- In the proxy CSRF/source gate, read `Origin` and `Referer` into local variables and reject requests that fail the same-origin checks as before.
- For tokenless tailnet mode (`COVEN_CAVE_TAILNET_TRUST=1`) when no sidecar token is configured, additionally reject production webhook `GET` requests that have both `Origin` and `Referer` absent by returning a `403` with message `missing request source`.
- Update the middleware source-contract test in `src/middleware.test.ts` to pin the new header caching and the dedicated tokenless-tailnet GET-webhook guard.

### Testing
- Ran typecheck with `pnpm exec tsc --noEmit`, which succeeded.
- Executed `node src/middleware.test.ts` and `node src/proxy-behavior.test.ts`, both of which completed successfully (tests passed with only Node module-type warnings).
- Existing behavior for header-token-authenticated requests and other API requests was preserved and asserted by the updated tests.

### DIFF
--- a/src/middleware.test.ts
+++ b/src/middleware.test.ts
@@ -20,9 +20,13 @@ assert.match(source, /req\.headers\.get\("host"\)/, "middleware should reject un
 assert.match(source, /const requestHost = req\.headers\.get\("host"\)/, "proxy should capture the forwarded request host once");
 assert.match(source, /isAllowedApiHost\(requestHost, mobileAccessAuthenticated \|\| tailnetTrusted\)/, "valid mobile access or tailnet-trust should satisfy the API host gate");
 assert.match(source, /const tailnetTrusted = process\.env\.COVEN_CAVE_TAILNET_TRUST === "1"/, "tokenless app mode (COVEN_CAVE_TAILNET_TRUST) should relax the host gate for tailnet-forwarded requests");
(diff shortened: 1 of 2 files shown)


BunsDev:
🔒 Security review (comment; can't self-approve): **Sound, tightly-scoped CSRF hardening.** Tokenless Tailscale + no sidecar secret previously accepted state-changing GET webhooks (`/api/flows/webhook*`) with neither Origin nor Referer — the `Referrer-Policy: no-referrer` cross-site bypass. Now requires at least one same-origin header. Scope correct (only that path, only tailnet+tokenless). Clean header refactor. LGTM.

PR #2418, opened 2026-07-05
fix(onboarding): require manual sudo for npm installs

BunsDev:
### Motivation
- Prevent a privilege-escalation vector where the onboarding API could spawn `sudo -n npm install -g ...` from the server process when passwordless sudo or cached creds were present.
- Ensure network-accessible routes never perform non-interactive, root-level package installs without explicit operator action.

### Description
- Remove automatic elevation logic by deleting `passwordlessSudoAvailable()` and by avoiding a `sudo` spawn plan in `spawnPlanFor` when global npm dirs are not writable; the function now returns `{ sudoRequired: true }` in that case instead of a `sudo` command plan in `src/app/api/onboarding/install/route.ts`.
- Update the `sudoRequired` response hint to make it explicit that the operator must run the `sudo npm install -g <pkg>` command manually from a terminal.
- Add a source-level regression assertion to `src/app/onboarding-install-route.test.ts` that prevents reintroduction of auto-elevation by checking the route does not reference `sudo`, `passwordlessSudoAvailable`, or `-n`.

### Testing
- Ran the route source test with `node src/app/onboarding-install-route.test.ts`, which passed.
- Ran type checking with `pnpm typecheck` (`tsc --noEmit`), which succeeded.

### DIFF
--- a/src/app/api/onboarding/install/route.ts
+++ b/src/app/api/onboarding/install/route.ts
@@ -166,21 +166,6 @@ async function npmGlobalDirsWritable(npm: string): Promise<boolean> {
   return true;
 }
 
-/** `sudo -n true` succeeds only when sudo needs no password (cached creds or
- *  a NOPASSWD rule). We never spawn a password-prompting sudo from this
- *  non-interactive server context — with no TTY it would hang forever. */
-async function passwordlessSudoAvailable(): Promise<boolean> {
-  try {
-    await execFileAsync("sudo", ["-n", "true"], {
-      env: covenSpawnEnv(),
-      timeout: 3000,
-    });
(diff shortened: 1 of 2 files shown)


BunsDev:
🔒 Security review (comment; can't self-approve): **Correct and conservative.** Removes automatic `sudo -n` elevation from a non-interactive API route — a global npm install runs arbitrary lifecycle scripts + writes system paths, so silent priv-esc from a web request is the risk. User-owned prefixes (nvm/fnm/brew) still one-click; only system-owned prefixes need operator sudo. Regression test guards against sudo reappearing. LGTM.

PR #2419, opened 2026-07-05
fix(board): validate assigned task chat project root

BunsDev:
### Motivation
- Prevent caller-forged `projectRoot` values from redirecting task-linked daemon sessions into projects the assigned card/familiar shouldn't access by resolving assigned project roots server-side and enforcing project grants.

### Description
- Resolve the runtime `projectRoot` from the registered project when a card has `projectId` by using `projectById` and `loadProjects` instead of trusting `body.projectRoot`.
- Reject requests where a provided `body.projectRoot` does not match the assigned project's normalized root with a `403` response, and enforce the familiar's `session-launch` grant via `assertProjectAccess`.
- Preserve existing daemon-launch and worktree isolation logic while ensuring persisted `cwd` on the card uses the server-resolved root for project-assigned cards.
- Update source-contract tests in `src/components/task-chat-cwd.test.ts` and `src/components/board-chat-link.test.ts` to require server-side project resolution and authorization behavior.

### Testing
- Ran the component source-contract checks with `node --experimental-strip-types src/components/task-chat-cwd.test.ts` and `node --experimental-strip-types src/components/board-chat-link.test.ts`, both succeeded.
- Type-checked with `pnpm typecheck` and verified wired tests with `pnpm check:tests-wired`, both succeeded.
- Executed the full app test suite with `pnpm test:app`, and the app tests passed (all automated checks green).

### DIFF
--- a/src/app/api/board/[id]/chat/route.ts
+++ b/src/app/api/board/[id]/chat/route.ts
@@ -88,12 +88,14 @@ export async function POST(
       return NextResponse.json({ ok: false, error: "assigned project root is invalid" }, { status: 409 });
     }
 
-    const requestedProjectRoot = body.projectRoot ? normalizeProjectRoot(body.projectRoot) : null;
-    if (requestedProjectRoot && requestedProjectRoot !== projectRoot) {
-      return NextResponse.json(
-        { ok: false, error: "project root does not match assigned task project" },
-        { status: 403 },
-      );
+    if (body.projectRoot !== undefined && body.projectRoot !== null) {
+      const requestedProjectRoot = normalizeProjectRoot(body.projectRoot);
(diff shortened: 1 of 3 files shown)


BunsDev:
🔒 Security review (comment; can't self-approve): **Most important of the batch — approving in spirit.** Resolves assigned project root server-side from `card.projectId` (not client-supplied `body.projectRoot`), rejects mismatched roots (403), and authorizes via `assertProjectAccess(..., 'session-launch')`. Closes a forged-projectRoot hole (familiar launched in arbitrary dir). I rebased onto current main (conflicted with #2308's UX rewrite of the same block); kept this PR's server-authoritative posture + main's stricter no-project 409; Copilot-autofix follow-up preserved. Both route tests pass against the resolved code + route typechecks clean.

PR #2433, opened 2026-07-06
fix: strip bundled GLib from AppImage so one build works on all Linux

CodeDoes:
## Problem

The Linux AppImage was built on **ubuntu-22.04** (GLib **2.72**), but system libraries loaded at runtime via `dlopen` — `libayatana-appindicator3` (tray icon) and GVFS GIO modules — are compiled against GLib **2.80+** and require symbols the bundled GLib lacks:

| Symbol | Required by | Added in |
|---|---|---|
| `g_once_init_leave_pointer` | `libayatana-ido3.so.0` | GLib 2.80 |
| `g_task_set_static_name` | `libgvfscommon.so` (GIO) | GLib 2.68 |

Since the AppImage resolves these to the **bundled** GLib first (via `RUNPATH`), the dlopen fails — tray icon panics and GVFS module errors are logged.

## Solution

**Stop bundling GLib in the AppImage.** A single build on `ubuntu-22.04` produces one AppImage for all of Linux. A post-processing step removes the bundled GLib libraries after the initial build, so the host system's GLib is used at runtime.

## Why one AppImage instead of two

The two-build approach (build on both 22.04 and 24.04) works but ships two AppImages, requires artifact disambiguation, per-Ubuntu renaming, and manifest preference logic — and still exposes the 22.04 build to the same crash for manual downloaders.

The strip-GLib approach produces **one AppImage** that works on **every Linux distro** because it always uses the host's GLib — which is always the right version for the host's dlopen'd libraries. The GLib exclusion reduces the bundle size by ~8 MB and eliminates the symbol mismatch at the root.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -88,14 +88,8 @@ jobs:
             arch_suffix: aarch64
           - platform: ubuntu-22.04
             family: linux
-            label: Linux (AppImage, ubuntu-22.04)
+            label: Linux (AppImage)
             args: '-vv --bundles appimage --config ''{"bundle":{"createUpdaterArtifacts":false}}'''
-            dist_suffix: ubuntu-22.04
-          - platform: ubuntu-24.04
-            family: linux
-            label: Linux (AppImage, ubuntu-24.04)
-            args: '-vv --bundles appimage --config ''{"bundle":{"createUpdaterArtifacts":false}}'''
-            dist_suffix: ubuntu-24.04
           - platform: windows-latest
             family: windows
             label: Windows (MSI)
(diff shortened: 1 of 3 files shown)


BunsDev (changes requested):
I would request changes before merging.

Findings:

1. The AppImage rename happens after updater signing, so the final suffixed assets do not have matching `.sig` assets. `scripts/generate-latest-json.mjs` only includes artifacts where `<artifact>.sig` exists with the exact same basename, so `CovenCave_*_ubuntu-24.04.AppImage` would be skipped by `latest.json`. Rename before signing, or re-sign/re-upload the signature after the rename.

2. The manifest generator still has only one `linux-x86_64` slot and picks the first signed `.AppImage`. With both Ubuntu builds present, auto-update can select the ubuntu-22.04 build and miss the intended GLib-symbol fix. Prefer the `ubuntu-24.04` AppImage for the updater manifest, while keeping both assets available for manual downloads if desired.

3. `scripts/release-macos-signing.test.mjs` still asserts the old ubuntu-22.04-only workflow shape, so CI will fail once the fork run is approved. Update the workflow regression test to cover the new `matrix.family == 'linux'` conditions, both Ubuntu matrix entries, and the rename-before-signing order.

I patched these locally in the maintainer checkout as a reference fix: move Linux AppImage suffixing before signing, re-upload the suffixed AppImage and race-tolerantly delete the unsuffixed asset, make `generate-latest-json.mjs` prefer the signed `ubuntu-24.04` AppImage, and update/add tests for those invariants.

CodeDoes:
Rebased and replaced the two-Ubuntu-build approach with a single-build approach that strips bundled GLib from the AppImage after build. This makes all three review findings moot:

**Finding 1 (rename before signing):** Not applicable — there is no rename/suffix step. Only one AppImage per release, so no disambiguation is needed. The strip step runs after initial signing and re-signs the repacked artifact.

**Finding 2 (manifest prefers ubuntu-24.04):** Not applicable — there is only one `.AppImage` on the release. `generate-latest-json.mjs` uses a simple predicate `(n) => n.endsWith(".AppImage")` with no preference callback.

**Finding 3 (tests):** Updated `release-macos-signing.test.mjs` to test the GLib-strip pipeline instead of the two-build path. All 9 tests pass.

**Why one build is better:**
- Ships one AppImage that works on every Linux distro (uses host GLib at runtime)
- No per-Ubuntu matrix entries, artifact disambiguation, or manifest preference logic
- Reduces AppImage size by ~8 MB (GLib libraries removed from bundle)
- Root-cause fix: the symbol mismatch exists because the bundle GLib version differs from the host GLib version. By not bundling GLib at all, the host's GLib is always used — which always matches the host's dlopen'd libraries.

BunsDev (changes requested):
I still need changes before this can land.

Blocking issue: the new single-build GLib-strip flow repacks and re-signs the local AppImage, but it only uploads `${APPIMAGE}.sig` after repacking. The AppImage asset itself was already uploaded earlier by `tauri-action`, before the strip step. So the release would still serve the original bundled-GLib AppImage while publishing a signature for the modified local file. That both misses the runtime fix and risks a signature/artifact mismatch.

Please upload the repacked AppImage with `gh release upload "$RELEASE_TAG" "$APPIMAGE" --clobber` after `appimagetool` succeeds and before/with the post-strip signature upload, and add a workflow regression assertion for the repacked AppImage upload.

The redesign does make the earlier two-Ubuntu artifact naming findings moot, but this new upload gap is still release-blocking.

BunsDev:
Re-landed through internal PR #2478 with the remaining AppImage upload blocker fixed, required checks green, and CodeDoes credited in the squash merge trailer and CONTRIBUTORS.md. Closing this fork PR as superseded by #2478. Thank you for the release fix.

PR #2435, opened 2026-07-06
feat: add opencode as a compatibility adapter in web UI

CodeDoes:
Adds opencode as a compatibility adapter in the coven-cave web UI, so users can select and configure it alongside Codex, Claude Code, Hermes, and OpenClaw.

## Changes
- Add opencode entry to COMPATIBILITY_ADAPTERS with install hint
- Add opencode to RUNTIME_MODEL_CATALOG with OpenAI model support
- Update adapter sort ranking to include opencode

### DIFF
--- a/src/lib/harness-adapters.ts
+++ b/src/lib/harness-adapters.ts
@@ -68,6 +68,16 @@ export const COMPATIBILITY_ADAPTERS: CompatibilityAdapter[] = [
     source: "bundled",
   },
   {
+    id: "opencode",
+    label: "OpenCode",
+    binary: "opencode",
+    chatSupported: true,
+    versionArgs: ["--version"],
+    installHint:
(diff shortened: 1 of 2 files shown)


BunsDev (changes requested):
I found a test break that needs to be fixed before this lands.

`node --experimental-strip-types src/lib/harness-adapters.test.ts` fails because the test still expects the exact built-in adapter list to be `["codex", "claude", "hermes", "openclaw"]`, but this PR adds `opencode` between `claude` and `hermes`. Please update the adapter regression test(s) to include `opencode` and cover the intended ordering/defaults.

Observed failure:

```
AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

  [
    'codex',
    'claude',
+   'opencode',
    'hermes',
    'openclaw'
  ]
```

I stopped there because the first focused test already fails.

PR #2448, opened 2026-07-06
fix(layout): server-render boot scripts (React 19 warning) + guard growth analytics link

BunsDev:
Fixes two console/runtime errors seen in dev (Next 16.2.9 / React 19).

## 1. Console Error — "script tag while rendering React component"
`ThemeScript`, `DevCacheResetScript`, and `SidecarAuthBridge` rendered their boot
scripts through `next/script`, which is a **client component**. Under React 19 /
Next 16 that logs:

> Encountered a script tag while rendering React component. Scripts inside React
> components are never executed when rendering on the client. Consider using
> template tag instead.

All three now render as **plain server-rendered `<script>` tags** from the
server-component `RootLayout`. They execute at HTML parse — before hydration —
with **no client-render path**, so no warning. Behaviour is preserved:

- `theme-script` stays a render-blocking `<script src>` in `<head>` → flash-free theming.
- `sidecar-auth-bridge` / `dev-cache-reset` stay inline `<script>` at `<body>` top → `window.fetch` is patched before any app code runs.

This effectively restores the original pre-#2366 placement (the `next/script`
switch is what introduced the warning under Next 16).

## 2. Runtime TypeError — `Cannot read properties of null (reading 'familiar')`
`FamiliarGrowthView`'s header Analytics link dereferenced `selected.familiar.id`,
but `selected` is `null` when the familiar roster is empty (`… ?? reports[0] ?? null`).
Wrapped it in a `{selected ? … : null}` guard; the empty-roster state already
communicates "No familiars available."

## Tests
Updated the three co-located tests that pinned `import Script from "next/script"`
/ `strategy="beforeInteractive"` to assert the plain-`<script>` form instead.

- `theme-script`, `dev-cache-reset-script`, `sidecar-auth-bridge` (6/6), `familiar-analytics-navigation` (3/3) — all green.
- Full `node scripts/run-tests.mjs app` suite was green on the same changeset (519 files) before this was moved into an isolated worktree.

## Note for reviewers
The 6 boot-script conversions were also independently staged (identical approach,
uncommitted, no PR) in a concurrent `codex/boot-scripts-server-rendered` worktree.
This PR is the complete version and additionally carries the growth-view null
guard. If a duplicate boot-script PR shows up, close whichever is redundant.

### DIFF
--- a/src/components/dev-cache-reset-script.test.ts
+++ b/src/components/dev-cache-reset-script.test.ts
@@ -5,8 +5,8 @@ import { readFileSync } from "node:fs";
 const src = readFileSync(new URL("./dev-cache-reset-script.tsx", import.meta.url), "utf8");
 const layout = readFileSync(new URL("../app/layout.tsx", import.meta.url), "utf8");
 
-assert.match(src, /import Script from "next\/script"/, "dev cache reset uses Next Script, not a raw script tag");
-assert.match(src, /strategy="beforeInteractive"/, "dev cache reset should run before hydration");
+assert.doesNotMatch(src, /import Script from "next\/script"/, "dev cache reset must render a plain server <script>, not next/script (a client component → React 19 'script never executes on client' warning)");
+assert.match(src, /<script\b/, "dev cache reset renders a plain, server-rendered inline <script> that runs before hydration");
 assert.match(src, /process\.env\.NODE_ENV !== "development"/, "script should render only in development");
 assert.match(src, /navigator\.serviceWorker\.getRegistrations\(\)/, "script should inspect existing service workers before hydration");
 assert.match(src, /registration\.unregister\(\)/, "script should unregister stale service workers");
--- a/src/components/dev-cache-reset-script.tsx
+++ b/src/components/dev-cache-reset-script.tsx
@@ -1,5 +1,3 @@
-import Script from "next/script";
-
 const DEV_CACHE_RESET_SCRIPT = `
 (function () {
   try {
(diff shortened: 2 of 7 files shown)


BunsDev:
Closing in favor of #2453.

The boot-script portion of this PR (converting `theme-script`, `dev-cache-reset-script`, and `sidecar-auth-bridge` from `next/script` → plain server `<script>`) was fully superseded by #2447 (`021073e` "fix(layout): render boot scripts from document"), which already landed all three conversions on `main`. Rebasing this PR would leave those files as no-ops and produce a conflicting/empty diff.

The one piece of unique value here — the null-guard on the growth-view Analytics link (`familiar-growth-view.tsx`) — has been extracted into the clean, conflict-free #2453 off current `main`.

BunsDev:
Superseded and closed — no action needed. This combined PR was split into two focused PRs, both merged to `main`:

- **#2447** — boot scripts (`theme-script`, `dev-cache-reset`, `sidecar-auth-bridge`) rendered server-side instead of via `next/script` (fixes the React 19 "script tag while rendering React component" warning).
- **#2453** — `FamiliarGrowthView` header Analytics link guarded against a null `selected` (fixes `Cannot read properties of null (reading 'familiar')` on an empty familiar roster).

Both fixes were verified in the running app (dev mode): 0 boot-script warnings + theme boots flash-free; growth page renders the empty-roster state with the Analytics link correctly hidden, and shows the link again when a familiar exists. Branch `fix/boot-scripts-react19` has been deleted.

PR #2487, opened 2026-07-06
feat(home): show the project selector in the composer toolbar

BunsDev:
## What

The home composer auto-selected `projects[0]` but gave the user **no way to change it** — the project picker had been dropped in an earlier toolbar consolidation (`bf3a5d2c`), leaving its `hc-project-selector` styling behind as dead CSS. This restores it so a new chat's project is a visible, selectable control, mirroring the chat composer.

**Toolbar now:** 📎 · 🎙 · ⚙ · Chat/Task · **Nova** (familiar) · **📁 No project** (new) · … · ✦ · ↑

## How

- Render the shared `ProjectPicker` in the composer utility row, right after the familiar select — wired to `selectedProjectId` / `setSelectedProjectId` with `createProject` + `allowNoProject` (the exact wiring it had before removal).
- It keeps its **own search popover**, so it sits *alongside* the ⚙ Options menu rather than nesting inside it — nested popovers close on outside-click (the same constraint that shaped the Options menu).
- Reuses the existing `.hc-project-selector` CSS (30px, accent-tinted, 178px label cap) — no new styles.
- Flip the home-composer test pin from "removed" → "rendered".

## Verification

- `tsc --noEmit` clean.
- All 5 home-composer test files green (`home-composer`, `-polish`, `-centering`, `-hide-archived`, `-mobile-gaps`).
- Built + ran the app: picker renders in the toolbar, opens its popover (**No project** ✓ / **Add project…**), positioned and styled correctly, no page errors.

### DIFF
--- a/src/components/home-composer.test.ts
+++ b/src/components/home-composer.test.ts
@@ -37,12 +37,13 @@ assert.match(
   "HomeComposer headline should reflect the selected project name",
 );
 
-// Project picker was moved out of the toolbar (no longer rendered inline in
-// the composer card). Project context is preserved in the headline and enhancer.
-assert.doesNotMatch(
+// Project selector lives in the composer toolbar so the user can choose which
+// project a new chat runs in (mirrors the chat composer). It's a standalone
+// ProjectPicker — its own search popover, so it can't nest in the ⚙ menu.
+assert.match(
   source,
-  /className="hc-project-selector"/,
-  "ProjectPicker selector is removed from the home composer toolbar",
+  /<ProjectPicker[\s\S]*value=\{selectedProjectId \|\| null\}[\s\S]*onChange=\{setSelectedProjectId\}[\s\S]*className="hc-project-selector"/,
+  "ProjectPicker is rendered in the home composer toolbar, wired to selectedProjectId",
 );
 
 assert.match(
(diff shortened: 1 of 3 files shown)


BunsDev (commented):
Reviewed against the new chat-parity composer structure (#2484) — **LGTM**.

- Picker slots correctly at the end of `cave-composer-utility-row`; its popover is portaled (`ui/popover.tsx` → `createPortal`) so the panel's `overflow: hidden` can't clip it — right call vs. nesting in the Options menu.
- State wiring is coherent: `value={selectedProjectId || null}` + trigger's `?? projects[0]` fallback matches the send path's `selectedProject … ?? projects[0]`, so the displayed project is always the one used. Explicit No-project (`NO_PROJECT_ID`) survives the reconciliation effect.
- `.home-composer-card .cave-project-picker__trigger.hc-project-selector` styling survived the #2484 CSS cleanup, and both guard tests were flipped coherently.

**Minor (non-blocking):** in `home-composer.css`'s `@container (max-width: 620px)` block, the familiar chip and destination pills get `min-height: var(--touch-target)`, but the project trigger keeps its 30px desktop height on phones (its old mobile rules were removed in #2484 while the picker wasn't rendered). Suggest a follow-up:

```css
@container (max-width: 620px) {
  .home-composer-card .cave-project-picker__trigger.hc-project-selector {
    max-width: none;
    min-height: var(--touch-target);
    padding: 0 30px 0 10px;
  }
}
```

PR #2500, opened 2026-07-06
fix(browser): close native webviews on surface leave and avoid 1x1 offscreen parking

mithril-logic:
## Problem

Two related native-browser defects on macOS (found via an accessibility/computer-use audit of v0.0.140):

1. **Browser pane renders black.** `hide_webview` parked webviews offscreen *and* shrank them to 1×1. Collapsing the layer lets WKWebView drop its backing surface, so a later `browser_set_bounds` re-seat can land as an unpainted (black) layer.
2. **Pages linger after leaving the Browser surface.** Hiding only parks a webview offscreen — its page and process stay alive, and the page content remains in the OS accessibility tree. Reproduction: with VoiceOver/a11y inspection, AXWebArea count goes 1 → 3 after opening the browser, and stays at 3 after leaving the surface.

## Fix

- `hide_webview` keeps the webview at its current size while offscreen, so the layer stays realized and repaints immediately when shown again.
- New `browser_close_all(pane_label)` Tauri command destroys every native webview belonging to a pane (prefix-scoped by label, so the `main` and `companion` panes can't affect each other).
- `BrowserPane` closes its native webviews on unmount instead of leaving them parked; the existing navigate path recreates them on the next mount. The cleanup reads the Tauri bridge through a ref since the bridge arrives async after mount.
- ACL entries added for the new command (`permissions/pty.toml`, `permissions/default.toml`, `capabilities/loopback-browser.json`, `desktop-permissions.test.mjs`) — without them the frontend invoke is denied silently.

## Tradeoff (deliberate, please weigh)

Tab ids are stable across mounts, so the old parking behavior effectively kept pages warm across surface leave/return (scroll position, sessions) via label reuse. This change reloads tabs on return. The same mechanism was the leak: webviews whose tabs went away were orphaned invisibly with live page processes. We chose deterministic teardown over warm pages; if maintainers prefer persistence, the a11y issue would need a different mechanism (e.g. detaching hidden webviews from the a11y tree).

Minor: in dev, React StrictMode's double-mount closes and recreates the webviews once when the surface opens (one reload flicker, dev only).

## Verification

- AXWebArea count now 1 → 3 (browser open) → 1 (surface leave); previously stuck at 3.
- Browser paints opencoven.ai (previously black).
- 518 app test files pass, `tsc` clean, `cargo check` clean; new `browser-polish.test.ts` covers the lifecycle.

### DIFF
--- a/src-tauri/capabilities/loopback-browser.json
+++ b/src-tauri/capabilities/loopback-browser.json
@@ -24,6 +24,7 @@
     "allow-browser-hide",
     "allow-browser-hide-all-except",
     "allow-browser-close",
+    "allow-browser-close-all",
     "allow-browser-reload",
     "allow-browser-report-title",
     "allow-browser-report-scroll"
--- a/src-tauri/permissions/default.toml
+++ b/src-tauri/permissions/default.toml
@@ -12,6 +12,7 @@ permissions = [
   "allow-browser-hide",
   "allow-browser-hide-all-except",
   "allow-browser-close",
+  "allow-browser-close-all",
   "allow-browser-reload",
   "allow-browser-report-title",
   "allow-browser-report-scroll",
--- a/src-tauri/permissions/desktop-permissions.test.mjs
+++ b/src-tauri/permissions/desktop-permissions.test.mjs
@@ -29,6 +29,7 @@ const requiredPermissionIds = [
   "allow-browser-hide",
   "allow-browser-hide-all-except",
   "allow-browser-close",
+  "allow-browser-close-all",
   "allow-browser-reload",
   "allow-shell-open",
 ];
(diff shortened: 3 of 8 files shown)


BunsDev (approved):
Verified the security-sensitive parts: `safe_browser_label` guarantees the close prefix always starts with `cave-browser-`, so `browser_close_all` can never touch the main app webview, and the computed `cave-browser-<pane>-tab-` prefix exactly matches how tab labels are constructed on the frontend; the unmount cleanup is race-safe since the navigate debounce timer is cleared on unmount and `browser_hide` no-ops on closed webviews. Permission/capability additions match the scope of the existing `browser-close` grant, and nothing depends on the removed 1×1 shrink (the bounds loop always re-seats with full geometry) — LGTM, accepting the documented warm-page tradeoff.

BunsDev:
Landed on main as ecfa08dc via #2545 with your authorship credited (Co-authored-by trailer) — thank you! Closing this one only because fork PRs can't satisfy the required CodeQL default-setup check (it never reports on fork head SHAs); the code merged verbatim.

PR #2501, opened 2026-07-06
fix(nav): honest Schedules top-bar button and split-open sidebar state

mithril-logic:
## Problem

1. **The desktop top-bar "Inbox" button opens Schedules.** The `inbox` workspace mode renders `InboxEscalationsView` on its crons tab — nothing inbox-shaped renders behind the button. This looks like drift rather than a decision:
   - The mode-title map already names the mode Schedules (`workspace.tsx`: `inbox: "Schedules"`).
   - #2401 renamed the surface's tab group aria-label to "Schedules tabs" and removed the `InboxFeedList` render (it's now defined-but-unused in `automations-view.tsx`), but didn't revisit the surface's entry-point buttons.
   - Actual inbox items surface through the notification bell (`/api/inbox` SSE) and toasts.
2. **Drag-to-split leaves a stale sidebar active state.** Dropping a page open beside the primary doesn't change `mode`, so the sidebar keeps highlighting the wrong row with no indication of what's open in the split.

## Fix

- Top-bar button relabeled **Schedules** (calendar-check icon, matching aria-label). Its badge now uses the same `needsYou` count as the sidebar's Schedules row, so the button and the surface it opens agree. Note this changes the badge source from unresolved escalations to schedule needs-you items — escalations still badge the sidebar notification count, the mobile bell, and the mobile bottom tabs, so no visibility is lost.
- New pure helper `src/lib/sidebar-nav-state.ts` derives each sidebar row's state (`active` / `split` / `idle`), with hub aliasing (Roles/Capabilities light the Marketplace row) and active-wins precedence. Split-open pages get a `--split` visual modifier instead of a misleading active fill.

## Verification

518 app test files pass, `tsc` clean. New unit tests for the helper plus updated coverage in `familiar-menu-bar`, `sidebar-minimal`, `schedules-tabs`, and `roles-tools-navigation` tests.

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -29,6 +29,7 @@ export const SUITES = {
     "src/lib/use-code-rail.test.ts",
     "src/lib/workspace-tiles.test.ts",
     "src/lib/page-drag.test.ts",
+    "src/lib/sidebar-nav-state.test.ts",
     "src/components/chat-view-render-cap.test.ts",
     "src/lib/perf/web-vitals-format.test.ts",
     "src/lib/app-version.test.ts",
--- a/src/components/familiar-menu-bar.test.ts
+++ b/src/components/familiar-menu-bar.test.ts
@@ -8,7 +8,7 @@ const globals = readFileSync(new URL("../app/globals.css", import.meta.url), "ut
 const menuBarSwitcherRule = globals.match(/\.menu-bar__group--chat \.familiar-switcher__trigger\s*\{([\s\S]*?)\}/)?.[1] ?? "";
 
 // The bar provides desktop top chrome: chat with familiars, global
-// context-aware search, and view tasks/inbox. It is a labelled landmark so
+// context-aware search, and view tasks/schedules. It is a labelled landmark so
 // screen readers can find it.
 assert.match(
   source,
@@ -95,8 +95,8 @@ assert.doesNotMatch(
   "desktop menu-bar familiar selector must not use content-width sizing after the label/caret were removed",
 );
 
-// Right group — tasks. A Tasks button (board) and an Inbox button, each with a
-// live count badge that is hidden at zero.
+// Right group — tasks. A Tasks button (board) and a Schedules button, each
+// with a live count badge that is hidden at zero.
 assert.match(
   source,
   /className="menu-bar__task focus-ring"[\s\S]*onClick=\{onViewTasks\}/,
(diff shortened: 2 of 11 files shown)


BunsDev:
Thanks for this! The branch has merge conflicts with `main` now, which is also why CI never started (required checks can't run). Could you rebase onto current `origin/main` and re-push? Happy to review + merge once checks go green.

mithril-logic:
Rebased onto current `main` (169de4db) and re-pushed (`8538ca1b`). Conflicts were all in the sidebar files against the new §8 quiet-cluster work — resolved by composing the modifiers: `FolderRow` className now carries `--active`/`--split`/`--quiet`/`--quiet-lead` together, and both the quiet-cluster and split-marker CSS blocks + test assertions are kept. Full unit suite passes locally (746 test files). CI should be able to run now.

Heads-up in case it matters for merging: #2500 hit the fork-PR CodeQL limitation you re-landed via #2545 — this one may need the same treatment once checks are green.

BunsDev (approved):
Thanks for the rebase — this now merges cleanly against main tip, and I've verified the core claim holds: the top-bar button still labeled "Inbox" (`ph:tray`, `onViewInbox`) lands on `InboxEscalationsView`, whose tabs are only `calendar | crons` and whose mode title is already "Schedules", so the relabel is correct and the recent Schedules/Automations split didn't address it. I also confirmed `InboxFeedList` in `automations-view.tsx` is defined-but-unused, and that switching the button badge to `scheduleNeedsCount` loses no escalation visibility (unresolved escalations still badge the desktop notification bell and the mobile bell/bottom tabs).

The split-open sidebar state is a nice catch too. The pure `sidebarRowState` helper composes correctly with the workspace's redundant-split-clearing effect, so the defensive "active wins over split" case is unreachable in practice but correctly handled. Unit tests on the helper cover the cases that matter.

Two optional follow-ups, take or leave:
1. Deep-linked mode `"calendar"` renders the same Schedules surface but leaves the Schedules row unlit. That predates your PR, but since you now own `MODE_ALIASES` in `src/lib/sidebar-nav-state.ts`, adding `calendar: "inbox"` would finish the "honest highlight" story in one line.
2. The `--split` state is visual + `title`-text only. If you want it announced programmatically, `aria-description` or appending to the accessible name for split rows would cover screen-reader users; fine as a follow-up.

One process note: fork PRs can't satisfy the required CodeQL default-setup check here, so once CI is green this will be re-landed from a maintainer branch with `Co-authored-by` credit to you (same as #2545 did for #2500) — nothing you need to do there. Approving the substance.

BunsDev:
Landed on main via #2555 with your authorship credited (Co-authored-by trailer in the squash commit) — thank you for the fix and the quick rebase! Closing this one only because fork PRs can't satisfy the required CodeQL default-setup check; your code merged verbatim.

PR #2518, opened 2026-07-06
feat(familiars): expose the autoSelfReport toggle in Brain + ordering cross-links (cave-v1z)

BunsDev:
## Summary

Bead `cave-v1z` — two small familiar-studio gaps:

- **autoSelfReport was wired but unsettable.** The roster surfaces it and chat auto-writes a thread self-report on close when it's on (`chat-view.tsx`), but no UI could set it. The Brain tab now has a **Behavior** card with a `role=switch` row: on stores `true` through the tab's existing `/api/config` PATCH helper; off sends `null` so the key is **deleted** (daemon default is off — no redundant `false` keys accumulate in the config). Failed saves revert the switch, same as the tab's other draft fields.
- **Reciprocal ordering hints.** Appearance → Pin order now says the roster order lives in Familiars → Lifecycle, and the Lifecycle roster hint points back at Appearance → Pin order. Neither control moved (pinned decision).

## Test plan

- Pins appended to `familiar-studio-brain-tab.test.ts`: switch + `aria-checked`, null-delete save shape, failure revert, both hint copies — green; all pre-existing Brain-tab pins untouched and green
- Full app suite 531 ✓ · `tsc --noEmit` ✓ · minimalism accent-white invariant unaffected (knob pattern matches the projects-tab switch)
- **Verified live** (worktree server + Playwright, PATCH bodies captured): ON → `{"familiars":{"nova":{"autoSelfReport":true}}}` persisted to `cave-config.json`; state survives reload; OFF → `{"autoSelfReport":null}` → 200 → key deleted; config left clean

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -6811,6 +6811,12 @@ a.mobile-handoff__link:hover {
   font-size: var(--text-sm);
   font-weight: 600;
 }
+.familiar-studio-brain__hint {
+  margin: var(--space-2) 0 0;
+  font-size: 12px;
+  color: var(--text-muted);
+  line-height: 1.4;
+}
 .familiar-studio-brain__field-grid {
   display: grid;
   grid-template-columns: repeat(2, minmax(0, 1fr));
--- a/src/components/familiar-studio-brain-tab.test.ts
+++ b/src/components/familiar-studio-brain-tab.test.ts
@@ -93,4 +93,39 @@ assert.match(
   "Brain workspace should collapse to one column on narrow surfaces",
 );
 
+// ── autoSelfReport toggle + ordering cross-links (cave-v1z) ──────────────────
+
+assert.match(
+  source,
+  /role="switch"[\s\S]{0,120}aria-checked=\{draftAutoSelfReport\}/,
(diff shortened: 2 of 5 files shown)


BunsDev:
Superseded by #2515 — the session that claimed cave-v1z (visible as the in_progress flip before this PR started) shipped the same toggle + reciprocal hints and merged first. Their Appearance-side hint is a navigating button (richer than this PR's plain copy); no deltas worth salvaging.

BunsDev:
Heads-up from the session that shipped #2515: confirming the supersession is complete on main (`43e26bad`, ancestor-verified) — the Reflection card ships the same `role="switch"` toggle with the null-delete save shape, plus both ordering hints (the Appearance-side link lands directly on the Lifecycle tab via the studio context).

Leftovers on this side when convenient: the local worktree `.worktrees/feat-brain-auto-self-report` (clean) and the local + remote `feat/brain-auto-self-report` branch still exist. Nothing unmerged of value showed in `git log main..feat/brain-auto-self-report`, so standard cleanup is safe — left untouched since the worktree is yours. Also filed as bead memory `auto-self-report-superseded-cleanup`; delete it once cleaned.

PR #2534, opened 2026-07-06
fix(shell): mode-fade releases its transform — fixed overlays get the real viewport back (cave-cco)

BunsDev:
Fixes the root cause behind four shipped portal workarounds (#537 board drawer, #1984 lightbox/expand overlays, the github hover card) and the #2526 fullscreen code-rail clip. Bead: cave-cco.

## Root cause

`.cave-mode-fade { animation: cave-mode-in 120ms ease-out both }` kept the entrance animation's transform applied **forever**, making every mode wrapper the containing block for `position: fixed` descendants. Any fixed overlay rendered inside a surface resolved against the mode area — origin after the shell's icon rail, below the top bar — instead of the viewport.

Two findings worth recording (both pinned in `rail-files-panel.test.ts`):

1. **Dropping `transform` from the `to` keyframe is not enough.** A forwards fill keeps every *animated* property actively applied at its final interpolated value — transform is animated in the from-frame, so Chromium holds an **identity** transform (`matrix(1,0,0,1,0,0)`) on the wrapper for the life of the surface. An applied identity transform still creates the containing block. Verified empirically on a prod build.
2. **`fill-mode: backwards` is the fix.** The animation releases on finish; its end state equals the base state (opacity 1, no transform), so nothing visually jumps. The containing block now exists only during the 120ms entrance.

## Blast-radius audit (before changing anything)

Classified every `position: fixed` element that can render inside a surface (two parallel sweeps + manual verification of the workspace.tsx mount points):

- **Portaled to body** (Modal, Popover, mobile drawer, board drawer, lightboxes, artifact viewer, voice overlay, inspector fullscreen, mermaid viewer, kanban touch ghost): unaffected; portals stay (they also escape stacking contexts and the transient 120ms window).
- **Mounted outside the mode wrapper** in workspace.tsx (command palette, onboarding overlay, inbox toasts, glyph picker, reminder modal, deep-link takeover): unaffected.
- **JS-coordinate elements inline in surfaces** (gh profile card, artifact comment FAB, quick-chat anchor math): these assume viewport coords and were silently offset — **now repaired**.
- **`inset-0` modals/backdrops inline in surfaces** (capabilities preview, familiars memory overlays, github PAT modal, quick-open, cwd picker, skill drawer, chat sheets, flow template overlay, dash snooze backdrop, workflow dialog): grow from mode-area to full viewport — correct for scrims, which should cover the nav too.
- **Right/bottom-anchored elements**: don't move (the mode area shares those edges with the viewport). **No inline top-anchored fixed elements exist** — the only category that could regress has zero members.
- The fullscreen code rail becomes **true viewport fullscreen** (covers shell + top bar) — what the toggle promises; its inset-only sizing from #2526 was designed to be correct under both regimes.

## Verification

- Live prod-build probe: computed transform on `.cave-mode-fade` is `none` after entrance across chat → board → dashboard switches (was identity matrix); fullscreen rail rect is exactly `0,0 → 1600,900`; ancestor walk finds **no remaining containing block**.
- Entrance animation unchanged visually (4px slide + fade still runs; missing keyframe properties interpolate to the underlying value).
- `tsc` clean · 541 app test files green · bundle in budget.
- Comments updated where the old behavior was documented (board-inspector portal rationale, cave-chat.css fullscreen rule).

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -3397,12 +3397,26 @@ textarea.required-inputs-control {
 }
 
 /* ── UX-004: Mode transition crossfade ──────────────────────────────────── */
+/* fill-mode is `backwards`, NOT `both`/`forwards`, and the `to` frame has no
+   transform. A forwards fill keeps every animated property actively applied at
+   its final value forever — and because transform is animated (from-frame),
+   Chromium holds an IDENTITY transform on the wrapper for the life of the
+   surface (verified: omitting transform from `to` alone still computes
+   matrix(1,0,0,1,0,0) under `both`). Any applied transform, even identity,
+   makes the mode wrapper the containing block for position:fixed descendants:
+   fixed overlays inside surfaces resolved against the mode area (origin after
+   the shell rail / below the top bar) instead of the viewport — the bug behind
+   the #537 / #1984 / github hover-card portal workarounds and the #2526
+   fullscreen code-rail clip. With `backwards` the animation releases on
+   finish (end state equals base state, so nothing jumps) and the containing
+   block exists only during the 120ms entrance. Pinned in
+   rail-files-panel.test.ts. */
 @keyframes cave-mode-in {
   from { opacity: 0; transform: translateY(4px); }
-  to   { opacity: 1; transform: translateY(0); }
+  to   { opacity: 1; }
 }
 .cave-mode-fade {
-  animation: cave-mode-in 120ms ease-out both;
+  animation: cave-mode-in 120ms ease-out backwards;
 }
 
 /* Left nav collapsed strip -- mirrors the agent tab on the right */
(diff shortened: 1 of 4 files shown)


BunsDev:
Superseded by #2532 — the sibling session landed the identical fix (fill-mode both→backwards + blast-radius audit + fill-mode pin in globals.css.test.ts) while this PR's CI was running. One nuance from this branch worth the record: dropping transform from the `to` keyframe alone is NOT sufficient — under a forwards fill Chromium keeps the animated transform applied at its final interpolated value, so the wrapper computes matrix(1,0,0,1,0,0) forever; only releasing the fill restores transform:none. #2532's backwards fill is exactly right.

PR #2552, opened 2026-07-07
fix(api): retry transient daemon flakes on the familiars GET (cave-4po)

BunsDev:
Fixes bead **cave-4po**: `GET /api/familiars` intermittently returns 503 after ~5.6s.

## Root cause
The route makes a single `callDaemon` (4s timeout) with no retry. A transiently stalled daemon status probe — or a daemon mid-restart — times out that one call, and the route 503s the entire roster. Every cockpit/dashboard panel keyed on familiars degrades with it. The observed ~5.6s = 4s daemon timeout + route/app overhead.

**Repro** (fake unix-socket daemon stalling every 3rd request past the client timeout): 3/9 calls failed with `daemon timeout` at ~4.1s. With the fix: **9/9 succeed** (worst case ~4.2s success).

## Fix
`callDaemon`/`callDaemonTarget` gain an **opt-in** `retryTransient` flag:
- retries **exactly once**, **GET-only** (non-idempotent methods never fire twice)
- only for transient failures (`status 0` + `daemon timeout`/`daemon offline` via new `isTransientDaemonFailure`); real daemon HTTP errors (4xx/5xx) pass through unretried, so the honest daemon-offline states from #2511 stay honest
- 200ms backoff; second attempt capped at 2s

Only the familiars GET opts in; all other call sites are behavior-unchanged.

## Verification
- New unit tests: transient classification; live socket-server retry (stall → recover, exactly 2 hits); no retry without opt-in; POST never retries; HTTP 500 not masked.
- Route source pin: familiars GET carries `retryTransient: true`.
- `pnpm typecheck` clean; `test:app` 549 files, `test:api` 124 files green.

### DIFF
--- a/src/app/api/familiars/route.test.ts
+++ b/src/app/api/familiars/route.test.ts
@@ -4,6 +4,14 @@ import { readFileSync } from "node:fs";
 
 const source = readFileSync(new URL("./route.ts", import.meta.url), "utf8");
 
+// One transient daemon flake (stalled probe / mid-restart) must not 503 the
+// roster — the GET opts into the single transient retry (cave-4po).
+assert.match(
+  source,
+  /retryTransient: true/,
+  "Familiars GET should retry once on a transient daemon failure",
+);
+
 assert.match(
   source,
   /const configEntry = config\.familiars\[f\.id\] \?\? \{\}/,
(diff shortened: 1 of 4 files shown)


BunsDev:
Closing unmerged: #2551 (merged while this was in checks) fixes the same bead (cave-4po) with a superset approach — an always-on single retry for ALL transport-level GET failures plus settle/deadline hardening on every daemon call, so the familiars GET is covered without a per-route opt-in. Concurrent-session duplicate; no salvage needed. Verified main's implementation at src/lib/coven-daemon.ts (GET_RETRY_DELAY_MS block).

PR #2590, opened 2026-07-07
chore(avatar): retire legacy browser-local avatar store (cave-154)

BunsDev:
## Changes

-


BunsDev (commented):
Reviewed — clean, well-tested retirement. Approve-in-spirit with one rollout call to make consciously.

**✅ Strengths**
- **Complete removal:** no dangling references to `user-avatar-image` / `user-avatar-migrate` anywhere on the branch; the read path (`readUserAvatarImageSnapshot` / `whenUserAvatarHydrated` / `runUserAvatarMigration`) has **zero** remaining callers. Tests + `run-tests.mjs` wiring are updated in lockstep.
- **`legacy-svg-avatar-hint.ts` has excellent IndexedDB hygiene:** SSR guard, a `settled` latch against double-resolve, error/blocked/success handlers with `db.close()` on every path, and it opens **version-less and aborts `onupgradeneeded`** so the residue check never *creates* the DB. Correctly returns `false` when the `userAvatar` store is already gone.
- **Strong test:** only an SVG `userAvatar:user` record counts (PNG → false), missing/unreadable → false, `indexedDB` undefined → false, and it pins `version === undefined`.

**⚠️ The one thing to decide — the soak period**
cave-154 says retire this *"after a migration soak."* The migration (#2541) merged **2026-07-06 ~20:10 — under a day ago**, and this removes `runUserAvatarMigration` now. After it lands, the app no longer reads the browser-local avatar at all, so any user who had a local avatar but hasn't opened the app since yesterday evening won't be migrated — an SVG gets the re-upload hint, but a **non-SVG local avatar silently disappears** (server has nothing). For a solo/operator instance this is almost certainly fine; just worth an intentional "the soak is enough" call, ideally recorded in the (currently empty) **Risk + rollout notes**.

**🔻 Minor nits**
- **PR body is empty** (all template placeholders) — a Summary/Verification/Risk note would help the next reviewer, especially the soak rationale above.
- **`"use client"` in `legacy-svg-avatar-hint.ts`** is unnecessary — it's a pure async utility, not a component; the directive only matters at a component boundary. Harmless, just noise.

Overall: high quality and a thorough removal. The only real gate is the soak decision.

PR #2612, opened 2026-07-07
fix(windows): resolve coven through PATH before the literal-name fallback

### LINKED ISSUE #2610: [Windows] Chat send always fails with "coven CLI not found on PATH" while Setup shows everything green
## Environment

- Windows 11 Pro (10.0.26200)
- CovenCave 0.0.149, unsigned MSI from GitHub Releases
- Node LTS + npm 11.16.0 via scoop
- coven CLI via `npm i -g @opencoven/cli` (0.0.53, registry latest), coven-code 0.0.22

## Symptom

Every chat send fails with:

> coven CLI not found on PATH. Open Setup to install it, then try again.

(`src/lib/coven-spawn-error.ts:14`, surfaced by `src/app/api/chat/send/route.ts:1609`)

…and clicking through to Setup shows **all steps green**, including the coven CLI check. The user bounces between a broken chat and a wizard that insists nothing is wrong. Deterministic; survives app restarts.

## Repro

1. Windows box where Node/npm come from scoop (or anything that isn't nvm/fnm/`%APPDATA%\npm`).
2. `npm i -g @opencoven/cli` — `where coven` resolves fine in any terminal.
3. Install CovenCave from the MSI, complete onboarding (all green).
4. Create a familiar, send any message.
5. Error above; Setup still green.

## Root cause (verified live, including a repro with the sidecar's exact env extracted from its PEB)

Two different resolution strategies disagree:

- **Status checks** resolve coven via `where coven` with the augmented spawn env → succeeds (45 ms, every run). Wizard: green.
- **Chat send** resolves via `covenBin()` (`src/lib/coven-bin.ts`), which probes a hardcoded candidate-dir list: nvm, fnm, `%APPDATA%\npm`, `npm_config_prefix`, pnpm, bun, homebrew, `/usr/local/bin`, `~/.local/bin`, `~/.cargo/bin`. On a scoop box none of those contain coven, so it falls back to the **literal string `"coven"`** and spawns it without a shell. A bare Windows `spawn()` only resolves `.exe`/`.com` — npm installs `coven.cmd` — so the spawn ENOENTs even though the CLI is plainly on PATH.

Two adjacent hits from the same trap:

- `where coven` lists npm's **extensionless POSIX launcher first**, and existing call sites take the first line — so even where `where` is used, the returned path isn't spawnable, and `installedTool()`'s `--version` probe can fail (Node also refuses to `execFile` a `.cmd` directly since the batch-file hardening), reporting `version: null` → cosmetic "Needs update".
- The install route already documents this exact npm `.cmd` trap for its own spawns; the coven spawn path just missed it.

## Workaround

`setx npm_config_prefix "C:\Users\admin\scoop\persist\nodejs-lts\bin"` (the value npm already uses, so npm behavior is unchanged), then fully quit and relaunch Cave — `candidateDirs()` honors `npm_config_prefix`, finds `coven.cmd`, and the existing cmd-shim parser spawns `node <coven.js>` directly. Verified: sessions spawn after this.

## Suggested fix

On win32, `covenBin()` should fall back to `where coven` before the literal name, and every `where` call site should prefer a spawnable launcher (`.exe`/`.com`, then `.cmd`/`.bat` — converted via the existing `covenLaunchCommandForBinary()`) instead of taking the first line. No `shell: true` needed anywhere.

I have this working in branch `fix/windows-coven-launcher` (with tests, extending `src/lib/coven-bin.test.ts`) — happy to PR it.

mithril-logic:
Fixes #2610.

`covenBin()`'s last resort is the literal string `"coven"`, and a bare Windows `spawn()` only resolves `.exe`/`.com` — npm installs `coven.cmd` — so on any box where coven isn't in the well-known candidate dirs (scoop/winget Node), every chat send ENOENTs while Setup's `where`-based check stays green.

- `covenBin()`: on win32, fall back to `where coven` + spawnable-launcher picking before the literal name
- new `pickWindowsLauncher()` shared by every `where` call site: prefer `.exe`/`.com`, then `.cmd`/`.bat` (npm lists an unspawnable extensionless POSIX launcher first), then the first entry
- tools-status + onboarding status/install routes: pick the spawnable launcher from `where` output, and probe versions through the existing `covenLaunchCommandForBinary()` cmd-shim conversion so `.cmd` doesn't EINVAL under Node's batch-file hardening

Tests: extended `src/lib/coven-bin.test.ts` (behavioral cases for the picker + source-pattern asserts); suite passes. Verified end-to-end on Windows 11 — sessions spawn with this change on the box from #2610.

Note: as a fork PR this won't get a CodeQL result on the head SHA (same limitation as #2500 → re-landed in #2545) — not a failing check on our side.

### DIFF
--- a/src/app/api/onboarding/install/route.ts
+++ b/src/app/api/onboarding/install/route.ts
@@ -5,7 +5,12 @@ import { constants as fsConstants } from "node:fs";
 import { dirname, join } from "node:path";
 import { promisify } from "node:util";
 import { stripAnsi } from "@/lib/ansi";
-import { covenBin, covenSpawnEnv, refreshCovenSpawnEnv } from "@/lib/coven-bin";
+import {
+  covenBin,
+  covenSpawnEnv,
+  pickWindowsLauncher,
+  refreshCovenSpawnEnv,
+} from "@/lib/coven-bin";
 import { callDaemon } from "@/lib/coven-daemon";
 
 export const dynamic = "force-dynamic";
@@ -98,7 +103,13 @@ async function commandPath(
         env,
         timeout: 1500,
       });
-      return { path: stdout.trim().split(/\r?\n/)[0] || null };
+      const lines = stdout.split(/\r?\n/);
+      return {
+        path:
+          process.platform === "win32"
+            ? pickWindowsLauncher(lines)
+            : lines.map((l) => l.trim()).find(Boolean) ?? null,
+      };
     } catch (err) {
       const code = (err as { code?: unknown }).code;
       if (code === 1) return { path: null };
--- a/src/app/api/onboarding/status/route.ts
+++ b/src/app/api/onboarding/status/route.ts
@@ -6,7 +6,7 @@ import path from "node:path";
 import { promisify } from "node:util";
 import { callDaemon } from "@/lib/coven-daemon";
 import { loadConfig } from "@/lib/cave-config";
-import { covenLaunchCommand, covenSpawnEnv } from "@/lib/coven-bin";
+import { covenLaunchCommand, covenSpawnEnv, pickWindowsLauncher } from "@/lib/coven-bin";
 import {
   openCovenToolStatuses,
   type OpenCovenToolStatus,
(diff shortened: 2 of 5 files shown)


BunsDev:
Merged via #2625 — re-landed from an internal branch because fork PRs cannot satisfy the required CodeQL default-setup check. Your commit is credited with a Co-authored-by trailer on the squash commit. Thank you!

PR #2613, opened 2026-07-07
fix(windows): rebuild spawn PATH from the registry instead of a login shell

### LINKED ISSUE #2611: [Windows] PATH self-heal is macOS-shaped: "npm not found" after installing Node until a full app restart
## Environment

- Windows 11 Pro (10.0.26200), CovenCave 0.0.149 (MSI), Node LTS + npm 11.16.0 via scoop

## Symptom

Onboarding's install step reported **"npm not found — Node.js setup needed first"** even though npm was installed and on the user PATH. Nothing in the app recovers; only fully quitting Cave and relaunching from the Start menu fixes it.

(Trigger on this box: Cave was launched from the MSI finish screen, so the sidecar inherited the installer's environment — no per-user PATH, no scoop dirs. The same applies to anyone who installs Node *while* Cave is running.)

## Root cause

The recovery path can never learn anything new on Windows:

- `refreshCovenSpawnEnv()` → `loginShellPath()` (`src/lib/coven-bin.ts`) execs `$SHELL -ilc 'echo $PATH'`. `SHELL` is unset on Windows, so it tries `/bin/zsh` and silently returns null.
- `candidateDirs()` knows nvm/fnm/Homebrew/`%APPDATA%\npm` but not scoop or winget's Node layouts.
- `process.env.PATH` is frozen at process start.

So the "refresh once with a rebuilt PATH before concluding the binary is missing" logic (which works nicely on macOS for the Finder/Spotlight minimal-PATH case) is a no-op on Windows.

## Suggested fix

On win32, rebuild PATH from the registry instead of a login shell: `HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment` + `HKCU\Environment` (with `REG_EXPAND_SZ` expansion), machine + user concatenated — the same view a new terminal gets. Then npm-installed-after-Node self-heals with no app restart, matching the macOS behavior.

I have this working in branch `fix/windows-path-refresh` — happy to PR it.

mithril-logic:
Fixes #2611.

On Windows, `refreshCovenSpawnEnv()` can never learn anything new: `loginShellPath()` execs `$SHELL -ilc 'echo $PATH'`, `SHELL` is unset on Windows, so the probe always fails and a "refreshed" PATH equals the stale launch-time PATH. That's why installing Node/coven from onboarding still reports "npm not found" until a full app restart.

This reads PATH the way Windows itself builds it: `HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Environment` then `HKCU\Environment` via `reg query`, with `%VAR%` expansion for `REG_EXPAND_SZ` (case-insensitive, unknown vars left intact — matching Windows semantics). Installers write new tool dirs to the registry and broadcast `WM_SETTINGCHANGE`, which running processes never see — re-reading the registry is what makes a refresh actually pick them up. Non-Windows keeps the login-shell path.

Tests: exported the `reg query` parser and added behavioral cases (expansion, case-insensitivity, REG_SZ verbatim, missing value) to `src/lib/coven-bin.test.ts`; suite passes.

Note: as a fork PR this won't get a CodeQL result on the head SHA (same limitation as #2500 → re-landed in #2545) — not a failing check on our side.

### DIFF
--- a/src/lib/coven-bin.test.ts
+++ b/src/lib/coven-bin.test.ts
@@ -7,7 +7,7 @@ import assert from "node:assert/strict";
 import { mkdir, mkdtemp, readFile, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
-import { covenLaunchCommandForBinary } from "./coven-bin.ts";
+import { covenLaunchCommandForBinary, windowsPathFromRegQuery } from "./coven-bin.ts";
 
 const source = await readFile(new URL("./coven-bin.ts", import.meta.url), "utf8");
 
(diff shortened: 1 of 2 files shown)


BunsDev:
Merged via #2621 — re-landed from an internal branch because fork PRs cannot satisfy the required CodeQL default-setup check. Your commit is credited with a Co-authored-by trailer on the squash commit. Thank you!

Issue #2614, opened 2026-07-07
Native folder picker is dead code in the packaged app (no Tauri IPC on the localhost origin), and both fallbacks misbehave


mithril-logic:
Filing issue-first because the core fix is architectural and I'd like your direction before PRing.

## Environment

- Windows 11 Pro (10.0.26200), CovenCave 0.0.149 (MSI)
- Symptom first seen: home composer → "Add project"

## Symptom

Clicking **Add project** from the home composer shows a folder browser rendered *inside* the text-input box, top-clipped (~4 rows visible), with a stray empty box beside it. No native Windows folder dialog ever appears. (Screenshot available.)

### (a) `isTauri()` is false inside the actual desktop app — the native path never runs

The main window loads the UI from the bundled Node sidecar via `WebviewUrl::External(main_url)` (`src-tauri/src/lib.rs:1176`), i.e. an `http://localhost:<port>` origin. The repo defines **no Tauri capabilities**, so the localhost origin never gets the IPC bridge: `window.__TAURI_INTERNALS__` is undefined, `isTauri()` (`src/lib/tauri-platform.ts:22`) returns false, and every native-first JS path silently takes its web fallback — folder picker included, and likely on every platform, not just Windows.

Verified live: clicking Add project spawns no `powershell.exe` at all; Cave never attempts the native dialog.

### (b) Even when it does run, the Windows native dialog opens buried

Repro'd the exact `powershell.exe -Sta` WinForms `FolderBrowserDialog` invocation from `shell_pick_directory` (`src-tauri/src/lib.rs:711`, script at `:734`) manually: the dialog has no owner window, so it opens **behind every window**, unfocused, with no taskbar presence — it looks like the click did nothing. (I only found it by minimizing everything.)

### (c) The web-fallback `DirectoryPickerModal` renders clipped

It's meant to overlay near the caller's root; mounted from the home composer it ends up inside the input container and is clipped by it (the state in the screenshot).

Also: all three platform dialogs still prompt **"Choose a folder for Graphify"** (`lib.rs:717/:734/:749`) — looks like leftover copy from another app; I'll send a tiny separate PR for that (PR incoming).

## Suggested directions — which would you prefer?

For (a), either:
1. add a `remote` capability granting the localhost origin the needed IPC permissions, or
2. serve the frontend as `WebviewUrl::App` and talk to the sidecar over HTTP only.

[...]

PR #2616, opened 2026-07-07
fix(desktop): folder picker still says "Choose a folder for Graphify"

mithril-logic:
Part of #2614 (smallest piece, separable): all three native `shell_pick_directory` prompts still say **"Choose a folder for Graphify"** — pre-rename copy, so the OS dialog looks like it belongs to another program. Renames the prompt to CovenCave on macOS (osascript), Windows (FolderBrowserDialog), and Linux (zenity). String-only change; the remaining "Graphify" mentions in the repo are historical CHANGELOG entries, deliberately untouched.

Note: as a fork PR this won't get a CodeQL result on the head SHA (same limitation as #2500 → re-landed in #2545).

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -714,7 +714,7 @@ fn shell_pick_directory() -> Result<Option<String>, String> {
         let output = std::process::Command::new("osascript")
             .args([
                 "-e",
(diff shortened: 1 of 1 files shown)


BunsDev:
Merged via #2623 — re-landed from an internal branch because fork PRs cannot satisfy the required CodeQL default-setup check. Your commit is credited with a Co-authored-by trailer on the squash commit. Thank you!

PR #2617, opened 2026-07-07
fix(ui): platform-aware search shortcut badge instead of hard-coded ⌘K

mithril-logic:
The search fields in the top bar and familiar menu bar always render a hard-coded `⌘K` badge, which is wrong on Windows/Linux where the palette binds Ctrl+K. Both now read the modifier from the existing `useKeySymbols()` hook — the same idiom the command palette already uses (`{keys.mod}K`).

Kept deliberately minimal: these files carry a few more hard-coded Mac glyphs in hover hints (`title="Quick chat (⌘J)"`, aria-labels ⌘B/⌘\) — happy to sweep those in a follow-up if you want them platform-aware too.

Tests: `top-bar.test.ts` and `familiar-menu-bar.test.ts` pass.

Note: as a fork PR this won't get a CodeQL result on the head SHA (same limitation as #2500 → re-landed in #2545).

### DIFF
--- a/src/components/familiar-menu-bar.tsx
+++ b/src/components/familiar-menu-bar.tsx
@@ -2,6 +2,7 @@
 
 import { Icon } from "@/lib/icon";
 import { FamiliarQuickSwitch } from "@/components/familiar-quick-switch";
+import { useKeySymbols } from "@/lib/platform-keys";
 import type { ResolvedFamiliar } from "@/lib/familiar-resolve";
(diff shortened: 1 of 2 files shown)


BunsDev:
Merged via #2624 — re-landed from an internal branch because fork PRs cannot satisfy the required CodeQL default-setup check. Your commit is credited with a Co-authored-by trailer on the squash commit. Thank you!

PR #2691, opened 2026-07-08
fix(comux): make terminal pane numbers direct jumps

CompleteDotTech:
## Summary

- Turns split terminal pane number badges into direct focus controls.
- Adds hover/focus styling so the controls remain compact but keyboard-visible.
- Extends the Comux wiring test to lock the direct pane-jump behavior.

## Testing

- `node --experimental-strip-types src/components/comux-broadcast-wiring.test.ts`

## Notes

- `pnpm typecheck` currently fails before this change on `.next/types/validator.ts` because it references missing generated `../../src/app/api/youtube/route.js`. This PR is draft until we test the interaction in-app.

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -3041,6 +3041,7 @@ textarea.required-inputs-control {
 /* Pane number badge — backs ⌘1…9 quick-jump and reads like a tmux pane index.
    Tints to the brand accent on the focused pane. */
 .comux-terminal-pane-num {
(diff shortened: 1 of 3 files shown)


CompleteDotTech:
Closing this draft because it addressed the wrong interpretation of the request. The correct webhook file-linking PR is https://github.com/OpenCoven/coven-github-webhook/pull/9.

PR #2709, opened 2026-07-08
fix(palette): Enter opens the best local match — Ask-Salem row trails the results (cave-42r5)

BunsDev:
## Problem (cave-42r5)
The Ask-Salem row was **prepended** for every non-slash query, and every keystroke resets `activeIdx` to 0 — so the launcher's core gesture (type a few chars → Enter → open best hit) always fired an AI network call via `/api/salem` instead. Local matches (sessions, familiars, cards, surfaces) were never Enter-reachable without arrowing past the Salem row first.

## Fix
`return [...localRows, ...salemRows]` — local matches lead, Ask-Salem is pinned last (still arrow-reachable; loading/answer regions unchanged). Ordering direction confirmed by @BunsDev in-session (the AI-first ordering originated in #1031).

## Verification
- 2 new source pins (salem-last + regression guard on the old ordering)
- `pnpm test:app` — 570 files pass · `pnpm typecheck` clean

Closes cave-42r5 (beads).

### DIFF
--- a/src/components/command-palette-a11y.test.ts
+++ b/src/components/command-palette-a11y.test.ts
@@ -46,4 +46,19 @@ assert.match(
   "palette keyboard handler ignores keydowns while an IME composition is in progress",
 );
 
+// cave-42r5: every keystroke resets activeIdx to 0, so whatever row is first
+// owns the type-then-Enter gesture. Local matches must lead and Ask-Salem must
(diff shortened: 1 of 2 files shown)


BunsDev:
Superseded by #2706, which landed the identical cave-42r5 fix (salem row last + matching regression pins) while this PR's checks were running. Nothing left to land.

PR #2713, opened 2026-07-08
fix(ui): Popover earns its dialog role; theme-editor reset clears persistence (cave-fu1y)

BunsDev:
## Summary

Bead `cave-fu1y` (source-verified audit findings), three fixes:

**Popover earns its `role=dialog`.** The shared popover announced a dialog to AT while Tab freely walked the page behind it — no containment, no focus entry, no `aria-modal`. It now: contains Tab (edge wrap + recapture when focus escaped) inside its existing capture-phase key handler; moves focus into the panel on open *unless a consumer already placed it* (child effects run first, so consumer traps/autofocus win); and declares `aria-modal="true"` + `tabIndex={-1}`. **Deliberate design choice:** containment is implemented natively rather than via `useFocusTrap`, because the shared trap's cleanup unconditionally returns focus to the trigger — which would steal focus from whatever control the user clicked to dismiss the popover, regressing the conditional return-focus behavior all 12+ call sites rely on. Light-dismiss (outside click) is unchanged.

**Theme-editor reset now clears persistence.** The editor's reset path reset React state and DOM vars only; the persisted custom theme in localStorage resurrected on the next reload — divergent from the Appearance-tab reset, which clears storage. The `onReset` handler now removes `COVEN_CUSTOM_THEME_KEY` before restoring the base preset (which the persist effect then writes).

**Saved-flash timer leak.** `setTimeout(() => setSaved(false), 2000)` was never cleared — setState after unmount, and rapid re-saves stacked timers so an early flash could truncate a later one. Now a tracked ref, cleared on re-save and unmount.

## Test plan

- [x] Pins in `ui/popover.test.ts` (aria-modal, Tab containment, recapture, consumer-first focus) and `settings-appearance.test.ts` (reset clears the key; tracked timer + unmount cleanup)

### DIFF
--- a/src/components/settings-appearance.test.ts
+++ b/src/components/settings-appearance.test.ts
@@ -414,3 +414,15 @@ assert.match(
   /function applyTokenOverride\(key: string, hex: string, mode: Mode\)/,
   "editing a token forks the active theme to a custom theme and re-syncs",
 );
+
+// ── Editor reset clears persistence (cave-fu1y) ──────────────────────────────
+// Resetting from the theme editor only reset React state; the persisted
+// custom theme resurrected on reload. The onReset path now drops the key,
+// matching the Appearance-tab reset. The editor's saved-flash timer also
+// clears on unmount/re-save instead of firing setState on a dead component.
+assert.match(settings, /localStorage\.removeItem\(COVEN_CUSTOM_THEME_KEY\); \} catch [\s\S]{0,80}setActiveTheme\(colorEditorBase\)/, "editor reset drops the persisted custom theme");
+{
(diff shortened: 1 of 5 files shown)


BunsDev:
Closing as superseded by #2712 (merged df94ddd9), which covers all three findings: the popover fix chose the non-modal fork (close-on-focus-out, no aria-modal) — a coherent, arguably more standard pattern for light-dismiss popovers that this PR's containment fork would directly contradict; the reset-persistence clear landed inside theme-color-editor (better placement — covers every consumer, not just settings-shell); and the saved-timer leak is fixed there too. Nothing here survives as a delta. (Session ca37f515 held the bead claim + ACTIVE note when #2712's branch was cut — third claim-race today; flagging for the coordination doc.)

PR #2716, opened 2026-07-08
perf(automations): stable in-flight run poll — no interval churn, no N+1 fan-out (cave-1e6k)

BunsDev:
## Summary

Bead `cave-1e6k` (Schedules audit follow-up, source-verified). While an automation run was in flight, the 2.5s status poll wasted work two ways:

**Interval churn.** The poll effect depended on the whole `automationRuns` array, and `refreshRuns` stored a fresh array reference on every tick — so the effect tore down and recreated its `setInterval` every 2.5 seconds. The runs write is now content-guarded (`arrayContentEqual` keeps the previous identity when nothing changed, the same guard `load()` uses for items/autos), and the effect depends on a derived `hasRunningRun` boolean, so the interval only re-arms when running-ness actually flips.

**N+1 fan-out on the same endpoint.** Each tick also called `refreshLastRuns()`, firing one `/api/codex-automations/{id}/runs` request *per automation* — and the selected automation's newest run was already present in the response `refreshRuns` had just fetched from the identical endpoint. `refreshRuns` now maintains the last-run badge map from its own response (identity-guarded); the full fan-out survives only where it belongs, on roster changes.

Net effect: one request per poll tick and one long-lived interval, regardless of roster size.

## Test plan

- [x] Pins: boolean poll gate, no runs-array dep, content-guarded write, no per-automation fan-out in the hot loop
- [x] Automations tests + full `pnpm test:app` (571 files) green; `pnpm typecheck` clean

### DIFF
--- a/src/components/automations-view.test.ts
+++ b/src/components/automations-view.test.ts
@@ -186,3 +186,14 @@ assert.match(source, /togglePauseAutomation: toggleCodex/, "cron pause/resume re
 assert.match(source, /const req = \+\+runLogReqRef\.current;/, "each log fetch takes a request token");
 assert.match(source, /if \(req !== runLogReqRef\.current\) return;/, "stale log responses are dropped");
 assert.match(source, /runLogReqRef\.current \+= 1;\s*\n\s*setOpenRunId\(null\);/, "closing the log invalidates the in-flight fetch");
+
+// ── In-flight run-poll stability (cave-1e6k) ─────────────────────────────────
+// The poll effect depends on a derived boolean, not the runs array — an
+// unchanged 2.5s tick must not tear down and recreate the interval. The runs
+// write is content-guarded for the same reason, and the poll no longer fans
+// out one request per automation: refreshRuns' own response maintains the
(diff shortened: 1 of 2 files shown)


BunsDev:
Reviewed during the branch-backlog sweep — **approve**.

- Content-guarded `setAutomationRuns` matches the app-wide poll pattern (same `arrayContentEqual` guard `load()` uses), and deriving `hasRunningRun` for the effect dep genuinely stops the 2.5s interval churn — the boolean only flips on running-ness changes.
- The `runs[0]` newest-run extraction in the hot path is exactly consistent with the pre-existing `refreshLastRuns` (`j?.runs?.[0]`) against the same endpoint — no new ordering assumption.
- Trade-off is sound and documented: other automations' badges go stale only while a run polls, and the full fan-out still runs on roster changes.
- Minor (non-blocking): the in-poll badge update upserts only the selected id while `refreshLastRuns` rebuilds the whole map — consistent with when each fires; worth keeping in mind if the map ever gains consumers that expect pruning.

PR #2717, opened 2026-07-08
a11y(schedules): make the two detail panels real dialogs (cave-qfdv)

BunsDev:
## What

From the Schedules-surface audit. The reminder DetailPanel and cron CodexDetailPanel rendered as bare divs — no role, no focus trap, no Escape-to-close, and a plain Close dropped focus to body.

## Fix
Adopt the shared useFocusTrap the sibling calendar surface uses: both panels trap focus, close on Escape, and restore focus to the opening row, with role=dialog + aria-labelledby. aria-modal is deliberately omitted (desktop split-pane keeps the list interactive; mobile hides it via display:none).

Pinned in automations-view.test.ts.

## Verification
typecheck, 571 app tests, build within budget. Audit LOW/perf findings filed as cave-1e6k.

Generated with Claude Code

### DIFF
--- a/src/components/automations-view.test.ts
+++ b/src/components/automations-view.test.ts
@@ -186,3 +186,21 @@ assert.match(source, /togglePauseAutomation: toggleCodex/, "cron pause/resume re
 assert.match(source, /const req = \+\+runLogReqRef\.current;/, "each log fetch takes a request token");
 assert.match(source, /if \(req !== runLogReqRef\.current\) return;/, "stale log responses are dropped");
 assert.match(source, /runLogReqRef\.current \+= 1;\s*\n\s*setOpenRunId\(null\);/, "closing the log invalidates the in-flight fetch");
+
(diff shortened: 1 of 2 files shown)


BunsDev:
Reviewed during the branch-backlog sweep — **approve** — and pushed a signed merge of current main into this branch (`0170410e`): #2716 and this PR each appended a pin section to `automations-view.test.ts`, resolved keep-both. Typecheck + full suite (571 test files) green on the merged tip.

On the substance: shared `useFocusTrap` adoption matches the calendar pattern (trap + Escape + return-focus), `role="dialog"` + `aria-labelledby` are right, and the deliberate `aria-modal` omission is correctly reasoned — the desktop split-pane keeps the list interactive, so modal semantics would be dishonest. Pins cover both panels and the omission.

Per Val's direction ("commit push and merge after reviewing"), I'll merge once required checks are green.

PR #2731, opened 2026-07-08
fix(bell): popover state on the trigger, aria-pressed toggles, RelativeTime timestamps, pointerdown dismissal (cave-jm6t)

BunsDev:
## Problem (cave-jm6t, from the notification-bell audit)
1. The bell trigger toggled a popover with no `aria-expanded`/`aria-haspopup` — AT users could not tell it opens a panel or whether it is open.
2. Sound-mode chips showed the active mode only visually; the mute toggles relied on `mute`/`muted` text alone.
3. Item timestamps were a bare `relativeTime()` string — no semantic `<time>`, no exact-time hover, not prefs-aware.
4. Outside-click dismissal listened to `mousedown` only.

## Fix
- `aria-expanded={open}` + `aria-haspopup="dialog"` on the trigger.
- `aria-pressed` on sound chips and both mute toggles (+ an aria-label naming the familiar on the settings-list toggle).
- Shared `<RelativeTime>` (semantic `<time>`, exact-date hover title, prefs-aware density) with the `—` fallback; the local string helper is removed. (Note: nothing in the repo ticks relative times on an interval — this adopts the shared component, matching every other surface.)
- `pointerdown` dismissal covers mouse, pen, and touch.

## Verification
- 5 new pins in `notification-bell-mutations.test.ts`
- `pnpm test:app` — 572 files pass · `pnpm typecheck` clean

Closes cave-jm6t (beads).

### DIFF
--- a/src/components/notification-bell-mutations.test.ts
+++ b/src/components/notification-bell-mutations.test.ts
@@ -58,4 +58,31 @@ assert.doesNotMatch(
   "the all-or-nothing Promise.all fan-out must not return",
 );
 
+// ── cave-jm6t: bell a11y/polish ──
+assert.match(
+  src,
+  /aria-expanded=\{open\}\s*\n\s*aria-haspopup="dialog"/,
+  "the bell trigger exposes popover state (aria-expanded) and kind (aria-haspopup)",
+);
+assert.match(
+  src,
+  /key=\{opt\.label\}\s*\n\s*aria-pressed=\{active\}/,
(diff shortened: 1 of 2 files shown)


BunsDev:
Superseded by #2728, which landed the same cave-jm6t fix (aria-expanded/haspopup, aria-pressed toggles, RelativeTime, pointerdown dismissal) while this PR's checks ran.

PR #2736, opened 2026-07-08
fix(projects): expand a leading ~ in a project root, server-side (cave-pgvc)

BunsDev:
## What

`normalizeRoot` only trimmed + slash-normalized — it never expanded a leading `~`. A user who hand-typed `~/code/app` in the New-project root field got a project whose root was the **literal** `~/code/app`, which never matched the absolute `project_root` the daemon reports — so its Sessions/Git/Tasks stayed empty and the project looked **permanently dead**. (The native picker and in-app browser return absolute paths, so only manual typers hit it.)

## Fix

Expand a leading `~` / `~/` against the already-imported `homedir()` before the trailing-slash trim. **Server-side only** (`createProject`/`patchProject` use `normalizeRoot`) so the CLI and API agree; `~user` is left untouched — we don't guess other users' homes. Runs **before** the one-per-root dedup (cave-729h), so the expanded absolute path is what dedup compares.

Split from the Projects-hub audit's cave-psp8 (its picker slice shipped in #2733; the `loadBoardCards` abort + deferred-delete parts remain deferred while `projects-view.tsx` is under active edit). Pinned in `cave-projects.test.ts` (`~/…`, bare `~`, and a non-leading `~` left alone).

## Verification

`typecheck` · **572** app test files · `build` within budget.

### DIFF
--- a/src/lib/cave-projects.test.ts
+++ b/src/lib/cave-projects.test.ts
@@ -66,6 +66,19 @@ try {
   });
   assert.equal(slashHeavy.root, "C:/tmp/slash-heavy");
 
+  // cave-pgvc: a leading ~ expands to the home dir server-side, so a hand-typed
+  // "~/code/app" is stored absolute and can match the daemon's project_root.
+  const home = os.homedir().replace(/\\/g, "/");
+  const tildeProject = await createProject({ name: "Tilde", root: "~/code/app/" });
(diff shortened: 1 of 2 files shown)


BunsDev:
Superseded by #2734 (cave-psp8), which landed the same leading-~ expansion in normalizeRoot before this could merge — now conflicting + redundant. Closing unmerged; the fix is already on main.

PR #2759, opened 2026-07-08
a11y(inbox): urgency-aware toasts that wait for you; audible snooze menus and dashboard actions (cave-bj68, cave-1y0d)

BunsDev:
Two P2 beads from today's Inbox data-plane audit, both source-verified before filing.

## Toast stack (`cave-bj68`)

- **Urgency-aware politeness** — `response-needed` items (a familiar blocked on the user) now render `role=alert` / `aria-live=assertive`; reminders and digests stay `polite`. Previously everything was polite, so time-critical toasts waited in the AT speech queue behind whatever else was talking.
- **Pausable auto-hide (WCAG 2.2.1)** — the 8s timer pauses on pointer hover and while keyboard focus is anywhere inside the toast (focus-within via relatedTarget check), resuming with a full window on leave. The inbox item always persisted underneath; the fix is that the popup can no longer vanish mid-read or mid-interaction.
- **One dismiss, with context** — the redundant header X (identical action to the labelled Dismiss button) is gone, and the remaining button reads "Dismiss notification: <title>".

## Snooze menus + dashboard actions (`cave-1y0d`)

- The **shared `SnoozeMenu`** (used by the toast stack and inspector) had zero menu semantics — bare divs. It now has the app's standard popover contract: `useFocusTrap` (first option focused, Tab cycles, Escape closes and returns focus), `aria-haspopup`/`aria-expanded` on the trigger, labelled `role=menu` with `menuitem` options, focus rings.
- **Audit correction recorded:** the agent claim that ActionInbox's local snooze menu lacked a focus trap was false — it already has trap + roles + backdrop and was left alone. Only the shared component was bare.
- **ActionInbox announces successes** — done/dismiss/snooze (single and bulk) now announce through the shared live region ("Marked done: X", "Snoozed for 1h: 3 items"); row removal was previously the only feedback, which is silent to AT. Errors already announced via `role=alert`. The icon-only dismiss now carries the item title.

## Verification

- `pnpm typecheck` clean; full `node scripts/run-tests.mjs app` — 569 files pass; `check:tests-wired` green.
- Pins: `labels-and-live-regions.test.ts` politeness pin updated to the urgency-conditional form; new `inbox-notifications-a11y.test.ts` pins the timer-pause filter, focus-within resume, single-dismiss invariant, menu semantics, and both announce sites.
- Not manually driven: toast firing requires live SSE reminder events — the changes there are markup/ARIA + timer-filter logic, held by the pins above.

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -475,6 +475,7 @@ export const SUITES = {
     "src/components/chat-switching.test.ts",
     "src/components/chat-view-first-class.test.ts",
     "src/components/labels-and-live-regions.test.ts",
+    "src/components/inbox-notifications-a11y.test.ts",
     "src/components/workspace-inspector-mount.test.ts",
     "src/middleware.test.ts",
     "src/lib/font-settings.test.ts",
--- a/src/components/dashboard/action-inbox.tsx
+++ b/src/components/dashboard/action-inbox.tsx
@@ -8,6 +8,7 @@ import type { InboxItem } from "@/lib/cave-inbox";
 import { KIND_ICON, KIND_LABEL, itemHasTarget, itemHref, relativeTime } from "@/lib/daily-report";
 import { formatTimestamp, readDateTimePrefs, useDateTimePrefs } from "@/lib/datetime-format";
 import { nextItemsAfterAction } from "@/lib/dashboard-model";
+import { useAnnouncer } from "@/components/ui/live-region";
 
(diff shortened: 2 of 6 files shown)


BunsDev:
Superseded by #2757 + #2760 — a sibling session shipped both beads (cave-bj68, cave-1y0d) while this was in CI, including the SnoozeMenu consolidation this PR deferred. Remaining audit work (cave-bzch) continues separately.

PR #2765, opened 2026-07-08
perf(inbox): SSE echo/reconnect guards; dashboard inbox hears the cockpit refetch (cave-bzch)

BunsDev:
The surviving piece of the inbox data-plane audit (cave-bzch), rebuilt on latest main after #2757/#2760 landed the a11y beads and the prefs lock (cave-g6ew) via a sibling session.

- **SSE reconnect snapshots** keep the previous `inboxItems` reference when content-identical (`arrayContentEqual`) — a reconnect used to re-render every `inboxItemsWithEphemeral` consumer with an identical list.
- **Content-identical `updated` echoes** (our own optimistic dismiss/done/snooze writes bounced back by the broadcast) bail before minting a new array.
- **ActionInbox actually hears the cockpit's 30s refetch** — `useState(initialItems)` read the prop exactly once, so items fired/done/dismissed elsewhere stayed in the dashboard widget until remount. The audit bead guessed "subscribe to SSE"; the real mechanism was the ignored prop, and the fix is a content-guarded reconcile effect that also preserves in-flight optimistic removals.

Pinned in `inbox-hygiene.test.ts` (wired into the app suite). Verification: typecheck clean, 571 suite files pass, `check:tests-wired` green.

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -365,6 +365,7 @@ export const SUITES = {
     "src/lib/automations/list-input.test.ts",
     "src/lib/automations/run-status.test.ts",
     "src/lib/inbox-feed.test.ts",
+    "src/lib/inbox-hygiene.test.ts",
     "src/lib/daily-summary-notifications.test.ts",
     "src/components/daily-summary-notifications.test.ts",
     "src/app/daily-report-page.test.ts",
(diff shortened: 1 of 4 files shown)


BunsDev:
Duplicate of #2762 (sibling session, opened 8 minutes earlier, same bead cave-bzch) — and its actedIdsRef reconcile is more robust than the content-guard here (survives a genuinely racing poll, not just no-op re-renders). One piece #2762 doesn't carry: the SSE *snapshot* branch still does an unconditional setInboxItems(e.items) on reconnect — an arrayContentEqual bail there keeps the array identity across reconnect snapshots. Will follow up with that one-liner after #2762 merges.

PR #2780, opened 2026-07-09
a11y(journal): audible mutations, real headings, visible focus, spoken disabled reason (cave-t1ou)

BunsDev:
## Summary

Bead `cave-t1ou` (journal audit, source-verified), four fixes:

**Silent mutations.** The surface had zero `useAnnouncer` usage — generate success, save, and delete-scheduling all changed the page silently for screen-reader users (errors and notices had live regions; success said nothing). All three announce now; the generate button also reports `aria-busy` while reflecting, and the notice toast gains `aria-atomic` so partial updates don't read as fragments.

**No heading structure.** "What happened · \<date\>" and "Reflection" were a span and a div — the detail pane had no outline for AT navigation (WCAG 1.3.1). Both are real `h3`/`h4` now, with a `font: inherit` reset class so the visual design is pixel-unchanged.

**Invisible keyboard focus.** The day-rail rows had no `:focus-visible` rule at all, and the entry actions' focus state merely reused hover colors — indistinguishable from a pointer hover. Both get the standard `--ring-focus` outline, mirroring the `.journal-next__chip` rule that already existed in the same file.

**Hover-only disabled reason.** The generate button's explanation ("today's entry was written by X" / "select today to generate") lived only in `title=` — an sr-only suffix now carries it into the accessible name for AT and keyboard users.

The bead's remaining small riders (filter result-count feedback, stat label association, skeleton roles) are noted there — diminishing returns for this pass.

## Test plan

- [x] Pins: announcer + all three announcements, aria-busy, sr-only reason, both headings, atomic notice, both focus rules

### DIFF
--- a/src/components/journal/journal-entries.test.ts
+++ b/src/components/journal/journal-entries.test.ts
@@ -118,4 +118,17 @@ assert.match(css, /\.journal-day:active \{ transform:/, "day rows have a tactile
 assert.match(css, /\.journal-entry__action:active:not\(:disabled\) \{ transform: scale/, "entry action icons have a tactile press");
 assert.match(css, /@media \(prefers-reduced-motion: reduce\)[\s\S]*?\.journal-next__chip,/, "the new interactions respect prefers-reduced-motion");
 
+// ── a11y: audible mutations, real headings, visible focus (cave-t1ou) ────────
+assert.match(entries, /const \{ announce \} = useAnnouncer\(\)/, "the surface uses the shared announcer");
+assert.match(entries, /announce\("Reflection generated\."\)/, "generate success is announced");
+assert.match(entries, /announce\("Journal entry saved\."\)/, "save success is announced");
+assert.match(entries, /announce\(`Deleting the entry for [\s\S]{0,80}undo available\.`\)/, "delete scheduling is announced");
+assert.match(entries, /aria-busy=\{generating\}/, "the generate button reports busy state");
+assert.match(entries, /unavailable — select today to generate/, "the disabled reason reaches the accessible name (not just title=)");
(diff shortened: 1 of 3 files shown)


BunsDev (commented):
Reviewed against the cave-t1ou audit findings (I filed the bead) — the fix targets the verified mechanisms, not the false positives. ✅ on the substance: announcer wired exactly where the paths were silent (generate/save had no live region), real headings with the inherit-styling reset, ring-token focus rules on both the day rail and entry actions, and the disabled reason reaching the accessible name via sr-only.

One real issue and three nits:

**Double-speak on delete (worth fixing before merge):** `UndoToast` is itself `role="status" aria-live="polite" aria-atomic` (ui/undo-toast.tsx:58) — so the new `announce("Deleting the entry for X — undo available.")` and the toast's own text both hit AT for the same event. This is the exact duplicate-announcement pattern we avoided in the inbox pass (#2760's handoff note). Recommend dropping the delete `announce()` + its pin and letting the UndoToast speak; generate/save announces are correct to keep (nothing else speaks there).

Nits:
- `aria-busy={generating}` was added to the rail button but not the empty-state `<Button>` — same state, same claim.
- The sr-only block pasted into the empty-state Button is mis-indented (cols ~10 inside a col-22 context) — compiles, but reads as a paste artifact.
- If the empty-state generate only renders for today's date, its "select today to generate" sr-only branch is dead code there.

Bead bookkeeping: the audit's small riders (loading skeletons `role=status`, filter result-count feedback, day-preview truncation exposure, stat number/label association) aren't in this PR — totally fine to punt, but please either note them when closing cave-t1ou or spin a follow-up so they don't silently evaporate.

PR #2783, opened 2026-07-09
fix(journal): stop silent write loss + false generation stamps between Grimoire and generate (cave-9f2e)

BunsDev:
## What

The journal is **one file per date**, written by two surfaces — Grimoire's debounced **autosave** and the journal **generate/edit** flow — with **no conflict handling**. Whichever `POST /api/journal` landed second silently discarded the other (a generation could vanish under an autosave mid-flight, and vice versa). The route also stamped `generatedAt: new Date()` on **every** save, so a manual hand-edit read as a fresh generation ("· 2m ago") in the entry meta.

Source-verified in `route.ts` POST + `grimoire-view.tsx` + `journal/journal-entries.tsx`. (cave-9f2e)

## Fix

**Route (`POST /api/journal`):**
- **Optimistic-concurrency guard**, opt-in via `expectedModified`, mirroring the memory-file **409**: read the current entry once; if it changed on disk since the caller loaded it, return **409** with the current record instead of clobbering. Backward-compatible — omitting `expectedModified` skips the check.
- **`generatedAt` is preserved** from the on-disk entry on a manual save (or `null` for a new entry); only an explicit `body.generatedAt` (the generate flow) moves it.

**Clients:**
- Grimoire `JournalMdEditor` captures the loaded mtime, sends it as the baseline, **advances it from each successful save** (so autosave never self-conflicts), and surfaces a 409 rather than overwriting.
- `journal-entries` `generate` stamps `generatedAt` + sends the baseline (409 → friendly notice); `saveEdit` sends the baseline and no `generatedAt` (preserved).

## Verification

New **`journal/route.test.ts`** (wired into `run-tests.mjs`) + grimoire-view / journal-entries source pins. `typecheck` · `check:tests-wired` (768) · journal test suites green · `build` within budget.

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -267,6 +267,7 @@ export const SUITES = {
     "src/lib/slash-prompt.test.ts",
     "src/lib/cave-canvas.test.ts",
     "src/components/journal/journal-entries.test.ts",
+    "src/app/api/journal/route.test.ts",
     "src/lib/journal.test.ts",
     "src/lib/journal-memory-stats.test.ts",
     "src/lib/journal-generate.test.ts",
--- a/src/app/api/journal/route.test.ts
+++ b/src/app/api/journal/route.test.ts
@@ -1,0 +1,49 @@
+// @ts-nocheck
+import assert from "node:assert/strict";
+import { readFile } from "node:fs/promises";
+
+const source = await readFile(new URL("./route.ts", import.meta.url), "utf8");
+
+// ── Optimistic-concurrency guard (cave-9f2e) ─────────────────────────────────
+// The journal is one file per date and two surfaces write it (Grimoire autosave
+// + the generate/edit flow). Without a conflict check the second writer silently
+// drops the first. POST accepts an opt-in `expectedModified` baseline and 409s
+// when the file changed underneath, mirroring the memory-file convention.
+assert.match(
+  source,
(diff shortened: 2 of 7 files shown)


BunsDev (commented):
Reviewed against the cave-9f2e audit findings (I filed the bead) — this is the recommended design executed cleanly: opt-in `expectedModified` 409 mirroring the memory-file convention, `generatedAt` stamped only by the generate flow and preserved on manual saves, baselines threaded through BOTH writers, and the advance-baseline-on-save detail (so autosave never self-conflicts) is exactly the subtlety that makes debounced autosave + optimistic concurrency coexist. The new pinned route test is a welcome addition.

Two notes, neither blocking:

1. **Residual create-create window:** the guard is `expectedModified !== null && current.exists && …`, so two writers both starting from a *nonexistent* entry (both baselines null — e.g. Grimoire opens today's empty doc while the journal tab generates today's first entry) still last-writer-wins on creation. Narrow, and the opt-in shape can't express "expect absent" without a sentinel — fine to accept, but worth a one-line comment in the route so the next reader knows it's a known edge rather than an oversight.

2. **Merge sequencing with #2780:** both PRs append to `journal-entries.test.ts`'s tail and edit adjacent regions of `journal-entries.tsx` — whichever lands second will need a rebase. Suggest merging in whichever order, but budget for the conflict rather than letting required checks flap on a stale base.

The delete-then-write case (baseline sent, entry deleted meanwhile → write recreates) reads as deliberate write-wins-over-delete and pairs fine with the undoable delete UX.

PR #2784, opened 2026-07-09
perf(journal): day reads no longer block on the memory-inventory scan (cave-tgx9)

BunsDev:
## Summary

Bead `cave-tgx9` (source-verified): `GET /api/journal?date=…` awaited the full memory-file inventory in the blocking response for **every day selection** — a directory walk + stat of ~1,900 files warm, and a multi-second head-read scan cold — only to feed the three-number stats block. Grimoire's journal reads and the iOS day fetch paid the same cost without ever reading the stats.

**Route.** The plain day GET is now a single journal-file read. The inventory-derived `stats` + `context` moved to a `?stats=1` branch on the same route (no api-contracts change).

**Surface.** `loadDay` paints the entry immediately (`stats: null` renders "–" placeholders) and fetches `?stats=1` non-blocking; the late block merges only into the same day it was requested for (guarded by the existing request token + a `prev.date === slug` check). If the user hits Generate before the stats land, the handler fetches the context inline — generation always carries the memory-scope note.

**Riders from the bead, verified:**
- `Intl.DateTimeFormat` instances in `daily-report`'s day labels hoisted to module scope — the day rail constructed one per row per render. Memoizing labels into `filteredDays` was deliberately skipped: it freezes "Today"/"Yesterday" past midnight (the #2310 foot-gun).
- `FamiliarStudioJournalTab` lazy-wrapping: **no action** — `settings-shell` is only imported by `app/settings/page.tsx`, so the tab rides the `/settings` route chunk, not the boot bundle.

## Measured (real inventory, 1,890 files, prod build)

| Request | Before | After |
| --- | --- | --- |
| Day entry GET (warm) | blocked on scan (~0.5s+) | **41–77ms** |
| Stats (cold / warm) | inline in every day read | 1.7s / 0.5s on its own request |

## Test plan

- [x] Pins: route must not `Promise.all` the entry read with the scan; `?stats=1` branch; non-blocking client fetch; entry-paints-with-null-stats; same-day merge guard; generate inline-context fallback
- [x] Live probe of both branches on the built worktree (numbers above)

### DIFF
--- a/src/app/api/journal/route.ts
+++ b/src/app/api/journal/route.ts
@@ -19,12 +19,19 @@ export const runtime = "nodejs";
  * Personal Journal — one reflective entry per day.
  *
  *   GET    /api/journal                       → { ok, days: JournalSummary[] }
- *   GET    /api/journal?date=YYYY-MM-DD        → { ok, ...JournalRecord, stats, context }
+ *   GET    /api/journal?date=YYYY-MM-DD        → { ok, ...JournalRecord }
+ *   GET    /api/journal?date=YYYY-MM-DD&stats=1 → { ok, date, stats, context }
  *   POST   /api/journal  body { date, reflection, reflectedBy } → { ok, ...JournalRecord }
  *   DELETE /api/journal?date=YYYY-MM-DD        → { ok, date, deleted }
  *
  * `date` is the only user-controlled input and is gated on a strict
  * `YYYY-MM-DD` real-day guard before any fs access.
+ *
+ * Stats ride their own request: they need the full memory-file inventory
+ * (a stat of ~1900 files warm, a multi-second head-read scan cold), which
+ * used to block EVERY day read — including Grimoire's and iOS's, which
+ * never look at the stats block. The entry response is now a single file
+ * read; the journal surface fetches ?stats=1 after the entry paints.
  */
 export async function GET(req: Request) {
   const { searchParams } = new URL(req.url);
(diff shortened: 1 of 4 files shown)


BunsDev (commented):
Reviewed against the cave-tgx9 audit findings (I filed the bead) — this takes the bead's recommended option (split stats out of the blocking day response) and improves on it twice: Grimoire's and iOS's day reads never used the stats block and now stop paying for the inventory walk entirely, and the date-label rider got a better fix than the bead suggested (hoisted shared `Intl.DateTimeFormat` instances benefit every daily-report consumer, not just the journal rail).

The details that make it correct are all present: the late stats merge is double-guarded (`reqId` recheck after the await AND `prev.date === slug`, matching the repo's stale-response convention), generate falls back to an inline context fetch when the user beats the stats request (pinned), and the null gap renders honest "–" placeholders with no layout shift.

Two notes, neither blocking:

1. **Degradation nuance worth a comment:** if the inline context fetch in `generate` also fails, generation now proceeds with `context: ""` — previously a stats failure failed the whole day load, so generation couldn't run context-less. Proceeding is probably the right call (a reflection without the memory-scope note beats no reflection), but it's a behavior change; one line in the code saying it's deliberate would stop a future reader from "fixing" it.

2. **Three-way merge sequencing:** #2780, #2783, and this all append to `journal-entries.test.ts`'s tail and edit overlapping regions of `journal-entries.tsx`; #2783 and this both rewrite the route's GET/POST area. All three diffs are against the same base, so landing them means two rebases — worth serializing deliberately rather than letting required checks flap.

Accepted-as-known: the cold ~25s inventory scan still happens on the first `?stats=1` — but it's off the critical path now, which is exactly what the bead asked for.

PR #2863, opened 2026-07-09
feat(queue): rename Work queue → Queue; resilient load when one source fails

BunsDev:
## What

- **Rename**: user-facing "Work queue" → **"Queue"** — Tasks segment tab, surface `<h1>`, workspace mode title, SR announcements, refresh aria-label.
- **Ensure it loads**: the gh PR bridge was the queue's hard dependency — if `gh` failed, the whole surface showed a dead error state even with healthy beads. `fetchQueue` now degrades: beads-only when the PR bridge fails (new truthful warn banner, raw error in tooltip), PRs-only when beads fail (existing). Only both failing shows the retry empty state.
- Resolves a leftover `.beads/interactions.jsonl` pull conflict (union; keeps the cave-i7wf closure interaction).

## Verification

- `beads-work-queue.test.ts` updated to pin the rename + resilience — passes
- `tsc --noEmit` clean
- Live check: `/api/beads?mode=ready` and `/api/beads/prs` both return ok on :3000; queue renders with data


BunsDev:
Superseded by #2867, which re-landed this branch's rename + resilient-load commit byte-for-byte on current main together with the bridge refresh-failure fix and the first-load degradation e2e. Merged as c379c9e1.

PR #2874, opened 2026-07-09
refactor(chat): slim the chat header to Find + kebab with compact project selection

BunsDev:
## What

Slims the open-chat header from seven competing controls down to **Find + the ⋮ kebab** (plus the hover-reveal "link a task" chip).

### Before
Find · thinking toggle · debug (duplicated in kebab) · delete · kebab (with a full inline list of every project) · reflect · link-task

### After
- **Find + kebab only** stay inline. Thinking toggle, reflect, debug, and delete all live in `SessionOverflowMenu`; delete keeps its two-step guard as an in-popover confirm view that disarms when the menu closes.
- **Project selection is one compact row** — `Project: <name>` — that opens the shared searchable picker anchored to the kebab trigger, replacing the inline list of every registered project.
- Extracted a controlled `ProjectPickerPopover` from `ProjectPicker` (public API unchanged; home composer and chat empty state untouched).
- Removed dead `.cave-chat-icon-button` / `.cave-chat-delete-trigger` CSS.

## Testing

- Updated source-shape tests: `chat-header-row`, `chat-view-polish`, `chat-view`, `task-chat-cwd`; `project-picker` + `chat-surface-polish` pass unchanged.
- `tsc --noEmit` clean; full `pnpm test:app` green (594 test files).
- No e2e selectors reference the removed controls.

Out of scope: mobile header, MetaLine, empty-state/home-composer pickers.

### DIFF
--- a/src/components/chat-view.tsx
+++ b/src/components/chat-view.tsx
@@ -119,7 +119,7 @@ import { handlePlaceholderTab } from "@/lib/prompt-placeholders";
 import { recordPromptRecent } from "@/lib/prompt-prefs";
 import { SaveTemplateModal } from "@/components/save-template-modal";
 import { readComposerDraft, useDraftPersistence } from "@/lib/use-composer-draft";
-import { ProjectPicker, useAddProjectFlow } from "@/components/project-picker";
+import { ProjectPickerPopover, useAddProjectFlow } from "@/components/project-picker";
 import { toolArgDetail, toolArgSummary } from "@/lib/tool-arg-summary";
 import { toolVisual } from "@/lib/tool-visual";
 import { toolReadableFields, prettyToolOutput, type ReadableField } from "@/lib/tool-readable";
@@ -791,11 +791,14 @@ function splitReasoning(text: string): { visible: string; reasoning: string } {
 // view arms/executes its "Start a task" card-follows-chat flow (see
 // handleEvent's "session" case).
 
-/** Codex/ChatGPT-style overflow menu. Collapses the session's secondary
- *  controls — project switch, voice call, debug — into a single kebab so the
- *  header reads as title + quiet metadata instead of a row of competing icons.
- *  Find and Delete stay inline (one-click); everything else lives one click away
- *  here. */
+/** Codex/ChatGPT-style overflow menu. Collapses ALL of the session's secondary
+ *  controls — project selection, thinking toggle, reflect, voice call, debug,
+ *  delete — into a single kebab so the header reads as title + quiet metadata
+ *  plus Find, instead of a row of competing icons. Project selection is one
+ *  compact "Project: <name>" row that opens the shared searchable picker
+ *  (anchored to the same kebab trigger), not an inline list of every project.
+ *  Delete keeps its two-step guard: the danger item swaps the menu body to a
+ *  confirm view before anything commits. */
 function SessionOverflowMenu({
   projects,
   projectId,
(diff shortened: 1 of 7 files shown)


src/components/chat-view.tsx
```diff
@@ -851,208 +872,121 @@ function SessionOverflowMenu({
...
-          >
-            Debug session
-          </PopoverItem>
-        </PopoverBody>
+            {projects.length > 0 || onAddProject ? (
+              <PopoverItem
+                icon="ph:folder"
+                title={activeProject?.root ?? "No project"}
+                onSelect={() => {
+                  // Chain popovers: the kebab closes on this click; the picker
```
BunsDev, line 916: Optional hardening (not blocking): the kebab and this project picker share `anchorRef={triggerRef}` and chain via `close()` → `setProjectPickerOpen(true)`. The reasoning holds *as long as* `Popover` binds its outside-click listener in a post-render effect (so it misses this same click) — but it's fragile if that ever changes to a synchronous `mousedown` bind. The delete path in this very menu uses an in-place body-swap (`confirmingDelete`) that avoids the two-popover timing dance entirely; applying that same pattern to the project list would be more robust and internally consistent.

src/components/project-picker.tsx
```diff
@@ -91,6 +91,142 @@ export function useAddProjectFlow(args: {
   return { beginAddProject, addProjectModal, adding, addError };
 }
 
+/** Resolve the effective selection: NO_PROJECT_ID → none; null → first project. */
+function selectedProject(value: string | null, sorted: CaveProject[]): CaveProject | null {
+  return value === NO_PROJECT_ID
+    ? null
+    : (value ? sorted.find((project) => project.id === value) ?? sorted[0] : sorted[0]) ?? null;
```
BunsDev, line 98: Now that the chat list is scoped to the familiar's grants (`useProjects({ familiarId })` in chat-view), this `?? sorted[0]` fallback means a session whose stored project isn't in the familiar's granted set will silently render as the *first granted* project instead of its real one. Edge case — a session's project is normally already granted to its familiar; only bites after a grant is revoked or the familiar is swapped. Consider showing the real project even when out-of-scope, or a "not available for this familiar" state. Low severity.

src/components/chat-view.tsx
```diff
@@ -2284,7 +2218,10 @@ export const ChatView = forwardRef<ChatViewHandle, Props>(function ChatView(
   // Two-step delete via the header trash button: it opens a confirm popover and
   // only the explicit Delete commits (HeaderDeleteButton owns the armed state).
```
BunsDev, line 2219: Stale comment: there's no longer a "header trash button" / `HeaderDeleteButton` — delete moved into `SessionOverflowMenu` (armed via `confirmingDelete`). Worth updating this to match. (Same lingering reference at :4520, just outside this diff.)

BunsDev (commented):
Automated review of the header-slim refactor. Clean and well-scoped — nice call splitting the command-palette / new-card-modal familiar-scoping into a separate change, and the pinned chat tests (`chat-header-row`, `chat-view`, `chat-view-polish`, `task-chat-cwd`) are carefully migrated to the new kebab structure. Frontend build / Rust check / CodeQL are green.

A few optional, low-severity notes inline. One extra that's outside this diff's hunks: a stale `HeaderDeleteButton` comment also lingers at `chat-view.tsx:4520` ("delete confirm resets itself — HeaderDeleteButton is keyed on sessionId") — worth updating alongside the one flagged at :2218. Nothing blocking. 🤖

PR #2876, opened 2026-07-09
fix(chat): right sidebar mirrors the left panel's glass — flush fit, whole tab labels (cave-wc78)

BunsDev:
## Summary

The chat right sidebar (Inspector / Debug / Changes) had no chrome of its own: a transparent aside that sometimes opened a dark gap at the far right, opaque `bg-raised` header bands, a double border+tint from the InspectorPane inside it, and md fill-tabs that shredded their labels ("Analyt…", "Autom…") at the 230px default width.

Now it's the mirrored twin of the left `.shell-nav`:

- **`.chat-right-aside` owns the panel chrome** — the same 88% bg-raised glass with `blur(14px) saturate(140%)`, hairline flipped to the panel's left edge, flush against the window edge, with `@supports` and `prefers-reduced-transparency` fallbacks
- **Tab bar + Changes header go translucent** so the glass reads through; InspectorPane's non-compact shell drops its own border/tint (single source of chrome, no double band)
- **Sidebar defaults to 260px (min 220px)** and the Inspector's section tabs use `size="sm"` — Familiar/Analytics/Automations all render whole

## Verification

- Playwright measurement against the running app (`?demo=1`, 1512×900 @2x): aside computes the **exact same glass color as `.shell-nav`** (`oklch(0.165 0.025 293 / 0.88)`), `gapToEdge: 0`, and all three Inspector section labels render untruncated after fonts load
- `pnpm test:app` — full suite green
- `npx tsc --noEmit` — clean

Bead: cave-wc78

### DIFF
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -3829,7 +3829,8 @@ textarea.required-inputs-control {
   background: var(--bg-hover);
 }
 
-/* Open tab bar (sits at top of the open aside) */
+/* Open tab bar (sits at top of the open aside) — translucent so the aside's
+   glass shows through, mirroring the left sidebar's chrome. */
 .right-panel-tabs {
   display: flex;
   align-items: center;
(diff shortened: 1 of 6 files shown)


BunsDev:
Duplicate of #2877 (same branch, opened concurrently by another session — that one is linked from bead cave-wc78).

PR #3000, opened 2026-07-12
fix: tolerate legacy familiar records without descriptions

CompleteDotTech:
## Summary

Prevent Cave from writing unreadable blank-description familiar records and preserve daemon error context.

## Root Cause

Cave could omit `description`, which older Coven parsed as required. When that happened, Cave discarded structured daemon error details and a `COVEN_HOME` override could split daemon roster state from Cave writes.

## Changes

- Always serialize blank `description = ""`.
- Propagate the structured daemon error message in Cave's stable `503` roster response.
- Route familiar lifecycle, config, and onboarding storage through `COVEN_HOME`.
- Add executable fake-daemon and isolated-home regression coverage.

## Validation

- `pnpm test:api` (`144/144`) in a clean source worktree
- `pnpm exec tsc --noEmit`
- Test wiring check (`834` wired)
- Windows direct route test with the alias loader
- Windows source-level UI loaded Ash and Briar with blank descriptions; human verified.

## Note

The Windows full API suite stops at an unrelated `beads-pr-bridge` path-spawn issue. The full same API suite passed in the clean WSL worktree, and the targeted Windows route test passed.

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -887,6 +887,7 @@ const STRIP_TYPES_MJS = new Set([
 // Tests whose import graph reaches the "@/..." path alias and therefore need
 // the alias-resolving loader (`scripts/test-alias-register.mjs`).
 const ALIAS_LOADER = new Set([
+  "src/app/api/familiars/route.test.ts",
   "src/app/api/prompts/route.test.ts",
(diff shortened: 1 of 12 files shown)


CompleteDotTech:
Superseded by a narrower Cave-side description requirement.

PR #3003, opened 2026-07-12
fix(onboarding): derive OpenCoven setup actions

BunsDev:
## Summary
- derive the OpenCoven setup CTA, manual command, and install targets from actual `coven` / `coven-code` status
- show `npm i -g @opencoven/cli@latest coven-code@latest` for fresh installs, but update/install only the missing or outdated tool when one side is already current
- align the fallback setup copy with the two-tool requirement and add regression coverage

## Root Cause
The onboarding setup step described installing both OpenCoven tools, but the manual command and primary action were still effectively hardcoded around the coven CLI. On Windows this could present an invalid setup path: a current Coven Code install plus an outdated coven CLI still showed the wrong combined action/command state.

## Validation
- `node --experimental-strip-types src/lib/opencoven-tools-install.test.ts`
- `node --experimental-strip-types src/components/onboarding-guided-steps.test.ts`
- `node --experimental-strip-types src/components/onboarding-polish.test.ts`
- `pnpm check:tests-wired`
- `pnpm exec tsc --noEmit --pretty false`
- `git diff --check HEAD~1..HEAD`
- `node scripts/run-tests.mjs app`

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -51,6 +51,7 @@ export const SUITES = {
     "src-tauri/permissions/desktop-permissions.test.mjs",
     "src-tauri/release-runtime.test.mjs",
     "src/components/open-coven-tools-update.test.ts",
+    "src/lib/opencoven-tools-install.test.ts",
     "src/lib/workflow-run-prompt.test.ts",
     "src/lib/workflow-step-progress.test.ts",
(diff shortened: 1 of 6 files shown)


BunsDev:
Closing as superseded by #3004, which landed the same setup-command fix on main. This branch also picked up two Copilot Autofix commits after review, but #3004 already shipped the intended patch through protected main.

PR #3026, opened 2026-07-12
Fix Windows Coven Code version detection

CompleteDotTech:
## Summary
- Resolve Windows npm `.cmd` shims from their quoted `%dp0`/`%~dp0` entry point without inferring a package path.
- Report an explicit unknown version when a shim target cannot be proven, rather than probing Coven CLI.
- Add Windows packaged-server coverage for `where` launcher shadows, JavaScript and extensionless targets, `.bat`, `.exe`, and missing shims.

## Root cause
Coven Code's extensionless npm launcher target bypassed the old JavaScript-only parser, which then fell back to `@opencoven/cli/bin/coven.js`. The About page consequently displayed Coven CLI's version for Coven Code.

## Verification
- Targeted Coven launcher, tool-status, and route tests
- `pnpm typecheck`
- `pnpm check:tests-wired`
- `pnpm test:conformance` on Windows
- `pnpm build`

`pnpm test:api` was also run. It is currently blocked by an unrelated Windows-only fixture issue in `scripts/beads-pr-bridge.test.mjs`, which uses `:` instead of the Windows PATH delimiter for its nested process.

### DIFF
--- a/scripts/cross-environment.test.ts
+++ b/scripts/cross-environment.test.ts
@@ -135,6 +135,29 @@ function skip(reason: string): void {
   );
 
   // Host branch — the genuinely per-OS assertion.
+  // Coven Code's npm target is extensionless. It must resolve from its own
+  // shim, never by inferring the Coven CLI's conventional JavaScript path.
+  const codeShimScript = path.join(shimDir, "node_modules", "@opencoven", "coven-code", "bin", "coven-code");
+  mkdirSync(path.dirname(codeShimScript), { recursive: true });
+  writeFileSync(codeShimScript, "console.log('coven-code');\n");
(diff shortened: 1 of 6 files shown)


CompleteDotTech:
@bunsdev could you review this Windows launcher-resolution fix? It removes the unsafe Coven CLI fallback and adds coverage for real npm cmd shadows plus the Coven Code extensionless target.

BunsDev (approved):
Reviewed the Windows launcher-selection and shim-target paths, including extensionless Coven Code targets and the unresolved-shim fail-closed behavior. Local targeted tests, conformance coverage, typecheck, and test wiring pass; CI is green and there are no unresolved conversations.

PR #3027, opened 2026-07-12
Fix daemon recovery after Coven CLI updates

CompleteDotTech:
## Summary

- Track the local daemon's pre-update health and restart it only when it was running before a Coven CLI update.
- Replace raw PID termination with verified graceful shutdown, so a stale or reused PID cannot kill an unrelated process.
- Refresh CLI executable discovery before recovery, verify daemon health afterward, and surface lifecycle progress and failures in About.
- Add lifecycle state-matrix coverage for running, stopped, failed-stop, failed-install recovery, restart failure, and PID reuse paths.

## Root cause

The one-click CLI updater stopped the daemon to unlock the executable, but considered a binary on PATH a complete update. It neither restored the daemon nor reported its health, so Cave could silently remain offline.

## Validation

- `pnpm typecheck`
- `pnpm build`
- `pnpm check:tests-wired`
- Focused installer, daemon-start, executable-discovery, lifecycle, and About-panel tests

`pnpm test:api` is blocked by this Windows environment after unrelated setup checks: `scripts/beads-pr-bridge.test.mjs` attempts to spawn a missing `C:\Program Files\nodejs\node.exe`.


CompleteDotTech:
@bunsdev Could you review this daemon recovery update when you have a moment? All CI checks are green.

BunsDev (changes requested):
Two blockers before merge:

1. `src/app/api/onboarding/install/route.ts` regresses the user-facing em dash to mojibake (`â€”`) in the PATH verification error. Restore the Unicode em dash (or plain `--`).
2. The fork head now conflicts with merged #3028. Reconcile onto current `main` while preserving #3028's atomic global npm lease/cancellation cleanup and this PR's graceful local-daemon stop/recovery lifecycle.

The daemon state machine itself is well scoped and avoids the stale/reused PID kill hazard; after reconciliation I will rerun the combined route/lifecycle tests.

BunsDev (approved):
Re-reviewed the integrated head at 5507fa0b. The mojibake is fixed; the merged runner preserves the atomic global npm lane and recovers the local daemon before releasing the lease on cancellation, timeout, spawn failure, and normal completion. Local focused tests, typecheck, test wiring, diff check, and the full 633-file app suite pass. Refreshed CI is fully green across frontend, E2E, Rust, sidecar, and cross-environment jobs. No unresolved review threads.

PR #3028, opened 2026-07-12
Fix global npm update concurrency

CompleteDotTech:
## What changed

- Reserve a process-wide global npm lane after asynchronous installer preparation, so only one `npm install -g` can start across all allowlisted npm targets.
- Return a retryable, specific `409` response for a competing request; script installers remain independent.
- Expose the global lane through the install API and make onboarding and Settings disable/describe other npm updates while it is busy.
- Add cancellation and lifecycle cleanup so errors, timeouts, polling, and HMR/module reloads cannot strand the lane.

## Why

The pre-check occurred before `spawnPlanFor`, while the post-await check only covered the same target. Concurrent Coven CLI and Coven Code requests could therefore both spawn npm against the same global prefix.

## Validation

- `pnpm typecheck`
- Global npm lane concurrency/lifecycle tests, installer-route tests, and onboarding/Settings UI behavior tests
- `pnpm check:tests-wired`
- `pnpm exec next build`
- `pnpm build:server`
- `pnpm test:bundle`

Full suites were attempted. `pnpm test:api` is blocked by the host's missing WSL `/bin/bash`; `pnpm test:app` is blocked by a missing `python3` executable in this Windows environment.

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -666,6 +666,7 @@ export const SUITES = {
     "src/lib/coven-paths.test.ts",
     "src/lib/server/cave-home-migration.test.ts",
     "src/lib/server/cave-home-migration-status.test.ts",
+    "src/lib/server/global-npm-install-lane.test.ts",
     "src/app/api/cave-home-migration/route.test.ts",
     "src/lib/server/agent-attachments.test.ts",
     "src/app/api/api-contracts.test.ts",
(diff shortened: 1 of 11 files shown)


CompleteDotTech:
@bunsdev, could you please review this global npm update concurrency fix? The full CI matrix is green.

BunsDev (approved):
Reviewed the process-wide npm lease, job finalization/cancellation paths, and both onboarding/Settings lane consumers. The reservation is atomic after asynchronous plan preparation, HMR-safe, and released on every terminal path. Targeted lane/route/UI/API-contract tests, typecheck, and test wiring pass; CI is green and there are no unresolved conversations.

PR #3029, opened 2026-07-12
Fix OpenCoven tool update verification

CompleteDotTech:
## Summary

- Verify post-install OpenCoven tool discovery against the expected npm package, launcher target, version probe, compatibility floor, and npm latest version.
- Preserve Windows PATH precedence when selecting npm launchers and reject unreadable or mismatched shims.
- Hold the UI result until its post-install status recheck agrees, then show exact verified path/version diagnostics or a recovery message.

## Root cause

The installer treated `npm install -g` exit code zero plus any matching PATH binary as success. A stale or shadowed launcher could therefore produce a green result before the independent status refresh showed the tool still needed an update.

## User impact

A green update result now means the requested executable is discoverable, belongs to the expected package, runs its version probe, meets Cave's compatibility floor, and satisfies npm latest. Shadowed, wrong-package, unreadable, and below-floor results stay actionable failures with support-safe diagnostics.

## Validation

- `pnpm typecheck`
- Focused installer, status, UI, verification, and Windows launcher tests
- `pnpm check:tests-wired`
- `pnpm build`
- Attempted `pnpm test:app`; it reached 70 passing tests before the unrelated Windows marketplace-sync test failed because `python3` resolves to the Windows Store alias (and its production check has separate path-normalization failures).


CompleteDotTech:
@bunsdev Could you review this when you have a moment? The PR hardens OpenCoven tool updates so success is reported only after the intended executable, package identity, and version are verified. All CI checks are passing.

BunsDev (changes requested):
Integration blockers before merge:

1. Reconcile onto current `main`, preserving #3026's stricter `.cmd`/`.bat` parser and explicit unresolved-shim signal; this branch overlaps `coven-bin.ts` and currently conflicts with that merged fix.
2. Preserve #3028's atomic global npm lane and cancellation-safe finalizers while adding post-install verification.
3. The new `executablePath` and `packagePath` status fields are machine-local paths. When combined with the safe About diagnostics work in #3030, they must be explicitly omitted from copied diagnostics rather than passing through structural object spread. Add a regression assertion covering both fields.

The package/entry-point/latest-version verification model is otherwise sound; I will rerun the integrated Windows launcher and installer verification coverage after reconciliation.

BunsDev (approved):
Verified the current-main integration. Package identity, executable/version checks, npm freshness, Windows shim handling, install serialization, daemon recovery, safe diagnostics, and truthful UI states are reconciled. Local exact-head gates and CI run 29215832507 are green; no review threads remain.

PR #3030, opened 2026-07-12
fix(settings): surface truthful update states

CompleteDotTech:
## What changed

- Treat a GitHub release-route error body, timeout, or malformed response as an unavailable update check instead of confirmed currency; retain and label a prior successful result as last-known data.
- Make Settings/About show compact, retryable daemon states for running, stopped, unreachable, and failed-to-check.
- Preserve stale tool rows visibly with a last-successful-check timestamp and error, and revalidate/dismiss tool-update banners after an in-page update completes.
- Make copied tool diagnostics transparent and safe by disclosing included/excluded data and omitting paths, commands, raw installer output, URL queries, and secret-like values.

## Why

The release endpoint returns HTTP 200 with an `error` body on GitHub failures. The old resolver treated that response as "Up to date," while daemon fetch errors disappeared behind an em dash and failed tool refreshes looked fresh.

## Validation

- `pnpm typecheck`
- `pnpm check:tests-wired`
- focused update, daemon, diagnostics, and component tests
- `pnpm build`
- Tauri debug compile/MSI creation

`pnpm test:app` is blocked by an unrelated marketplace test requiring a `python` executable. Final Tauri updater signing requires `TAURI_SIGNING_PRIVATE_KEY`, which is intentionally not available in this environment.


CompleteDotTech:
@bunsdev could you please review this when you have a moment? The full CI matrix is green, including E2E, frontend, Rust, and cross-platform sidecar checks.

BunsDev (changes requested):
One privacy blocker plus current-main reconciliation:

1. `sanitizeAboutDiagnosticText()` only consumes a local path until the first whitespace. A path such as `/Users/Timothy Gregg/.npmrc` or `C:\Users\Timothy Gregg\secret.log` leaves the suffix in copied diagnostics, despite the UI promise that local paths are omitted. Add a failing regression case with spaces and make path omission consume the complete quoted/unquoted path safely.
2. Reconcile the `open-coven-tools-update.tsx` changes with merged #3028 so the process-wide npm lane/busy UI remains intact.

The release-check classification, daemon-state distinctions, stale-status timestamps, and native-update coordination all reviewed cleanly; targeted tests/typecheck/test wiring pass on the current head.

BunsDev (approved):
Requested changes are addressed on c88876cc: safe diagnostics now use an explicit field allowlist, local-path text fails closed at the first path boundary (including paths with spaces), and the #3028 global npm lane UI/polling is preserved. Local verification: 632/632 app test files, targeted update/daemon/diagnostics/lane tests, typecheck, test wiring, and diff check all pass. Waiting only on the refreshed CI matrix before merge.

PR #3035, opened 2026-07-12
feat(inbox): manageable, dynamic notifications — read/unread, bulk actions, kind muting (cave-uu2d)

BunsDev:
## Why

Incoming inbox notifications were effectively unmanageable:

- **Badge and list disagreed by design** — the bell badge showed the *polled escalations count* while the popover listed *inbox items* (two unrelated data sources).
- **Nothing quieted the badge** short of destroying items one by one.
- **Toasts destroyed notifications**: the 8s auto-hide routed through the same handler as the explicit ✕, which POSTed `/dismiss` — so any notification that fired while you were present silently resolved itself 8 seconds later. The bell mostly only ever showed items that fired while the app was closed.
- **"Clear all" fanned out N racing POSTs** (N file rewrites + N SSE reconciles).

## What

**Read/unread model** — `InboxItem.readAt`: acknowledged-not-resolved. Absent = unread (pre-upgrade items keep their old badge semantics). The scheduler clears it on (re)fire so a read-then-snoozed reminder returns unread; recurrence siblings start unread.

**Bulk actions** — `POST /api/inbox/bulk {action: read|unread|dismiss|done|delete, ids?|all?}`: one locked load→save cycle, per-item SSE broadcasts (no new event type; existing clients reconcile). `all:true` only sweeps the non-destructive stack; `done`/`delete` require explicit ids. Local-origin guarded.

**Bell**:
- Badge = **live unread count derived from the same items the list shows** (`unreadInboxCount`, one shared definition), streamed over the existing SSE.
- Kind **filter chips with counts** (All / Waiting / Reminders / Familiars / Reports — only kinds that have items).
- **Mark read** header action (atomic bulk), per-row **Read** button + unread accent dots, **Open acknowledges**, Clear all = one bulk dismiss.

**Toasts** — explicit ✕ now *acknowledges* (marks read; the item stays in the bell). Auto-hide expiry is visual-only via a new `onExpire` prop — never a server write.

**Kind muting** — prefs gain `mutedKinds` (reminder / agent / daily-summary) mirroring the per-familiar mute: quiet delivery (no toast/native/sound) while items still land in the inbox and bell. `response-needed` is deliberately not mutable (a reply request clears by replying). Kind list lives in pure `inbox-prefs-shape.ts` so the client bell imports the value without dragging `fs/promises` into the browser bundle.

## Verification

- `node scripts/run-tests.mjs all` — ✓ 855 test files (app 628 · api 148 · mobile 77 · conformance)
- `pnpm typecheck`, `pnpm check:tests-wired`, `pnpm build` (bundle within budget) — all green
- New behavioral test `cave-inbox-bulk.test.ts` runs the bulk helper against a throwaway `COVEN_CAVE_HOME`; unread helpers covered in `inbox-feed.test.ts`; prefs pins for `toggleMuteKind` atomicity + kind validation; bell pins updated to the bulk endpoints; `/inbox/bulk` added to the API contracts.
- Verified in-app (prod server + Playwright, mocked SSE): badge "2 unread", chips `All 3 · Reminders 1 · Familiars 1 · Reports 1`, filter narrows to 1 row, Quiet kinds settings render muted state.

Bead: cave-uu2d

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -62,6 +62,7 @@ export const SUITES = {
     "src/lib/familiar-roster-guard.test.ts",
     "src/lib/cave-projects.test.ts",
     "src/lib/cave-inbox-prefs.test.ts",
+    "src/lib/cave-inbox-bulk.test.ts",
     "src/lib/project-permissions.test.ts",
     "src/lib/project-icon-prompt.test.ts",
     "src/lib/project-icon-image-provider.test.ts",
@@ -899,6 +900,7 @@ const STRIP_TYPES_MJS = new Set([
 // Tests whose import graph reaches the "@/..." path alias and therefore need
 // the alias-resolving loader (`scripts/test-alias-register.mjs`).
 const ALIAS_LOADER = new Set([
+  "src/lib/cave-inbox-bulk.test.ts",
   "src/app/api/prompts/route.test.ts",
   "src/app/api/marketplace/pack-prompts-route.test.ts",
   "src/lib/cave-backdrop.test.ts",
--- a/src/app/api/api-contracts.test.ts
+++ b/src/app/api/api-contracts.test.ts
@@ -96,6 +96,7 @@ const contracts: RouteContract[] = [
   { route: "/inbox/[id]/done", methods: ["POST"], kind: "json", localOriginGuard: true },
   { route: "/inbox/[id]", methods: ["PATCH", "DELETE"], kind: "json", readsJson: true, invalidJson: "guarded", localOriginGuard: true },
   { route: "/inbox/[id]/snooze", methods: ["POST"], kind: "json", readsJson: true, invalidJson: "guarded", localOriginGuard: true },
+  { route: "/inbox/bulk", methods: ["POST"], kind: "json", readsJson: true, invalidJson: "guarded", localOriginGuard: true },
   { route: "/inbox/daily-summary", methods: ["POST"], kind: "json", readsJson: true, invalidJson: "fallback-empty", localOriginGuard: true },
   { route: "/inbox/prefs", methods: ["GET", "PATCH"], kind: "json", readsJson: true, invalidJson: "guarded", localOriginGuard: true },
   { route: "/inbox", methods: ["GET", "POST"], kind: "json", readsJson: true, invalidJson: "guarded", localOriginGuard: true },
--- a/src/app/api/inbox/bulk/route.ts
+++ b/src/app/api/inbox/bulk/route.ts
@@ -1,0 +1,63 @@
+import { NextResponse } from "next/server";
+import { isLocalOrigin } from "@/lib/server/local-origin";
+import { applyBulkAction, type BulkAction } from "@/lib/cave-inbox";
+import {
+  broadcastDeleted,
+  broadcastUpdated,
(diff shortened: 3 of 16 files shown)


BunsDev:
## Comprehensive review (retrospective — PR merged at 23:23Z)

Reviewed post-merge as part of the 3031/3034/3035/3036 review sweep. Verdict: **solid work — would have approved**. No regressions spotted that need follow-up beyond the two small notes at the end.

### What this does
Makes notifications manageable: per-item read/unread (`readAt`), one atomic `/api/inbox/bulk` endpoint (read/unread/dismiss/done/delete), kind-level muting (`mutedKinds`), kind filter chips in the bell, and — the sleeper fix — **stops toast auto-hide from destroying notifications**.

### Strengths
- **The toast-expiry fix is a real bug kill.** Previously the auto-hide timer routed through the same handler as the explicit ✕, which POSTed `/dismiss` — so every notification that fired while you were present silently *resolved itself* after 8 seconds. Now expiry (`onExpire`) removes only the visual and leaves the item unread; explicit ✕ marks read (seen ≠ resolved); the distinction is documented on the prop. Exactly right semantics.
- **One unread definition.** `isInboxItemUnread` + `unreadInboxCount` feed both the badge and the list, replacing the old polled-escalations badge that routinely disagreed with the list it sat above. Test coverage includes the pre-upgrade migration posture (missing `readAt` ⇒ unread, matching the old count-everything badge).
- **Bulk is genuinely atomic.** `applyBulkAction` does one locked load→save cycle (vs. the old N-fetch fan-out racing its own SSE reconciles), reuses the existing per-item `updated`/`deleted` SSE events (no new event type for clients to learn), skips unknown ids (deleted-elsewhere mid-flight isn't an error), and is idempotent (second read-all updates nothing).
- **Destructive guard rails.** `all: true` is only accepted for `read`/`unread`/`dismiss`; `done`/`delete` must name ids — enforced in the route, documented at `eligibleForAll`. Scheduled pending items are excluded from sweeps (test-pinned).
- **Refire semantics.** The scheduler clears `readAt` on (re)fire so a read-then-snoozed reminder demands attention again — pinned by test.
- **Kind muting is quiet delivery, not suppression** (still lands in inbox/bell; skips toast/native/sound), `response-needed` is deliberately not mutable (clears by replying), and `mutedKinds` is validated on both load and patch so a hand-edited prefs file can't smuggle unknown kinds. `toggleMuteKind` follows the same single-lock atomicity contract as `toggleMute`.
- **Client-bundle hygiene**: `inbox-prefs-shape.ts` is a pure module (test-enforced: no `node:` imports) so the bell can import `MUTABLE_KINDS` without dragging `fs/promises` into the browser bundle.
- Ephemeral `eph:*` / `missed-*` synthetic ids are consistently skipped by markRead paths; `/api/inbox/bulk` is registered in the API contracts test with the local-origin guard.

### Minor notes (non-blocking)
1. `markInboxItemRead` in `workspace.tsx` is fire-and-forget (no `res.ok` check), unlike the bell's `mutate()`-guarded mutations. Worst case an item stays unread — self-healing and low-stakes, but worth aligning if it ever grows consequences.
2. In the bulk route, a request carrying *both* `ids` and `all: true` silently prefers `ids` — fine, but a 400 would be stricter contract hygiene.

PR #3036, opened 2026-07-12
feat(ios): modern-Apple polish — six optimization vectors for chat + shared chrome (cave-v6eq)

BunsDev:
## Objective

Identify 4–6 ways to optimize the chat interface and the iOS app's components for a sleek, modern, glassmorphic native-Apple experience — then implement them.

## The six vectors

| # | Vector | What changed | Why it's more native |
|---|--------|--------------|---------------------|
| 1 | **Zero-latency haptics** | `Haptics.swift` now caches one generator per impact style + a shared notification generator, and re-`prepare()`s after every fire | Apple's documented Taptic pattern: no first-tap engine spin-up latency, no per-tap allocation churn (40+ call sites app-wide) |
| 2 | **Streaming-scroll contract** | Token streaming only auto-scrolls while the reader is parked at the bottom; new messages reveal only at bottom **or** for your own send; `defaultScrollAnchor(.bottom, for: .initialOffset)` opens flash-free at the latest message | Messages never yanks you down while you're re-reading; returning to the bottom re-engages following via the existing `atBottom` tracking |
| 3 | **Glassmorphic composer** | The field becomes a frosted `.control` glass capsule (same treatment as the drawer's search field) with the accent halo while focused/recording | The composer was the one hairline-only control on the glass bar; now it follows the app's own design language |
| 4 | **Modern bubbles** | User bubbles: iMessage-style vertical accent gradient + luminance-aware `ChromePalette.accentForeground` (iOS mirror of `--accent-presence-foreground`). Assistant bubbles track `chrome.bgRaised`. New bubbles rise-and-fade in | **Fixes a real contrast bug** — white text was hard-coded on the accent, unreadable with light theme accents; assistant bubbles now follow the desktop theme instead of fixed `secondarySystemBackground` (fallback palette is identical, so no visual change untthemed) |
| 5 | **PhaseAnimator indicators** | `TypingIndicator` becomes a real staggered dot wave; `StreamingDot` breathes via `PhaseAnimator` | **Fixes an a11y violation** — the old `repeatForever` hacks ignored Reduce Motion (now: calm static dots) and the typing 'stagger' never actually staggered |
| 6 | **Pressed-state micro-interactions** | Shared `GlassPressStyle` (springy dip + dim; Reduce Motion collapses to dim) on `CircularIconButton`, `PillSelector`, `FloatingActionMenu` rows, `DrawerRow`, `EmptyChatSuggestionRow`, `SuggestionPills`, jump-to-latest | `.plain` gave zero touch feedback on custom-backgrounded buttons; these shared controls power Chats home, the drawer, chat, and the empty states — every touch now answers |

## Verification

- `xcodebuild -scheme CovenCave -destination 'generic/platform=iOS Simulator' build` → **BUILD SUCCEEDED** (Xcode 26.6, arm64 + x86_64)
- `node scripts/run-tests.mjs mobile` → **78/78 test files pass** (includes the new pin)
- `node scripts/check-tests-wired.mjs` → all 851 test files wired
- New regression pin: `scripts/ios-modern-polish.test.mjs` locks all six vectors (wired into `test:mobile`)
- Existing pins untouched and green: `ios-chat-restyle`, `ios-reply-feedback`, `ios-swipe-reply`, `ios-chat-draft-lag`

Bead: **cave-v6eq**

### DIFF
--- a/apps/ios/CovenCave/CovenCave/Haptics.swift
+++ b/apps/ios/CovenCave/CovenCave/Haptics.swift
@@ -1,16 +1,36 @@
 import UIKit
 
 /// Lightweight haptic feedback for chat interactions (send, copy, retry).
+///
+/// Generators are cached and re-`prepare()`d after every fire so the Taptic
+/// Engine stays primed — Apple's recommended pattern. The old approach
+/// (allocating a fresh `UIImpactFeedbackGenerator` per tap) both added
+/// first-tap latency while the engine spun up and churned an allocation on
+/// every send/copy/toggle.
+@MainActor
 enum Haptics {
+    private static var impacts: [UIImpactFeedbackGenerator.FeedbackStyle: UIImpactFeedbackGenerator] = [:]
+    private static let notifier = UINotificationFeedbackGenerator()
+
     static func tap(_ style: UIImpactFeedbackGenerator.FeedbackStyle = .light) {
-        UIImpactFeedbackGenerator(style: style).impactOccurred()
+        let generator: UIImpactFeedbackGenerator
+        if let cached = impacts[style] {
+            generator = cached
+        } else {
+            generator = UIImpactFeedbackGenerator(style: style)
+            impacts[style] = generator
+        }
+        generator.impactOccurred()
+        generator.prepare()
     }
 
     static func success() {
-        UINotificationFeedbackGenerator().notificationOccurred(.success)
+        notifier.notificationOccurred(.success)
+        notifier.prepare()
     }
 
     static func error() {
-        UINotificationFeedbackGenerator().notificationOccurred(.error)
+        notifier.notificationOccurred(.error)
+        notifier.prepare()
     }
 }
--- a/apps/ios/CovenCave/CovenCave/Theme/ChatChrome.swift
+++ b/apps/ios/CovenCave/CovenCave/Theme/ChatChrome.swift
@@ -23,6 +23,30 @@ enum ChatChrome {
     static let iconWell: CGFloat = 34
 }
 
+// MARK: - GlassPressStyle
+
+/// The app's standard pressed state: a quick springy dip plus a slight dim, so
+/// every glass control answers the finger the way native Apple controls do
+/// (`.plain` gives no feedback at all on custom-backgrounded buttons). Under
+/// Reduce Motion the scale is dropped and only the dim remains.
+struct GlassPressStyle: ButtonStyle {
+    var scale: CGFloat = 0.96
+    @Environment(\.accessibilityReduceMotion) private var reduceMotion
+
+    func makeBody(configuration: Configuration) -> some View {
+        configuration.label
+            .scaleEffect(reduceMotion ? 1 : (configuration.isPressed ? scale : 1))
+            .opacity(configuration.isPressed ? 0.85 : 1)
+            .animation(reduceMotion ? nil : .spring(response: 0.28, dampingFraction: 0.75),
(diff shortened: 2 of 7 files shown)


BunsDev:
## Comprehensive review (retrospective — PR merged at 23:28Z)

Reviewed post-merge as part of the 3031/3034/3035/3036 review sweep. Verdict: **solid work — would have approved**. Verified the platform-API constraints below; no follow-ups required.

### What this does
Six native-Apple polish vectors across chat + shared chrome: cached/prepared haptics, a streaming-scroll contract, glassmorphic composer, accent-gradient bubbles with luminance-aware foreground, PhaseAnimator indicators, and a shared pressed-state style.

### Verified platform constraints
- `defaultScrollAnchor(_:for:)` (iOS 18) and `PhaseAnimator` (iOS 17) are both within the project's `IPHONEOS_DEPLOYMENT_TARGET = 18.0` — checked against `project.pbxproj`. ✅
- `Haptics` is `@MainActor` with cached static generators; call sites are SwiftUI view actions (main-thread), so no isolation friction. Cache-then-`prepare()` after each fire is Apple's documented Taptic pattern — this removes real first-tap latency and per-tap allocation at 40+ call sites. ✅

### Strengths
- **The scroll contract fixes the worst chat-UX sin.** Token streaming only auto-scrolls while `atBottom`; new messages reveal at bottom or for your own send; otherwise the unread waits behind jump-to-latest. Re-engagement is implicit via the existing `atBottom` tracking. `defaultScrollAnchor(.bottom, for: .initialOffset)` kills the post-layout jump on open (the `onAppear` scroll stays as a restore backstop).
- **Real a11y fixes, not just polish**: the old `repeatForever` indicator hacks ignored Reduce Motion (and the typing "stagger" never actually staggered — all three dots shared one phase). Both indicators now use `PhaseAnimator` with calm static fallbacks under Reduce Motion; `TypingIndicator` gains an accessibility label; every new animation in the diff is `reduceMotion`-guarded.
- **Real contrast fix**: user-bubble text was hard-coded `.white` on the accent — unreadable with light accents. `ChromePalette.accentForeground` (YIQ brightness, threshold 0.62, near-black ≤ white) mirrors the desktop's `--accent-presence-foreground` convention, falling back to white when the accent can't resolve — i.e. exactly the pre-theme behavior. Assistant bubbles track `chrome.bgRaised`, and the fallback palette resolves to the same `secondarySystemBackground` as before, so unthemed installs see no change.
- `Color.toned(lighter:)` degrades to the flat original when RGBA resolution fails rather than misrendering — good defensive posture for dynamic colors.
- `GlassPressStyle` centralizes the pressed state (springy dip + dim; Reduce Motion collapses to dim-only) and is applied across all seven shared controls — `.plain` previously gave zero touch feedback on custom-backgrounded buttons.
- **Regression pin** (`scripts/ios-modern-polish.test.mjs`) locks all six vectors including the *negative* assertions (no throwaway generator, no `repeatForever`, no unguarded animations) and is wired into `test:mobile`. Build verified on Xcode 26.6 for both simulator arches; 78/78 mobile test files pass.

### Minor notes (non-blocking)
- YIQ on gamma-encoded sRGB components is the standard web heuristic (matches how the desktop token is derived) — fine; a relative-luminance/WCAG computation would be marginally more accurate at the threshold boundary but isn't warranted here.
- The count-keyed `.animation(..., value: thread.messages.count)` also animates removals (fade) — intended per the comment, and Reduce Motion disables the whole thing.

PR #3048, opened 2026-07-13
fix(settings): verified OpenCoven tool updates — post-install verification

BunsDev:
Re-lands #3029 ("Fix OpenCoven tool update verification" by @romgenie) reconciled onto current main, addressing the CHANGES_REQUESTED integration blockers:

1. **#3026 preserved** — main's stricter Windows `.cmd`/`.bat` shim parser and explicit unresolved-shim signal in `coven-bin.ts` are untouched.
2. **#3028 preserved** — the atomic global npm lane and cancellation-safe finalizers stay; post-install verification layers on top.
3. **Machine-local paths** — the new `executablePath`/`packagePath` status fields are omitted from copied About diagnostics (the omission landed with #3030 and is regression-covered).

New: `src/lib/opencoven-tool-verification.ts` — package/entry-point/latest-version verification gating update success — wired into tools status, the settings update flow, and the onboarding install route.

Supersedes #3029.

## Validation
- `pnpm exec tsc --noEmit`
- Targeted `node --experimental-strip-types` runs on all touched test files
- `node scripts/check-tests-wired.mjs`

PR #3060, opened 2026-07-13
Fix Settings reattachment to running OpenCoven installs

CompleteDotTech:
## Summary

- Reattach Settings → About → OpenCoven tools to a Coven CLI or Coven Code install already owning the shared global npm lane.
- Poll attached jobs through completion with one race-safe observer and route terminal jobs through the existing authoritative status refresh, verification reconciliation, and banner revalidation.
- Refresh retained tool rows when the lane clears and restrict lane owners to a fixed client allowlist.
- Replace the misleading reattachment source assertion with behavioral coverage for reload, completion-before-attachment, failure, lane changes, duplicate completion prevention, and unmount cleanup.

## Root cause

The About panel polled the server-owned npm lane but only polled job completion for targets already present in its local `installJobs` state. Reloading while another client owned a Coven tool install displayed the shared-lane message without attaching to that job, so completion, verification, status refresh, and banner revalidation never occurred.

## User impact

Reloading Settings during an OpenCoven tool install now resumes truthful progress and completion reporting without starting another install or weakening the process-wide npm serialization contract.

## Validation

- `pnpm typecheck`
- `node --experimental-strip-types src/lib/opencoven-install-job-observer.test.ts`
- `node --experimental-strip-types src/components/open-coven-tools-update.test.ts`
- `node --experimental-strip-types src/lib/opencoven-tools-state.test.ts`
- `node --experimental-strip-types src/app/onboarding-install-route.test.ts`
- `node --experimental-strip-types src/app/api/onboarding/install/route.test.ts`
- `pnpm check:tests-wired` (889 test files wired)

`pnpm test:app` reached 82 passing files before the existing Windows `python3` app-execution alias failure in `scripts/sync-marketplace.test.mjs:188` (exit 9009).

## Integration notes

Rebased after #3031 merged and #3048 closed, preserving the six-state tool presentation and post-install verification model.


BunsDev:
Pushed 05225625 on top of your main-merge: your resolution guarded `document.hidden` inside the observer's default interval, but main's structural pin (`src/lib/pausable-poll-discipline.test.ts`, cave-e794) requires the always-mounted lane poll to literally go through `usePausablePoll(() => void refreshNpmLane(), 2000)` — the merged tree failed that assertion, which would have failed `Frontend build`. The fix disables the observer's self-scheduling in this surface (`schedule: () => null`) and drives `observer.pollNow()` from `usePausablePoll`, which also restores the instant on-return refresh for both browser and Tauri focus. Kept your hidden-guard on the observer's default interval as defense-in-depth for other callers. Validated locally: pausable-poll-discipline, open-coven-tools-update, opencoven-install-job-observer tests + tsc all pass. Will merge once checks are green.

PR #3061, opened 2026-07-13
Preserve safe tool verification states in About diagnostics

CompleteDotTech:
## Summary

- preserve canonical npm freshness and safe error states in copied About diagnostics
- include package/executable verification booleans and allowlisted discovery errors
- reject unknown future enum values, invalid timestamps, and malformed versions instead of copying arbitrary text
- update the diagnostics disclosure and cover the requested failure modes

## Why

The Settings → About diagnostics payload dropped structured verification facts that the UI already uses. Support therefore could not distinguish registry failures, unexpected packages, unreadable launchers, or failed version probes from the sanitized payload.

## Impact

Copied diagnostics remain backwards-readable and continue to omit local paths, install commands, raw process output, URL query/fragment values, and secret-like strings. No installation behavior or user-facing status copy changes.

## Validation

- `node --experimental-strip-types src/lib/about-diagnostics.test.ts`
- `node --experimental-strip-types src/components/open-coven-tools-update.test.ts`
- `node --experimental-strip-types src/app/api/opencoven-tools/status/route.test.ts`
- `pnpm typecheck`
- `pnpm check:tests-wired`
- `git diff --check`

The broader `pnpm test:app` run reached 80 passing files before an unrelated Windows host failure in `scripts/sync-marketplace.test.mjs`: `python3` resolved the Microsoft Store alias and exited 9009.

Bead: `cave-2l4`


BunsDev (approved):
Reviewed for code integrity + security: safeLatestCheck/safeDiscoveryError are strict allowlists that fail closed on unknown enum values, checkedAt must parse and is re-serialized via toISOString (no text smuggling through timestamps), SEMVER-gated latest, and booleans coerced with === true. The future-value injection test locks all of it in. Nice hardening.

PR #3114, opened 2026-07-14
fix(familiars): preserve server hub roster

iamcobolt:
## Summary
- preserve the Server hub roster by resolving its target from the same Cave config snapshot and applying local seed/tombstone policy only to a local daemon
- preserve explicitly authored reserved/default familiars, including the complete Sage/Forge/OpenCode trio
- honor `COVEN_HOME` for familiar roster reads and creation instead of hard-coding `~/.coven`

## Root cause
The roster guard introduced in #2994 compared every daemon response with the Cave machine’s local `familiars.toml` and tombstones. In Server hub mode those files belong to a different authority, so valid hub familiars with reserved/default ids were removed before the UI received them.

## Validation
- `pnpm typecheck`
- `pnpm check:tests-wired` (926 files wired)
- `pnpm test:api` (168 files)
- `pnpm test:app` (684 files)
- `pnpm build` (including bundle budget)
- focused roster guard, familiars route, and lifecycle tests

## Scope
This fixes the configured Server hub roster path. Independent SSH `remoteHosts` and executor URLs are execution targets rather than roster authorities today; aggregating those would require host-scoped familiar identity and routing and is intentionally not included here.


BunsDev:
Thanks for this fix! While it was open, main independently landed hub-aware roster shaping (#3118, cave-7cv4), which overlapped the core mechanism here. I merged main into this branch (d108d811) and reconciled:

- **Kept from main:** mode-aware roster shaping (hub bypasses the install-seed guard, live-state entries are never hidden, locally-declared familiars merge into the response so POST's 409 targets stay visible), and tombstone filtering in both modes so Remove stays effective from this Cave.
- **Kept from this PR:** `covenHome()` in GET+POST (COVEN_HOME honored — main still hard-coded `~/.coven`), the single config-snapshot daemon target (`callDaemonTarget(target)` + `target.mode` for the hub check), removal of the all-defaults early-return in `filterInstallSeedFamiliars` so an explicitly authored Sage/Forge/OpenCode trio survives (your test pins this), and your test coverage.
- **Dropped:** `filterFamiliarRosterForAuthority` — the route now expresses authority awareness through main's landed shaping, so the helper would have been dead code.

Verified: `pnpm typecheck`, `pnpm test:api` (168), `pnpm test:app` (693). Will merge once required checks are green.

PR #3243, opened 2026-07-15
fix(dashboard): needs-attention triage is live, single-headed, and honest when cleared (cave-ckxk)

BunsDev:
## Problem (dashboard-command-center goal — needs-attention/caught-up UX pass)

- **Frozen model**: `buildDashboardModel` ran once server-side and was never refreshed — the cockpit's 30s poll updated every panel **except** Needs attention, the Needs-you vital, caught-up state, and Today's summary. The cave-bzch 'adopt fresh `initialItems`' effect in ActionInbox could never fire because its prop never changed.
- **Double header**: `Panel "Needs attention" [N]` wrapped ActionInbox's own `dr-section` head ("Needs you N" + Select) — two names, two counts, one list.
- **No caught-up moment**: clearing the last row returned `null`, leaving a hollow panel with a stale count until reload. The sister daily-report surface has an explicit caught-up state; the cockpit didn't.
- **Dishonest count, dead-end tile**: the Needs-you vital showed the capped-at-8 `needsAttention.length`, not the true open total, and was the only KPI tile without a drill-in.

## Fix

- Cockpit rebuilds the model client-side from `/api/inbox` on the existing 30s poll (`liveModel ?? initialModel`; a failed fetch keeps the last good model instead of wiping to caught-up). The one inbox fetch also feeds the agenda's pending items.
- `DashboardModel.openCount` (uncapped) feeds the vital + panel badge; both track optimistic done/dismiss/snooze between polls via the live count ActionInbox reports up.
- ActionInbox: internal header → slim select bar (Panel chrome owns title/count); emptied list renders a **caught-up confirmation**; open items beyond the visible cap render **"+N more — open Rituals"** (`/?mode=inbox`); Needs-you vital + panel head deep-link the same surface.

## Verification

- `pnpm test:app` — 730 files pass (model + page-contract pins updated/added)
- `pnpm typecheck` — clean
- Playwright `tests/dashboard-cockpit.spec.ts` — 8/8 pass locally, incl. 2 new tests: live triage + caught-up confirmation; uncapped badge/vital + overflow drill-through (the mocked `/api/inbox` now fully determines the panel — it couldn't before this fix)

Bead: cave-ckxk

### DIFF
--- a/src/app/dashboard-page.test.ts
+++ b/src/app/dashboard-page.test.ts
@@ -200,6 +200,37 @@ assert.match(cockpit, /aria-label=\{`Drag to rearrange: \$\{title\}`\}/, "each g
   assert.match(inbox, /actedIdsRef\.current\.add\(item\.id\)/, "single actions register the acted id");
   assert.match(inbox, /ids\.forEach\(\(id\) => actedIdsRef\.current\.add\(id\)\)/, "bulk actions register acted ids");
 }
+
+// ── Needs-attention is live, single-headed, and honest when cleared (cave-ckxk) ──
+// The server-rendered model used to be the only model — the poll refreshed
+// every panel EXCEPT needs-attention/Needs-you/today-summary. The cockpit now
+// rebuilds the model from the live inbox each poll, so the cave-bzch adoption
+// above actually has fresh props to adopt.
+{
+  const inbox = readFileSync(new URL("../components/dashboard/action-inbox.tsx", import.meta.url), "utf8");
+  const root = readFileSync(cockpitUrl, "utf8");
+  assert.match(root, /getJson<\{ items: InboxItem\[\] \}>\("\/api\/inbox"\)/, "the poll fetches the full live inbox");
+  assert.match(root, /setLiveModel\(buildDashboardModel\(r\.items, new Date\(\)\)\)/, "each poll rebuilds the dashboard model client-side");
+  assert.match(root, /const model = liveModel \?\? initialModel;/, "the server model is only the first frame");
+  assert.match(root, /if \(!r\?\.items \|\| !aliveRef\.current\) return;/, "a failed inbox fetch keeps the last good model instead of wiping to caught-up");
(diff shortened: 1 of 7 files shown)


BunsDev:
**Session race notice** (copilot-cli 9803c5c6, same goal `dashboard-command-center`): PR **#3241** (cave-456r) merged to main at `ceaf0fe1` ~20 min before this PR was opened, covering the same core: live client-rebuilt model per poll (full `/api/inbox`, last-known-good on failed fetch, pending filter preserved for agenda), de-duplicated chrome, and a designed caught-up state. This branch is now unmergeable against main.

**Not superseded — this PR carries three genuine deltas #3241 lacks:**
1. `DashboardModel.openCount` (uncapped) — the merged KPI still reads capped `needsAttention.length` (lies at 9+ open items).
2. Live count reported up (`onOpenCount`) so the **KPI tile** tracks optimistic done/dismiss/snooze between polls — in the merged design only the section's own count is live-local.
3. `+N more — open Rituals` overflow drill-through + the needs KPI/panel href into `/?mode=inbox`.

**Chrome direction differs**: #3241 kept ActionInbox's dr-section as the single owner (bare mount, today-summary precedent, panel persists when caught up); this PR kept the outer Panel and slimmed ActionInbox. Merged reality is the former — suggest re-scoping this branch onto main to just the three deltas above (cave-ckxk claim is yours; notes updated there too). Happy to leave the queue to your session — nothing here is claimed by mine.

BunsDev:
Superseded by #3241 (cave-456r), which landed the same core fix — live client-side model rebuild from /api/inbox with last-known-good on failure, single chrome (bare ActionInbox mount), designed caught-up state — from a parallel session ~30 min earlier. Closing rather than rebasing: the remaining genuine deltas (uncapped DashboardModel.openCount for the Needs-you vital + trends, overflow '+N more — open Rituals' drill-through past the visible cap, KPI deep-link to /?mode=inbox, clear-last-item e2e) are small and integrate into #3241's bare-widget architecture, so they'll re-land as a focused follow-up instead.

PR #3270, opened 2026-07-16
Fix lossless Cave home migration on Windows

### LINKED ISSUE #3269: Make Cave home migration transactional and data-safe on Windows
## Summary

Cave can leave real legacy `~/.coven/cave-*.json` files beside canonical `~/.coven/cave/*.json` files. The current migration classifies the pair as an unmergeable conflict and tells the user that the canonical copy wins, even when the legacy copy contains user data absent from the destination.

Beads source of truth: `cave-l30`.

## Reproduced environment

On the affected Windows environment:

- Developer Mode is disabled.
- Cave is running non-elevated.
- Creating a file symlink fails with `Administrator privilege required for this operation`.
- Directory junctions work, so `cave-conversations` and `cave-preferences.json.locks` have compatibility junctions.
- The live `GET /api/cave-home-migration` response reports:
  - `cave-state.json`
  - `cave-inbox.json`
  - `cave-preferences.json`
- Each legacy file and canonical file is a different real file with a different hash and modification history.
- The legacy state contains session mappings absent from the canonical state, demonstrating that unconditional destination-wins cleanup is not data-safe.

## Current causes

- Startup migration is process-local and effectively one-shot.
- File symlink creation is best-effort and its failure is silently ignored.
- A stale or older writer can recreate a real legacy file when no compatibility link exists.
- Once both paths exist, the migrator refuses to reconcile them and always preserves the destination as authoritative.
- The banner exposes only a count and dismissal; it does not identify files or provide recovery/merge controls.

Relevant implementation:

- `instrumentation.ts`
- `src/lib/server/cave-home-migration.ts`
- `src/lib/server/cave-home-migration-status.ts`
- `src/app/api/cave-home-migration/route.ts`
- `src/components/cave-home-migration-banner.tsx`

## Proposed production design

1. Keep `~/.coven/cave/` as canonical storage.
2. Acquire a cross-process migration lock before Cave state readers or writers start.
3. Persist an atomic migration journal containing schema version, hashes, timestamps, compatibility-mirror state, and unresolved decisions.
4. Back up both copies with a hash manifest before any mutation.
5. Add schema-aware reconciliation:
   - inbox items by stable ID and revision/timestamp;
   - session-state maps by key;
   - preferences by revision and independent section;
   - explicit user choice for irreconcilable data.
6. Treat symlinks as an optional optimization, never a correctness requirement.
7. On systems without file-symlink privileges, use journaled compatibility mirrors and detect later legacy-side changes by hash.
8. Replace the count-only warning with a review surface naming each file and offering safe merge, keep current, recover legacy, and open-backup actions.
9. Retry reconciliation until the durable journal confirms completion.

## Acceptance criteria

- [ ] Every conflicting input is preserved or backed up before modification.
- [ ] Migration works on Windows with Developer Mode disabled and without elevation.
- [ ] State, inbox, and preferences conflicts use tested schema-aware reconciliation or require explicit user choice.
- [ ] Interrupted and concurrent migration attempts recover idempotently.
- [ ] Legacy files modified after migration are detected without repeatedly warning about unchanged managed mirrors.
- [ ] The UI identifies affected files and offers **Review**, **Merge safely**, **Keep current**, **Recover legacy**, and **Open backup folder**.
- [ ] Tests cover symlink `EPERM`, stale writers, divergent copies, crash interruption, concurrency, and backup restoration.
- [ ] Existing `COVEN_HOME`, `COVEN_CAVE_HOME`, and `COVEN_PREFERENCES_PATH` overrides remain backward-compatible.

CompleteDotTech:
## What changed

- replace destination-wins startup migration with a cross-process locked, atomic reconciliation journal
- create verified, hash-manifested recovery bundles before changing divergent data and retain unresolved recovery sources
- merge inbox, session state, preferences, and directory data with explicit per-schema strategies; leave ambiguous data unresolved
- support non-elevated Windows with hash-tracked ordinary compatibility mirrors when file symlinks are unavailable
- gate the primary config/state, inbox, and preferences readers and writers on reconciliation
- add a file-by-file review dialog with timestamps, bounded diff summaries, merge/keep/recover/defer actions, and backup-folder access
- preserve `COVEN_HOME`, `COVEN_CAVE_HOME`, `COVEN_PREFERENCES_PATH`, `COVEN_THEME_PATH`, and `COVEN_BACKDROP_PATH` behavior

## Root cause

The old migration moved only when the canonical destination was absent. If both copies existed it always kept the canonical copy, even when the legacy copy contained unique user data. It also silently ignored file-symlink failures. On non-elevated Windows with Developer Mode disabled, those symlinks require administrator privilege, leaving ordinary legacy files that later appeared as permanent conflicts.

## User impact

Upgrades are now lossless and resumable. Cave validates and journals decisions, preserves both divergent inputs in a verified recovery bundle, detects writes made through old-tool compatibility mirrors, and asks the user to resolve only cases that cannot be merged safely.

## Validation

- `pnpm typecheck`
- `pnpm check:tests-wired` (1010 test files wired)
- `pnpm test:app` (746 files)
- `pnpm test:api` (189 files)
- `pnpm test:mobile` (80 files)
- `pnpm test:conformance` (Linux checks passed; Windows-only migration case wired for `windows-latest`)
- `pnpm test:supply-chain`
- `pnpm build` (passed, including bundle budgets)
- `git diff --check`

The repository has no lint script or ESLint/Biome dependency, so there is no separate local lint command.

Fixes #3269

Bead: `cave-l30`

### DIFF
--- a/src/lib/server/cave-home-reconciliation.ts
+++ b/src/lib/server/cave-home-reconciliation.ts
@@ -623,14 +623,19 @@ async function mergeJson(strategy: ReconciliationStrategy, legacyPath: string, c
   return { ok: false, summary: "This file type requires an explicit choice." };
 }
 
-async function validateCanonical(entry: CaveHomeReconciliationEntry, canonicalPath: string): Promise<void> {
-  const info = await pathInfo(canonicalPath);
+async function validateCanonical(
+  entry: CaveHomeReconciliationEntry,
+  canonicalPath: string,
+  knownInfo?: PathInfo,
+): Promise<void> {
+  const info = knownInfo ?? await pathInfo(canonicalPath);
   if (info.kind === "missing" || info.kind === "symlink") throw new Error("canonical path is missing");
   if (entry.strategy === "directory") {
     if (info.kind !== "dir") throw new Error("canonical directory is invalid");
     await readdir(canonicalPath);
     return;
   }
+  if (info.kind !== "file") throw new Error("canonical file is invalid");
   if (entry.strategy === "inbox" || entry.strategy === "state" || entry.strategy === "preferences" || entry.next.endsWith(".json")) {
     const parsed = JSON.parse(await readFile(canonicalPath, "utf8")) as unknown;
     if (entry.strategy === "inbox") {
@@ -685,6 +690,9 @@ async function syncManagedMirror(
   const canonicalInfo = await pathInfo(canonicalPath);
   if (legacyInfo.kind === "missing" || canonicalInfo.kind === "missing") return null;
   if (!prior.managedMirrorHash || legacyInfo.hash !== prior.managedMirrorHash) return null;
+  // Do not replace the last known-good mirror until the newer canonical copy
+  // passes the same schema/type checks used by the initial migration.
+  await validateCanonical(entry, canonicalPath, canonicalInfo);
   const current: MigrationJournalEntry = {
     ...prior,
     legacyHash: legacyInfo.hash,
@@ -924,7 +932,12 @@ export async function caveHomeReconciliationStatus(
     if (legacyInfo.kind === "missing" || legacyInfo.kind === "symlink") continue;
     const canonicalInfo = await pathInfo(canonicalPath);
     const prior = journal.entries[entry.legacy];
-    const managed = Boolean(prior?.managedMirrorHash && legacyInfo.hash === prior.managedMirrorHash);
+    const canonicalValid = await validateCanonical(entry, canonicalPath, canonicalInfo).then(() => true, () => false);
+    const managed = Boolean(
+      canonicalValid &&
+      prior?.managedMirrorHash &&
+      legacyInfo.hash === prior.managedMirrorHash,
+    );
     if (managed) continue;
     const isPending = canonicalInfo.kind === "missing";
     (isPending ? pending : conflicts).push(entry.legacy);
--- a/src/lib/server/cave-home-reconciliation.ts
+++ b/src/lib/server/cave-home-reconciliation.ts
@@ -1217,7 +1217,16 @@ async function reconcileEntry(
   }
 
   let merged: MergeOutcome;
-  if (entry.strategy === "directory") merged = (await reconcileDirectory(entry, legacyPath, canonicalPath, legacyInfo, canonicalInfo, result, options)).outcome;
+  if (entry.strategy === "directory") {
+    // Files in migrated directories are not append-only (conversations, for
+    // example, can be deleted). A child present only in the legacy snapshot is
+    // therefore ambiguous: it may be a new file or a stale file deleted from
+    // canonical storage. Do not resurrect it until the user explicitly asks
+    // to merge the two directory snapshots.
+    merged = options.action === "merge"
+      ? (await reconcileDirectory(entry, legacyPath, canonicalPath, legacyInfo, canonicalInfo, result, options)).outcome
+      : { ok: false, summary: "Directory entries differ and require an explicit merge or whole-directory choice." };
+  }
   else if (["inbox", "state", "preferences"].includes(entry.strategy)) {
     // Merge the exact snapshots recorded in the verified bundle. Reading the
     // live paths here would let an uncoordinated store writer change either
--- a/src/lib/server/cave-home-reconciliation.ts
+++ b/src/lib/server/cave-home-reconciliation.ts
@@ -301,35 +301,37 @@ async function acquireLock(options: ReconciliationOptions): Promise<() => Promis
       if (!["EEXIST", "ENOTEMPTY"].includes((error as NodeJS.ErrnoException).code ?? "")) throw error;
       try {
         const info = await stat(lock);
-        if (Date.now() - info.mtimeMs > LOCK_STALE_MS) {
-          const owner = await readLockOwner(lock);
-          let alive = false;
-          if (owner.pid) {
-            try {
-              process.kill(owner.pid, 0);
-              alive = true;
-            } catch (ownerError) {
-              alive = (ownerError as NodeJS.ErrnoException).code === "EPERM";
-            }
+        const owner = await readLockOwner(lock);
+        let alive = false;
+        if (owner.pid) {
+          try {
+            process.kill(owner.pid, 0);
+            alive = true;
+          } catch (ownerError) {
+            alive = (ownerError as NodeJS.ErrnoException).code === "EPERM";
           }
-          if (!alive) {
-            await options.lockProbe?.("stale-observed");
-            const takeover = path.join(lock, ".takeover");
-            const takeoverToken = randomBytes(16).toString("hex");
-            try {
-              await writeFile(takeover, JSON.stringify({ takeoverToken, ownerToken: owner.token }), { flag: "wx" });
-              const currentOwner = await readLockOwner(lock);
-              if (currentOwner.token !== owner.token) {
-                await rm(takeover, { force: true });
-              } else {
-                const reclaimed = `${lock}.reclaimed-${takeoverToken}`;
-                await rename(lock, reclaimed);
-                await rm(reclaimed, { recursive: true, force: true });
-                continue;
-              }
-            } catch (takeoverError) {
-              if (!["EEXIST", "ENOENT"].includes((takeoverError as NodeJS.ErrnoException).code ?? "")) throw takeoverError;
+        }
+        // A recorded owner that no longer exists is safe to reclaim
+        // immediately. The age threshold is only needed for an unreadable or
+        // incomplete owner record, where liveness cannot be established.
+        const reclaimable = owner.pid ? !alive : Date.now() - info.mtimeMs > LOCK_STALE_MS;
+        if (reclaimable) {
+          await options.lockProbe?.("stale-observed");
+          const takeover = path.join(lock, ".takeover");
+          const takeoverToken = randomBytes(16).toString("hex");
+          try {
+            await writeFile(takeover, JSON.stringify({ takeoverToken, ownerToken: owner.token }), { flag: "wx" });
+            const currentOwner = await readLockOwner(lock);
+            if (currentOwner.token !== owner.token) {
+              await rm(takeover, { force: true });
+            } else {
+              const reclaimed = `${lock}.reclaimed-${takeoverToken}`;
+              await rename(lock, reclaimed);
+              await rm(reclaimed, { recursive: true, force: true });
+              continue;
             }
+          } catch (takeoverError) {
+            if (!["EEXIST", "ENOENT"].includes((takeoverError as NodeJS.ErrnoException).code ?? "")) throw takeoverError;
           }
         }
       } catch (lockError) {
--- a/src/lib/server/cave-home-reconciliation.ts
+++ b/src/lib/server/cave-home-reconciliation.ts
@@ -1,0 +1,961 @@
+import { createHash, randomBytes } from "node:crypto";
+import {
+  copyFile,
+  cp,
+  lstat,
+  mkdir,
+  readFile,
+  readlink,
+  readdir,
+  rename,
+  rm,
+  stat,
+  symlink,
+} from "node:fs/promises";
+import path from "node:path";
+
+import { caveHome, covenHome } from "../coven-paths.ts";
+import { normalizeCavePreferences, type CavePreferences } from "../preferences-schema.ts";
+import { writeJsonAtomic } from "./atomic-write.ts";
+
+export type ReconciliationStrategy = "inbox" | "state" | "preferences" | "directory" | "manual";
+
+export type CaveHomeReconciliationEntry = {
+  legacy: string;
+  next: string;
+  strategy: ReconciliationStrategy;
+};
+
+export type ReconciliationDecision =
+  | "absent"
+  | "moved"
+  | "linked"
+  | "managed-mirror"
+  | "identical"
+  | "merged"
+  | "kept-canonical"
+  | "recovered-legacy"
+  | "unresolved"
+  | "deferred";
+
+export type MigrationJournalEntry = {
+  legacy: string;
+  next: string;
+  strategy: ReconciliationStrategy;
+  legacyPath: string;
+  canonicalPath: string;
+  legacyHash?: string;
+  canonicalHash?: string;
+  legacyMtimeMs?: number;
+  canonicalMtimeMs?: number;
+  decision: ReconciliationDecision;
+  compatibility?: "symlink" | "mirror";
+  managedMirrorHash?: string;
+  backupId?: string;
+  decidedAt: string;
+  summary?: string;
+};
+
+export type MigrationJournal = {
+  version: 1;
+  migrationVersion: 2;
+  updatedAt: string;
+  entries: Record<string, MigrationJournalEntry>;
+};
+
+export type CaveHomeConflictDetail = {
+  legacy: string;
+  next: string;
+  strategy: ReconciliationStrategy;
+  legacyPath: string;
+  canonicalPath: string;
+  legacyHash?: string;
+  canonicalHash?: string;
+  legacyMtimeMs?: number;
+  canonicalMtimeMs?: number;
+  state: "pending" | "unresolved" | "managed";
+  summary: string;
+  differences: string[];
+  backupPath?: string;
+  actions: Array<"merge" | "keep-canonical" | "recover-legacy" | "defer">;
+};
+
+export type CaveHomeReconciliationStatus = {
+  pending: string[];
+  conflicts: string[];
+  migrated: boolean;
+  details: CaveHomeConflictDetail[];
+  backupRoot: string;
+  journalPath: string;
+};
+
+export type CaveHomeReconciliationResult = {
+  moved: string[];
+  linked: string[];
+  skipped: string[];
+  merged: Array<{ legacy: string; files: number; collisions: number }>;
+  backedUp: string[];
+  resolved: string[];
+  errors: Array<{ legacy: string; error: string }>;
+};
+
+export type ReconciliationAction = "merge" | "keep-canonical" | "recover-legacy" | "defer";
+
+export type ReconciliationOptions = {
+  action?: ReconciliationAction;
+  legacy?: string;
+  /** Test-only fault boundary. Production callers must omit it. */
+  faultAt?: string;
+  /** Test-only compatibility bridge override. */
+  createSymlink?: typeof symlink;
+};
+
+type PathInfo = {
+  kind: "missing" | "symlink" | "file" | "dir";
+  hash?: string;
+  mtimeMs?: number;
+  size?: number;
+};
+
+type MergeOutcome =
+  | { ok: true; value: unknown; summary: string }
+  | { ok: false; summary: string };
+
+const JOURNAL_VERSION = 1 as const;
+const MIGRATION_VERSION = 2 as const;
+const BACKUP_RETENTION = 10;
+const LOCK_STALE_MS = 5 * 60_000;
+const LOCK_WAIT_MS = 10_000;
+
+export const migrationJournalPath = () => path.join(caveHome(), "migration-state.json");
+export const migrationBackupRoot = () => path.join(caveHome(), "migration-backups");
+const migrationLockPath = () => path.join(caveHome(), ".migration.lock");
+
+function canonicalPathFor(entry: CaveHomeReconciliationEntry): string {
+  const override = entry.next === "preferences.json"
+    ? process.env.COVEN_PREFERENCES_PATH?.trim()
+    : entry.next === "theme.json"
+      ? process.env.COVEN_THEME_PATH?.trim()
+      : entry.next === "backdrop.jpg"
+        ? process.env.COVEN_BACKDROP_PATH?.trim()
+        : undefined;
+  return override || path.join(caveHome(), entry.next);
+}
+
+function nowIso(): string {
+  return new Date().toISOString();
+}
+
+function fault(options: ReconciliationOptions, boundary: string): void {
+  if (options.faultAt === boundary) throw new Error(`injected migration fault at ${boundary}`);
+}
+
+async function sha256File(target: string): Promise<string> {
+  return createHash("sha256").update(await readFile(target)).digest("hex");
+}
+
+async function sha256Dir(target: string): Promise<string> {
+  const hash = createHash("sha256");
+  for (const name of (await readdir(target)).sort()) {
+    if (name === ".DS_Store") continue;
+    const child = path.join(target, name);
+    const info = await lstat(child);
+    hash.update(name);
+    if (info.isSymbolicLink()) hash.update(`l:${await readlink(child)}`);
+    else hash.update(info.isDirectory() ? `d:${await sha256Dir(child)}` : `f:${await sha256File(child)}`);
+  }
+  return hash.digest("hex");
+}
+
+async function pathInfo(target: string): Promise<PathInfo> {
+  try {
+    const info = await lstat(target);
+    if (info.isSymbolicLink()) return { kind: "symlink", mtimeMs: info.mtimeMs, size: info.size };
+    if (info.isDirectory()) return { kind: "dir", hash: await sha256Dir(target), mtimeMs: info.mtimeMs, size: info.size };
+    return { kind: "file", hash: await sha256File(target), mtimeMs: info.mtimeMs, size: info.size };
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code === "ENOENT") return { kind: "missing" };
+    throw error;
+  }
+}
+
+async function readJournal(): Promise<MigrationJournal> {
+  try {
+    const parsed = JSON.parse(await readFile(migrationJournalPath(), "utf8")) as Partial<MigrationJournal>;
+    if (parsed.version !== JOURNAL_VERSION || parsed.migrationVersion !== MIGRATION_VERSION || !parsed.entries) {
+      throw new Error("unsupported migration journal");
+    }
+    return parsed as MigrationJournal;
+  } catch (error) {
+    if ((error as NodeJS.ErrnoException).code === "ENOENT") {
+      return { version: JOURNAL_VERSION, migrationVersion: MIGRATION_VERSION, updatedAt: "", entries: {} };
+    }
+    throw error;
+  }
+}
+
+async function commitJournal(journal: MigrationJournal, options: ReconciliationOptions): Promise<void> {
+  fault(options, "before-journal-write");
+  journal.updatedAt = nowIso();
+  await writeJsonAtomic(migrationJournalPath(), journal);
+  fault(options, "after-journal-write");
+  const reread = JSON.parse(await readFile(migrationJournalPath(), "utf8")) as MigrationJournal;
+  if (JSON.stringify(reread) !== JSON.stringify(journal)) {
+    throw new Error("migration journal verification failed");
+  }
+}
+
+async function delay(ms: number): Promise<void> {
+  await new Promise<void>((resolve) => setTimeout(resolve, ms));
+}
+
+async function acquireLock(): Promise<() => Promise<void>> {
+  const lock = migrationLockPath();
+  const deadline = Date.now() + LOCK_WAIT_MS;
+  for (;;) {
+    try {
+      await mkdir(lock);
+      try {
+        await writeJsonAtomic(path.join(lock, "owner.json"), { pid: process.pid, startedAt: nowIso() });
+      } catch (error) {
+        await rm(lock, { recursive: true, force: true });
+        throw error;
+      }
+      return async () => { await rm(lock, { recursive: true, force: true }); };
+    } catch (error) {
+      if ((error as NodeJS.ErrnoException).code !== "EEXIST") throw error;
+      try {
+        const info = await stat(lock);
+        if (Date.now() - info.mtimeMs > LOCK_STALE_MS) {
+          const owner = await readFile(path.join(lock, "owner.json"), "utf8")
+            .then((value) => JSON.parse(value) as { pid?: unknown })
+            .catch(() => null);
+          const pid = typeof owner?.pid === "number" && Number.isSafeInteger(owner.pid) ? owner.pid : null;
+          let alive = false;
+          if (pid) {
+            try {
+              process.kill(pid, 0);
+              alive = true;
+            } catch (ownerError) {
+              alive = (ownerError as NodeJS.ErrnoException).code === "EPERM";
+            }
+          }
+          if (!alive) {
+            await rm(lock, { recursive: true, force: true });
+            continue;
+          }
+        }
+      } catch {
+        continue;
+      }
+      if (Date.now() >= deadline) throw new Error("timed out waiting for cave home migration lock");
+      await delay(50);
+    }
+  }
+}
+
+async function pruneBackups(): Promise<void> {
+  await mkdir(migrationBackupRoot(), { recursive: true });
+  const journal = await readJournal();
+  const protectedIds = new Set(
+    Object.values(journal.entries)
+      .filter((entry) => entry.decision === "unresolved" || entry.decision === "deferred")
+      .map((entry) => entry.backupId)
+      .filter((id): id is string => Boolean(id)),
+  );
+  const directories: Array<{ name: string; mtimeMs: number }> = [];
+  for (const name of await readdir(migrationBackupRoot())) {
+    const target = path.join(migrationBackupRoot(), name);
+    const info = await lstat(target).catch(() => null);
+    if (info?.isDirectory()) directories.push({ name, mtimeMs: info.mtimeMs });
+  }
+  directories.sort((a, b) => b.mtimeMs - a.mtimeMs);
+  // Leave room for the bundle about to be created. Never prune a bundle that
+  // is the only recovery source for an unresolved/deferred decision.
+  const removable = directories.filter((candidate) => !protectedIds.has(candidate.name));
+  const removeCount = Math.max(0, directories.length - (BACKUP_RETENTION - 1));
+  for (const stale of removeCount > 0 ? removable.slice(-removeCount) : []) {
+    await rm(path.join(migrationBackupRoot(), stale.name), { recursive: true, force: true });
+  }
+}
+
+async function copyPath(source: string, destination: string, kind: PathInfo["kind"]): Promise<void> {
+  if (kind === "dir") await cp(source, destination, { recursive: true, errorOnExist: true });
+  else await copyFile(source, destination);
+}
+
+async function copyPathAtomically(
+  source: string,
+  destination: string,
+  kind: PathInfo["kind"],
+  expectedHash?: string,
+): Promise<void> {
+  const temporary = path.join(
+    path.dirname(destination),
+    `.${path.basename(destination)}.migration-${process.pid}-${randomBytes(6).toString("hex")}`,
+  );
+  try {
+    await copyPath(source, temporary, kind);
+    const copied = await pathInfo(temporary);
+    if (!expectedHash || copied.hash !== expectedHash) throw new Error("migration copy verification failed");
+    await rename(temporary, destination);
+  } catch (error) {
+    await rm(temporary, { recursive: true, force: true }).catch(() => {});
+    throw error;
+  }
+}
+
+async function createRecoveryBundle(
+  entry: CaveHomeReconciliationEntry,
+  legacyPath: string,
+  canonicalPath: string,
+  legacyInfo: PathInfo,
+  canonicalInfo: PathInfo,
+  options: ReconciliationOptions,
+): Promise<{ id: string; directory: string }> {
+  await pruneBackups();
+  const id = `${nowIso().replace(/[:.]/g, "-")}-${randomBytes(4).toString("hex")}`;
+  const directory = path.join(migrationBackupRoot(), id);
+  await mkdir(directory, { recursive: true });
+  fault(options, "after-backup-directory");
+
+  const files: Array<{ role: string; sourcePath: string; storedPath: string; hash: string; size?: number }> = [];
+  for (const [role, source, info] of [
+    ["legacy", legacyPath, legacyInfo],
+    ["canonical", canonicalPath, canonicalInfo],
+  ] as const) {
+    if (info.kind === "missing" || info.kind === "symlink") continue;
+    const storedName = `${role}-${path.basename(source)}`;
+    const storedPath = path.join(directory, storedName);
+    await copyPath(source, storedPath, info.kind);
+    const storedInfo = await pathInfo(storedPath);
+    if (!info.hash || storedInfo.hash !== info.hash) throw new Error(`backup verification failed for ${role}`);
+    files.push({ role, sourcePath: source, storedPath: storedName, hash: info.hash, size: info.size });
+    fault(options, `after-backup-${role}`);
+  }
+  const manifest = {
+    version: 1,
+    createdAt: nowIso(),
+    entry: { legacy: entry.legacy, next: entry.next, strategy: entry.strategy },
+    files,
+  };
+  await writeJsonAtomic(path.join(directory, "manifest.json"), manifest);
+  const verified = JSON.parse(await readFile(path.join(directory, "manifest.json"), "utf8")) as typeof manifest;
+  if (JSON.stringify(verified) !== JSON.stringify(manifest)) throw new Error("backup manifest verification failed");
+  fault(options, "after-backup-manifest");
+  return { id, directory };
+}
+
+async function restoreBundleRole(directory: string, role: "legacy" | "canonical", destination: string): Promise<void> {
+  const manifest = JSON.parse(await readFile(path.join(directory, "manifest.json"), "utf8")) as {
+    files?: Array<{ role?: string; storedPath?: string; hash?: string }>;
+  };
+  const file = manifest.files?.find((candidate) => candidate.role === role);
+  if (!file?.storedPath || !file.hash) throw new Error(`verified ${role} backup is unavailable`);
+  const source = path.resolve(directory, file.storedPath);
+  if (path.dirname(source) !== path.resolve(directory)) throw new Error(`${role} backup path is invalid`);
+  const sourceInfo = await pathInfo(source);
+  if (sourceInfo.hash !== file.hash) throw new Error(`${role} backup hash changed before recovery`);
+  const destinationInfo = await pathInfo(destination);
+  await rm(destination, { recursive: destinationInfo.kind === "dir", force: true });
+  await copyPath(source, destination, sourceInfo.kind);
+  if ((await pathInfo(destination)).hash !== file.hash) throw new Error(`${role} recovery verification failed`);
+}
+
+function record(value: unknown): Record<string, unknown> | null {
+  return value && typeof value === "object" && !Array.isArray(value) ? value as Record<string, unknown> : null;
+}
+
+function parseDate(value: unknown): number {
+  return typeof value === "string" && Number.isFinite(Date.parse(value)) ? Date.parse(value) : 0;
+}
+
+async function describeDifferences(
+  entry: CaveHomeReconciliationEntry,
+  legacyPath: string,
+  canonicalPath: string,
+  legacyInfo: PathInfo,
+  canonicalInfo: PathInfo,
+): Promise<string[]> {
+  if (canonicalInfo.kind === "missing") return ["Canonical copy is missing; the legacy copy is the only source."];
+  if (legacyInfo.kind !== canonicalInfo.kind) return [`Path types differ: ${legacyInfo.kind} and ${canonicalInfo.kind}.`];
+  if (legacyInfo.hash === canonicalInfo.hash) return ["Contents are identical."];
+  if (entry.strategy === "directory" && legacyInfo.kind === "dir") {
+    const legacyNames = new Set(await readdir(legacyPath));
+    const canonicalNames = new Set(await readdir(canonicalPath));
+    const legacyOnly = [...legacyNames].filter((name) => !canonicalNames.has(name)).length;
+    const canonicalOnly = [...canonicalNames].filter((name) => !legacyNames.has(name)).length;
+    const shared = [...legacyNames].filter((name) => canonicalNames.has(name)).length;
+    return [`Directory entries: ${legacyOnly} legacy-only, ${canonicalOnly} canonical-only, ${shared} shared.`];
+  }
+  if (entry.next.endsWith(".json")) {
+    try {
+      const legacy = record(JSON.parse(await readFile(legacyPath, "utf8")));
+      const canonical = record(JSON.parse(await readFile(canonicalPath, "utf8")));
+      if (!legacy || !canonical) return ["JSON shapes differ."];
+      const keys = [...new Set([...Object.keys(legacy), ...Object.keys(canonical)])]
+        .filter((key) => JSON.stringify(legacy[key]) !== JSON.stringify(canonical[key]));
+      return keys.length > 0
+        ? [`Changed top-level fields: ${keys.slice(0, 8).join(", ")}${keys.length > 8 ? `, and ${keys.length - 8} more` : ""}.`]
+        : ["JSON formatting or ordering differs; parsed values match."];
+    } catch {
+      return ["One or both copies are not valid JSON."];
+    }
+  }
+  return [`Binary contents differ (${legacyInfo.size ?? 0} bytes legacy, ${canonicalInfo.size ?? 0} bytes canonical).`];
+}
+
+function mergeInbox(legacy: unknown, canonical: unknown): MergeOutcome {
+  const left = record(legacy);
+  const right = record(canonical);
+  if (!left || !right || !Array.isArray(left.items) || !Array.isArray(right.items)) {
+    return { ok: false, summary: "Inbox data is malformed and requires review." };
+  }
+  const byId = new Map<string, Record<string, unknown>>();
+  for (const raw of [...left.items, ...right.items]) {
+    const item = record(raw);
+    if (!item || typeof item.id !== "string" || !item.id) return { ok: false, summary: "Inbox item without a stable ID requires review." };
+    const existing = byId.get(item.id);
+    if (!existing) {
+      byId.set(item.id, item);
+      continue;
+    }
+    if (JSON.stringify(existing) === JSON.stringify(item)) continue;
+    const existingRevision = typeof existing.revision === "number" ? existing.revision : 0;
+    const itemRevision = typeof item.revision === "number" ? item.revision : 0;
+    const existingTime = parseDate(existing.updatedAt) || parseDate(existing.createdAt);
+    const itemTime = parseDate(item.updatedAt) || parseDate(item.createdAt);
+    if (existingRevision === itemRevision && existingTime === itemTime) {
+      return { ok: false, summary: `Inbox item ${item.id} differs without a newer revision or timestamp.` };
+    }
+    if (itemRevision > existingRevision || (itemRevision === existingRevision && itemTime > existingTime)) byId.set(item.id, item);
+  }
+  const items = [...byId.values()].sort((a, b) => parseDate(a.createdAt) - parseDate(b.createdAt));
+  return { ok: true, value: { ...right, version: Math.max(Number(left.version) || 1, Number(right.version) || 1), items }, summary: `Merged ${items.length} inbox item(s) by stable ID.` };
+}
+
+const STATE_MAPS = [
+  "sessionFamiliar",
+  "sessionTitles",
+  "sessionArchived",
+  "sessionSacrificed",
+  "sessionKeep",
+  "sessionArchiveExtendedUntil",
+  "sessionOwned",
+  "mergedPrAutoArchived",
+] as const;
+
+const TIMESTAMP_STATE_MAPS = new Set<string>([
+  "sessionArchived",
+  "sessionSacrificed",
+  "sessionKeep",
+  "sessionArchiveExtendedUntil",
+  "sessionOwned",
+]);
+
+function mergeRecordMap(
+  name: string,
+  leftValue: unknown,
+  rightValue: unknown,
+): { ok: true; value: Record<string, unknown> } | { ok: false; summary: string } {
+  const left = record(leftValue);
+  const right = record(rightValue);
+  if (!left || !right) return { ok: false, summary: `State map ${name} is malformed.` };
+  const value = { ...left, ...right };
+  for (const key of new Set([...Object.keys(left), ...Object.keys(right)])) {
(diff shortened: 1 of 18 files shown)


CompleteDotTech:
Reviewed the migration/reconciliation path and pushed follow-up commit `4eb3e2f8` with three fixes:

- validate a changed canonical file before replacing its last known-good Windows compatibility mirror
- stop treating an unchanged mirror as fully migrated when the canonical copy is missing or invalid, so recovery remains visible
- recheck managed mirrors every 30 seconds (paused while hidden/input-active) so writes from older tools surface without restarting Cave

Added regressions for damaged and deleted canonical copies plus the stale-writer poll wiring. Verified with `pnpm typecheck`, `git diff --check`, the targeted migration/banner tests, and the polling-discipline test.

CompleteDotTech:
Follow-up review found and fixed two data-safety gaps in `fdd2392c`:

- Preference reconciliation was using one file-wide revision to choose every divergent section. With independent changes on the legacy and canonical sides, the larger revision could silently discard a unique change from the other snapshot. Non-theme section divergence now stays unresolved for explicit review; theme remains auto-mergeable because it has its own selection revision and timestamp.
- Migration-banner dismissal was keyed only by filenames. If an older Windows tool changed a managed mirror after that filename had been dismissed, the 30-second poll detected it but the stored dismissal suppressed the warning. The key now includes both content hashes and state, so changed data re-surfaces.

Validation passed: `pnpm typecheck`, `pnpm check:tests-wired` (1012), `pnpm test:app` (747), `pnpm test:api` (190), focused migration/banner regressions, and `git diff --check`. CI is rerunning on the pushed commit.

CompleteDotTech:
Follow-up review found and fixed two more data-safety issues in `782eb9f2`:

- If `COVEN_PREFERENCES_PATH`, `COVEN_THEME_PATH`, or `COVEN_BACKDROP_PATH` intentionally pointed at the historical legacy path, reconciliation treated one file as both source and destination and could delete it while creating the compatibility bridge. Same-path overrides are now recognized as already canonical and left intact.
- Travel queue statuses were ranked as monotonic, but failed work can be retried back to syncing. Divergent snapshots now remain unresolved for review rather than replacing a newer retry or suppressing unsynced work without an item revision/timestamp.

Added regression coverage for both cases. Validation passed: targeted migration tests, `pnpm typecheck`, `pnpm check:tests-wired` (1012), `pnpm test:api` (190), and `git diff --check`.

CompleteDotTech:
Fresh review found and fixed one more stale-writer race in `cfa48b53`:

- On non-elevated Windows, an older tool can recreate a legacy file after reconciliation removes it but before the symlink attempt finishes. The mirror fallback previously used an overwriting copy, so that concurrent write could be silently replaced with canonical data. Compatibility and recovery copies now use exclusive creation: the new legacy write is preserved, reconciliation fails closed, and the conflict remains visible for review.

Added a regression that recreates the legacy config during bridge installation and verifies both canonical and stale-writer versions survive. Validation passed: targeted migration test, `pnpm typecheck`, `pnpm check:tests-wired` (1012), `pnpm test:api` (190), and `git diff --check`.

CompleteDotTech:
Fresh review found and fixed one more recovery-retention issue in `713cc68b`:

- When more than 10 entries became unresolved during one reconciliation transaction, backup pruning re-read the last committed journal. It could therefore delete an earlier entry's newly created recovery bundle before the current journal was committed, leaving that conflict pointing at a missing backup. Pruning now uses the in-memory journal so every unresolved/deferred bundle created earlier in the same run is protected.

Added a regression with 11 same-run conflicts that verifies every journaled recovery bundle remains available. Validation passed: focused migration test, `pnpm typecheck`, `pnpm check:tests-wired` (1012), `pnpm test:api` (190), and `git diff --check`.

CompleteDotTech:
Fresh review found and fixed two recovery-path issues in commit `538a4399`:

- **Recover legacy** now validates the verified backup before replacing a valid canonical store, so malformed legacy JSON cannot clobber usable Cave data.
- **Defer** now creates/retains a verified recovery bundle and keeps its `backupId` journaled, so retention pruning cannot remove the only recovery source for a deferred conflict.

Regression coverage was added for both scenarios. Validation passed: focused migration test, migration banner test, migration API route test, `pnpm typecheck`, `pnpm check:tests-wired` (1012 files), and `git diff --check`.

src/lib/server/cave-home-reconciliation.ts
```diff
@@ -0,0 +1,996 @@
...
+  // is the only recovery source for an unresolved/deferred decision.
+  const removable = directories.filter((candidate) => !protectedIds.has(candidate.name));
+  const removeCount = Math.max(0, directories.length - (BACKUP_RETENTION - 1));
+  for (const stale of removeCount > 0 ? removable.slice(-removeCount) : []) {
+    await rm(path.join(migrationBackupRoot(), stale.name), { recursive: true, force: true });
+  }
+}
+
+async function copyPath(source: string, destination: string, kind: PathInfo["kind"]): Promise<void> {
+  if (kind === "dir") await cp(source, destination, { recursive: true, errorOnExist: true });
```
CompleteDotTech: [P1] Make directory copies actually exclusive. With `fs.cp`, `errorOnExist: true` is ignored while the default `force: true` remains enabled. If an older tool recreates `cave-conversations` (or another migrated directory) after `ensureCompatibility` removes it but before symlink fallback, this copy merges into and overwrites that concurrently written directory—the file-copy regression does not cover this directory path. Pass `force: false` (and test the stale directory writer) so the reconciliation fails closed instead of replacing user data.

src/lib/server/cave-home-reconciliation.ts
```diff
@@ -0,0 +1,996 @@
...
+          if (pid) {
+            try {
+              process.kill(pid, 0);
+              alive = true;
+            } catch (ownerError) {
+              alive = (ownerError as NodeJS.ErrnoException).code === "EPERM";
+            }
+          }
+          if (!alive) {
+            await rm(lock, { recursive: true, force: true });
```
CompleteDotTech: [P1] Reclaim stale locks without deleting a successor's lock. When two processes observe the same dead stale lock, both can pass this check; after one removes it and reacquires the path, the other can execute this recursive `rm` against the newly created lock, acquire it itself, and let both migrations mutate the journal/data concurrently. Use an ownership token plus an atomic stale-lock takeover/removal that verifies the same lock instance before deletion.

CompleteDotTech (commented):
Found two remaining cross-process data-safety races in the reconciliation implementation. The focused migration test and `pnpm typecheck` pass, but neither case is currently covered.

CompleteDotTech:
Fixed the two remaining review findings in `6f6adca9`:

- stale lock reclamation now publishes tokenized owners atomically and uses an exclusive takeover claim plus rename, preventing competing reclaimers from deleting a successor lock
- directory compatibility copies now set `force: false`, so a concurrently recreated legacy directory is preserved instead of merged/overwritten

Added regressions for two simultaneous stale-lock observers and for a stale directory writer during compatibility fallback. Validation passed: `pnpm typecheck`, `pnpm check:tests-wired` (1012), `pnpm test:api` (190), five repeated focused migration runs, and `git diff --check`. All 16 remote CI/CodeQL checks passed on the pushed commit, including Windows cross-environment and sidecar runtime.

CompleteDotTech:
Fresh review found and fixed three additional reconciliation issues in `91f88044`:

- a stale legacy writer can change an existing file after backup verification but before compatibility-bridge removal; reconciliation now revalidates the live legacy hash and fails closed instead of deleting that late write
- travel mode fields (`manualOffline`, `staleCache`, reachability/wake state) can transition in both directions and have no per-transition revision, so divergent snapshots now remain unresolved rather than preserving a stale `true`/timestamp automatically
- non-race lock inspection errors no longer enter an unbounded retry loop; only an expected disappearing-lock `ENOENT` retries

Added regressions for the pre-bridge stale-writer window and ambiguous return-online state. Validation passed: focused migration test, `pnpm typecheck`, `pnpm check:tests-wired` (1012 files), `pnpm test:api` (190 files), and `git diff --check`.

CompleteDotTech:
Fresh review found and fixed three additional data-safety gaps in `68876927`:

- compatibility bridge installation now atomically retires and re-hashes the legacy path before replacement, closing the final check-to-delete window for an older writer
- managed-mirror refreshes now stage and hash-verify the validated canonical snapshot before retiring the last known-good mirror, so a concurrent canonical write cannot replace both copies with unvalidated data
- state reconciliation no longer unions one-sided title/archive/keep entries, because those maps delete keys during normal clear/unarchive/unkeep actions and a union could resurrect stale state

Added regressions for legacy and canonical bridge races, managed-mirror refresh races, and deleted state-map keys. Validation passed: `pnpm typecheck`, `pnpm check:tests-wired` (1012 files), focused migration tests, `pnpm test:api` (190 files), and `git diff --check`. CI is rerunning on the pushed commit.

CompleteDotTech:
Fresh review found and fixed three additional data-safety issues in `fca7cf4b`:

- JSON merge and **Recover legacy** could overwrite a canonical write that landed after the recovery bundle was verified. They now merge the verified snapshots and replace canonical storage only if the live destination still matches the backed-up hash; late writes fail closed and remain reviewable.
- Atomic copy installation used destination-replacing `rename`, so a file created by another process during the final install window could be overwritten. File snapshots now publish with an exclusive atomic hard link.
- A compatibility symlink created just before canonical validation failed could leave the original legacy bytes hidden under a retired filename, and arbitrary/broken legacy symlinks were treated as completed migrations. Failed links are now removed before fallback, and only a validated link to canonical storage counts as migrated.

Regression coverage was added for late canonical writes during merge/recovery, post-link validation failure, and foreign legacy symlinks. Validation passed: focused migration tests, `pnpm typecheck`, `pnpm check:tests-wired` (1012 files), `pnpm test:api` (190 files), and `git diff --check`. CI is rerunning on the pushed commit.

CompleteDotTech:
The first CI rerun exposed an unrelated clock-boundary failure in `familiar-analytics-view.test.ts`: at 12:00 UTC today the June 25 fixture crossed the 21-day stale-memory threshold, changing the expected heal-request count from 1 to 2. Follow-up `d6b1f646` passes the model clock through to growth-signal derivation and pins this assertion to a date that intentionally produces one request. The focused analytics test, migration test, typecheck, test-wiring check, and diff check all pass; CI is rerunning on the new head.

CompleteDotTech:
Fresh review found and fixed two remaining migration-safety issues in `2b4ff696`:

- Windows returns `EPERM` when a lock candidate directory is renamed over an existing migration lock. That ordinary contention previously failed the second migration instead of waiting; it is now recognized only when the destination lock exists, and a Windows-native concurrent-lock regression covers it.
- Inbox items are physically deleted without tombstones. Unioning IDs present on only one snapshot could resurrect a reminder the user had deleted, so one-sided IDs now remain unresolved for explicit review while shared IDs still merge by revision/timestamp.

Validation passed: focused migration tests, the Windows-native migration/lock test via `node.exe`, `pnpm typecheck`, `pnpm check:tests-wired` (1012 files), `pnpm test:api` (190 files), and `git diff --check`. CI is rerunning on the pushed commit.

CompleteDotTech:
Fresh review found and fixed one interrupted-upgrade availability issue in `6c825cb9`:

- a crashed migration process left a fresh lock whose dead owner was not examined until the five-minute stale-age threshold; because contenders time out after 10 seconds, config/state/inbox/preferences readers could remain unusable for roughly five minutes after a crash
- recorded dead owners are now reclaimed immediately through the existing tokenized takeover path, while ownerless/incomplete locks still retain the age safeguard

Added a regression for a fresh dead-owner lock. Validation passed: focused migration test, `pnpm typecheck`, `pnpm check:tests-wired` (1012 files), `pnpm test:api` (190 files), and `git diff --check`. CI is rerunning on the pushed commit.

CompleteDotTech:
Fresh review found and fixed a deferred-migration store-gating issue in `04741fb3`:

- **Defer** previously caused later reconciliation runs to skip an entry without validating canonical storage. If the canonical file was missing or became malformed, config/state/inbox/preferences readers could continue against defaults and later overwrite recoverable data.
- Deferral now requires a valid canonical store, deferred entries continue validating canonical storage on every reader/writer gate, and the review UI no longer offers unsafe actions when canonical storage is unavailable.

Added regressions for a deferred canonical file disappearing and for a crafted defer request against a legacy-only store. Validation passed: focused migration/status/banner/API tests, `pnpm typecheck`, `pnpm check:tests-wired` (1012 files), and `git diff --check`. CI is rerunning on the pushed commit.

CompleteDotTech:
Fresh review found and fixed a store-availability regression in `645560d7`:

- reader/writer gates treated any migration error as fatal for every gated store, so an unrelated problem such as a foreign `cave-backdrop.jpg` symlink could take config, state, inbox, and preferences offline
- partial startup results were also discarded globally, causing every later store read to repeat the full migration while that unrelated problem persisted
- gates are now scoped to their own legacy store, partial results remain cached, and only the affected entry is retried after a repair

Added regressions for scoped availability and retry-after-repair. Validation passed: focused migration test, `pnpm typecheck`, `pnpm check:tests-wired` (1012 files), and `git diff --check`. CI is rerunning on the pushed commit.

CompleteDotTech:
Fresh review found and fixed an interrupted-transaction recovery gap in `e283bcbe`:

- a hard process termination after `retireExpectedPath()` renamed a live legacy or canonical path could bypass the `finally` restoration block; on restart, migration treated that side as missing, potentially journaled the legacy bridge as complete, or rebuilt canonical storage from an older compatibility copy while the newest bytes remained hidden under `.migration-retired-*`
- reconciliation now restores the sole atomically retired copy before classifying either path and fails closed if multiple candidates make recovery ambiguous

Added a regression for both legacy and canonical retirement artifacts. Validation passed: focused migration test, `pnpm typecheck`, `pnpm check:tests-wired` (1012 files), and `git diff --check`. CI is rerunning on the pushed commit.

CompleteDotTech:
Fresh review found and fixed two additional issues in `8c90a890`:

- **Recover legacy** was the only action offered when canonical storage was an invalid symlink, but the replacement primitive rejected every symlink, so the advertised recovery could never succeed. Reconciliation now fingerprints the link target, atomically retires only the inspected link, replaces it with the verified legacy backup, and leaves the foreign target untouched.
- The stale-lock concurrency regression was flaky because the test-only `released` probe ran after the lock path was removed, allowing a successor to report `acquired` while the prior probe still counted itself active. The probe now closes the observed critical section before the atomic unlock.

Added the canonical-symlink recovery regression and ran the migration test five times. Validation passed: `pnpm typecheck`, `pnpm check:tests-wired` (1012 files), `pnpm test:api` (190 files), focused migration/status/banner/API-route tests, and `git diff --check`.

CompleteDotTech:
Rebased the PR onto current `main` (`51e468ed`) after #3284 landed and force-updated the head to `037e349a`. The one conflict was the familiar-analytics clock-boundary regression; the rebased version retains the model-clock propagation so the test remains deterministic.

Post-rebase validation passed: migration test five consecutive runs, focused familiar-analytics test (28 assertions), `pnpm typecheck`, `pnpm check:tests-wired` (1017 files), `pnpm test:api` (194 files), `pnpm test:app` (747 files), and `git diff --check`. CI is rerunning on the rebased head.

CompleteDotTech:
Fresh review found and fixed two additional losslessness gaps in `e221a3f1`:

- A deferred/unresolved store was only validated during the original reconciliation run. Once that successful result was cached, deleting or corrupting canonical storage later let the reader gate proceed and potentially fall back to defaults despite a preserved legacy/recovery copy. Gated stores now revalidate canonical storage while a conflict, deferral, or managed mirror remains journaled.
- Inbox items and offline queue entries were merged through ID-keyed maps without rejecting duplicate IDs within one snapshot. Malformed snapshots could therefore silently collapse one record. Duplicate stable IDs now fail closed and remain available for explicit review.

Added regressions for cached deferred-store damage plus duplicate inbox/queue IDs. Validation passed: focused migration test, `pnpm typecheck`, `pnpm check:tests-wired` (1017 files), `pnpm test:api` (194 files), and `git diff --check`.

CompleteDotTech:
Fresh review found and fixed two additional losslessness gaps in :\n\n- A deferred/unresolved store was only validated during the original reconciliation run. Once that successful result was cached, deleting or corrupting canonical storage later let the reader gate proceed and potentially fall back to defaults despite a preserved legacy/recovery copy. Gated stores now revalidate canonical storage while a conflict, deferral, or managed mirror remains journaled.\n- Inbox items and offline queue entries were merged through ID-keyed maps without rejecting duplicate IDs within one snapshot. Malformed snapshots could therefore silently collapse one record. Duplicate stable IDs now fail closed and remain available for explicit review.\n\nAdded regressions for cached deferred-store damage plus duplicate inbox/queue IDs. Validation passed: focused migration test, 
> coven-cave@0.1.1 typecheck /tmp/coven-cave-pr3270-review
> tsc --noEmit, 
> coven-cave@0.1.1 check:tests-wired /tmp/coven-cave-pr3270-review
> node scripts/check-tests-wired.mjs

✓ all 1017 test files wired into CI (1 allowlisted: scripts/release-notes.test.mjs) (1017 files), 
> coven-cave@0.1.1 test:api /tmp/coven-cave-pr3270-review
> node scripts/run-tests.mjs api

running 194 test file(s) [api]
dependency-policy.test.mjs: ok (sharp pinned coherently at 0.34.5)
build-sandbox-runtime.test.mjs: ok
sync-runtimes.test.mjs: ok
surface-claim-guard.test.mjs passed
worktree-guard.test.mjs passed
git-hooks-pre-commit.test.mjs: ok
secret-preflight.test.mjs: ok
uninstall-app.test.mjs: ok
beads-familiar-workflow.test.mjs: ok
beads-pr-bridge.test.mjs: ok
beads-pr-patrol.test.mjs: ok
cave-home-migration.test.ts: ok
cave-home-migration-status.test.ts: ok
global-npm-install-lane.test.ts: ok
✔ parallel Coven CLI and Coven Code requests start exactly one global npm child (0.658925ms)
✔ failure, timeout, and cancellation release the lease so a retry can start (0.159724ms)
✔ the global lease survives module reload and its original cleanup still releases it (1.703943ms)
ℹ tests 3
ℹ suites 0
ℹ pass 3
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 7.75112
cave-home-migration route.test.ts: ok
agent-attachments.test.ts: ok
api-contracts.test.ts: 219 route contracts passed
config route.test.ts: ok
opencoven-executions-route.test.ts: ok
opencoven-submissions-route.test.ts: ok
familiars route.test.ts: ok
avatar route.test.ts: ok
avatar-route.test.ts: ok
familiars/[id]/notes route: all assertions passed
projects icon route.test.ts: ok
chat-model-state route test: ok
prompt enhance route.test.ts: ok
prompts route.test.ts: ok
pack-prompts-route.test.ts: ok
app/latest-release route.test.ts: ok
opencoven-tools/status route.test.ts: ok
onboarding-update route.test.ts: ok
opencoven-tools-status.test.ts: skipped Windows packaged-server probe (requires win32)
✔ local readiness completes without waiting for a blocked registry probe (0.855135ms)
✔ Windows npm.cmd latest checks launch npm-cli.js through Node with fixed argv (1.367363ms)
✔ registry timeouts return explicit freshness and error state (746.446327ms)
✔ a verified newer npm version produces an outdated result (0.694445ms)
✔ missing npm, registry errors, and malformed versions remain explicit failures (648.156562ms)
✔ npm found only via the refreshed environment runs the query with that environment (0.261403ms)
ℹ tests 6
ℹ suites 0
ℹ pass 6
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 1403.080526
daemon status route.test.ts: ok
daemon capabilities route.test.ts: ok
travel client route.test.ts: ok
offline-work-queue.test.ts: ok
executor-status.test.ts: ok
daemon-start.test.ts: ok
daemon-update-lifecycle.test.ts: ok
projects route.test.ts: ok
coven exec route.test.ts: ok
coven-daemon.test.ts: ok
coven-bin.test.ts: ok
harness-version.test.ts (pickVersionLine): ok
hermes-shim: ok
openclaw-bin.test.ts: ok
openclaw-bridge.test.ts: ok
coven-identity-canon.test.ts: ok
copilot-stream: ok
board chat route.test.ts: ok
tool persistence tracker tests passed
model parity routing tests passed
copilot stream routing tests passed
harness-routing tests passed
onboarding-status route.test.ts: ok
onboarding install route.test.ts: ok
onboarding-setup-preserve-config.test.ts: ok
codex-port-preflight route.test.ts: ok
onboarding ssh-check route.test.ts: ok
skills/local route.test.ts: ok
skills/directory/install route.test.ts OK
skills/build route.test.ts OK
skills/draft route.test.ts OK
skills/caveman route.test.ts OK
skills/dry-run route.test.ts OK
skills/directory/use route.test.ts OK
skills-directory.test.ts OK
capabilities route.test.ts: ok
workflows route.test.ts: ok
workflow run route.test.ts: ok
flows run route.test.ts: ok
✔ create route is local-only, bounded, and guarded (0.697255ms)
✔ an explicit project root is resolved before the mission exists (0.143549ms)
✔ list requires a familiar and reconciles persisted missions (0.126286ms)
✔ detail route rejects unsafe ids and reconciles before returning (0.102974ms)
ℹ tests 4
ℹ suites 0
ℹ pass 4
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 5.220404
✔ mission actions are local-only, bounded, path guarded, and serialized by the runner (0.525067ms)
ℹ tests 1
ℹ suites 0
ℹ pass 1
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 8.338252
✔ mission scheduling is local-only, bounded, path guarded, and delegated to the runner (0.541811ms)
ℹ tests 1
ℹ suites 0
ℹ pass 1
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 5.169753
flows webhook route.test.ts: ok
flows webhook-test route.test.ts: ok
flows session-transcript route.test.ts: ok
cave-chat-titles.test.ts ok
session-title-canon: ok
chat-model-state.test.ts: ok
runtime-models.test.ts: ok
✔ contextWindowForModel resolves catalogued ids as known (1.008718ms)
✔ contextWindowForModel tolerates a bare model id (no provider prefix) (0.155314ms)
✔ contextWindowForModel falls back for unknown / non-string ids (0.125434ms)
✔ contextUsedTokens sums input + cache read + cache creation (0.123844ms)
✔ computeContextMeter returns null when no tokens are in the window (1.262932ms)
✔ computeContextMeter computes percent and level against the window (0.138903ms)
✔ computeContextMeter bands warn at >=70% and high at >=90%, clamping at 100% (0.130827ms)
✔ computeContextMeter marks unknown-model windows as estimates (0.089765ms)
✔ formatContextMeter renders a compact label (0.274564ms)
ℹ tests 9
ℹ suites 0
ℹ pass 9
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 8.513364
chat-usage-plan.test.ts: ok
✔ makeUsageWindow clamps used percent and preserves reset metadata (1.184527ms)
✔ makeUsageWindow exposes active usage without fabricating a quota percent (0.148839ms)
✔ aggregateTurnUsage sums all reported token counters and cost (0.211478ms)
✔ buildChatUsagePlanSnapshot renders configured plan consumption (0.298219ms)
✔ buildChatUsagePlanSnapshot is honest when plan limits are unconfigured (0.215885ms)
ℹ tests 5
ℹ suites 0
ℹ pass 5
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 8.528454
cave-conversations.test.ts: ok
cave-conversations cache test OK
session-initiator.test.ts: ok
chat-runtime-scope.test.ts: ok
chat-boundary-sentinel.test.ts: ok
chat-project-access tests passed
✔ empty / null conversations yield no block (0.670705ms)
✔ renders a labelled transcript along the active path (0.324534ms)
✔ drops system, empty, and errored turns (0.17849ms)
✔ windows to the most recent N turns (0.26812ms)
✔ follows the selected branch, not abandoned siblings (0.160116ms)
✔ prependPriorConversation is a no-op for an empty block (0.075557ms)
✔ prependPriorConversation joins block, rule, and prompt (0.126036ms)
✔ missing activeLeafId falls back to chronological order (0.214492ms)
✔ a dangling activeLeafId (no such turn) falls back instead of throwing (0.224167ms)
✔ a parentId cycle does not hang and still yields a bounded block (0.250681ms)
✔ oversized turn text is clamped with a truncation marker (0.177388ms)
✔ maxCharsPerTurn override clamps shorter and is honoured (0.157081ms)
✔ maxTurns override narrows the window (0.136762ms)
✔ MAX_FALLBACK_CHARS_PER_TURN keeps a full window bounded (0.180059ms)
✔ buildResumeRetryPrompt replays history ahead of the live prompt (0.119614ms)
✔ buildResumeRetryPrompt is a transparent passthrough when there is no history (0.119478ms)
✔ buildResumeRetryPrompt reports no replay when only unusable turns exist (0.139267ms)
ℹ tests 17
ℹ suites 0
ℹ pass 17
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 9.366427
✔ store + read roundtrip; only ok payloads with a conversation are stored (0.867175ms)
✔ entries expire after the TTL (0.32259ms)
✔ invalidateConversation drops a single entry (0.302604ms)
✔ cache is bounded: oldest entry evicted beyond the cap (0.264399ms)
✔ prefetch fetches, caches, and dedupes concurrent requests (0.528954ms)
✔ prefetch resolves from a fresh cache entry without a network request (0.17452ms)
✔ failed prefetch caches nothing and never throws (0.503807ms)
✔ hover prefetch fires after the intent delay; cancel disarms it (300.381201ms)
✔ hovering another row re-arms the singleton timer onto the new session (150.972861ms)
✔ chat-list rows arm hover prefetch and invalidate on delete (0.347388ms)
✔ thread-rail rows arm hover prefetch (0.088861ms)
✔ chat-view paints cached payloads, revalidates, and invalidates on send/delete (0.234779ms)
ℹ tests 12
ℹ suites 0
ℹ pass 12
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 460.448094
✔ a full-input scaffold passes the contract with zero violations AND warnings (3.942407ms)
✔ a minimal scaffold (name only) still passes cleanly (0.338075ms)
✔ SOUL name and ward familiar match (cross-file invariant) (0.242408ms)
✔ names with quotes/hashes don't break the ward parser (0.187594ms)
✔ defaults: person falls back to DEFAULT_PERSON and is a protected invariant (0.179187ms)
✔ creatureForGlyph maps known glyphs and falls back to Familiar (0.113389ms)
ℹ tests 6
ℹ suites 0
ℹ pass 6
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 11.47314
session-list-merge.test.ts: ok
session-git-enrich.test.ts: all assertions passed
chat-work-branch.test.ts: all assertions passed
github-tasks-cache.test.ts: ok
swr-cache.test.ts: all assertions passed
memory-file-sources.test.ts: ok
familiar-startup-context.test.ts: ok
▶ buildOperatorProfileContext
  ✔ returns null for empty profiles (0.473986ms)
  ✔ renders only the set fields (0.236036ms)
  ✔ renders links as label — url lines (0.105853ms)
✔ buildOperatorProfileContext (1.466283ms)
ℹ tests 3
ℹ suites 1
ℹ pass 3
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 6.003817
knowledge-vault.test.ts: ok
knowledge-vault-collections.test.ts: ok
knowledge-packs.test.ts: ok
skill-package-install.test.ts: ok
▶ pin kinds and ids
  ✔ accepts only known pin kinds (0.605774ms)
  ✔ validates thread ids as strict slugs (1.139288ms)
✔ pin kinds and ids (2.336195ms)
▶ excerpting and caps
  ✔ collapses whitespace and caps the excerpt (0.199894ms)
  ✔ caps pin content at PIN_CONTENT_MAX (0.150136ms)
  ✔ makePin derives title, excerpt, and capped content (0.152278ms)
✔ excerpting and caps (0.645362ms)
▶ provenance refs
  ✔ compacts pins to kind/ref/title (0.618055ms)
  ✔ normalizes frontmatter pins tolerantly (0.18988ms)
✔ provenance refs (0.929755ms)
▶ sew prompt and output contract
  ✔ builds a sew prompt containing the contract and every pin (0.381704ms)
  ✔ folds a shape's scaffold and tag hints into the prompt without touching the contract (0.145041ms)
  ✔ builds a chat digest that lists pins without full content (0.226026ms)
  ✔ upgrades the chat digest to a finishing brief when the thread id is known (cave-x1za) (0.181696ms)
  ✔ parses well-formed sew output (0.414683ms)
  ✔ tolerates a fenced response and empty tags (0.131169ms)
  ✔ never strips code fences INSIDE the body (only a whole-response wrap) (0.117515ms)
  ✔ rejects malformed output instead of guessing (0.098735ms)
✔ sew prompt and output contract (1.873782ms)
▶ manual sew prefill
  ✔ concatenates pins under headings with source lines (0.167138ms)
  ✔ prefixes a pattern's scaffold as fill-in sections above the pins (cave-kwx4) (0.100253ms)
✔ manual sew prefill (0.357454ms)
ℹ tests 17
ℹ suites 5
ℹ pass 17
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 12.030415
▶ stitch patterns
  ✔ every pattern is a complete, unique shape (0.504138ms)
  ✔ looks up patterns by id and rejects everything else (0.154129ms)
✔ stitch patterns (1.333057ms)
ℹ tests 2
ℹ suites 1
ℹ pass 2
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 7.815062
stitch-threads.test.ts OK
▶ capturePin file containment
  ✔ reads vault files but blocks symlinks that escape the vault root (7.446737ms)
✔ capturePin file containment (8.462083ms)
▶ isPrivateIp — SSRF guard ranges
  ✔ blocks loopback, private, link-local, CGNAT, multicast v4 (0.510137ms)
  ✔ allows public v4 (0.199832ms)
  ✔ blocks loopback, ULA, link-local, and v4-mapped-private v6 (2.545291ms)
  ✔ treats non-IP strings as blocked (resolve first) (0.108437ms)
✔ isPrivateIp — SSRF guard ranges (3.831761ms)
▶ publicOnlyLookup — connect-time SSRF guard
  ✔ refuses hostnames that resolve to private addresses at CONNECT time (1.347339ms)
✔ publicOnlyLookup — connect-time SSRF guard (1.555718ms)
▶ htmlToText
  ✔ extracts title and readable text, dropping script/style (0.74183ms)
  ✔ drops script blocks even with whitespace/attribute end tags (0.235249ms)
  ✔ decodes numeric entities and keeps line structure (0.187093ms)
✔ htmlToText (1.509324ms)
▶ extractHeadingSection
  ✔ slices one section through the next same-or-higher heading (0.278892ms)
  ✔ returns the whole doc for empty or missing headings (0.079228ms)
  ✔ matches case-insensitively (0.081245ms)
✔ extractHeadingSection (0.579715ms)
▶ githubApiTarget
  ✔ maps repo roots to the readme endpoint (0.23788ms)
  ✔ maps issues and PRs to the issues endpoint (0.118909ms)
  ✔ maps blob URLs to raw.githubusercontent.com (0.138416ms)
  ✔ rejects unsupported or unsafe paths (0.137668ms)
✔ githubApiTarget (0.812541ms)
ℹ tests 16
ℹ suites 6
ℹ pass 16
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 20.886378
stitch-sew.test.ts OK
assist-runner.test.ts OK
memory-file-paths.test.ts: ok
skill-file-paths.test.ts: ok
✔ accepts a real .log inside the run-logs dir; rejects outside / non-log / symlink (12.900993ms)
ℹ tests 1
ℹ suites 0
ℹ pass 1
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 19.354126
▶ user avatar file store
  ✔ writes, reads back, and reports mime (4.254263ms)
  ✔ replacing with another format removes the old file (3.585487ms)
  ✔ rejects svg and oversized payloads (3.336719ms)
  ✔ rejects a dataUrl whose header mime disagrees with the declared mime (0.164029ms)
  ✔ delete removes the file; read returns null (5.104928ms)
✔ user avatar file store (18.016269ms)
ℹ tests 5
ℹ suites 1
ℹ pass 5
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 21.610284
▶ profile routes
  ✔ PATCH validates through the shared normalizer, not ad-hoc checks (0.533418ms)
  ✔ profile writes persist via saveConfig (atomic), never raw fs (0.146079ms)
  ✔ /api/config does not accept profile writes (validated route only) (0.16483ms)
  ✔ avatar GET serves bytes with content-type and etag; POST/DELETE go through the file store (0.167763ms)
  ✔ avatar route never accepts svg (0.108837ms)
✔ profile routes (1.98722ms)
ℹ tests 5
ℹ suites 1
ℹ pass 5
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 6.498228
✔ getVoiceProvider returns adapter for known id (0.856184ms)
✔ getVoiceProvider returns gemini stub (0.105696ms)
✔ gemini stub mintSession rejects with not_implemented (0.902287ms)
✔ getVoiceProvider returns the local loop provider (0.095524ms)
✔ getVoiceProvider returns the familiar-brain provider (0.179952ms)
✔ getVoiceProvider returns the elevenlabs provider (0.073092ms)
✔ getVoiceProvider returns null for unknown id (0.134098ms)
✔ listVoiceProviders returns stable order: openai, gemini, local, familiar, elevenlabs (1.84894ms)
ℹ tests 8
ℹ suites 0
ℹ pass 8
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 14.591239
✔ speech registry is static, reviewed, and grouped for readiness consumers (20.573655ms)
✔ path guard keeps model files under the configured coven model root (0.623004ms)
✔ readiness requires both expected size and verified sha256 (172.874224ms)
✔ download writes to a partial file, verifies sha256, then atomically publishes readiness (199.365575ms)
✔ failed checksum downloads leave no ready model behind (138.288114ms)
✔ removeSpeechModel removes only the allow-listed registry model directory (147.73766ms)
ℹ tests 6
ℹ suites 0
ℹ pass 6
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 683.637623
✔ localLlmBaseUrl defaults to the Ollama loopback and honors overrides (0.954183ms)
✔ buildLocalBrainMessages leads with the persona and caps the turn tail (0.251525ms)
✔ buildLocalBrainMessages clamps oversized turns and drops empty ones (review #3159) (0.204225ms)
✔ probeLocalLlm reports ok on 200 and detail on failure (18.664061ms)
✔ mintSession grants a keyless local-loop session carrying the persona (0.932738ms)
✔ mintSession fails actionably when no loopback server answers (0.577067ms)
✔ mintSession probes the COVEN_LOCAL_LLM_URL override (0.552212ms)
ℹ tests 7
ℹ suites 0
ℹ pass 7
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 27.722467
✔ emits each completed sentence exactly once as text accumulates (1.350757ms)
✔ flush returns the unterminated tail once (0.138591ms)
✔ short fragments buffer until a later break instead of tiny utterances (0.17028ms)
✔ question and ellipsis breaks with closing quotes are honored (0.121962ms)
✔ a user final runs the brain, hushes while speaking, listens after the queue drains (1.774694ms)
✔ empty and whitespace finals never reach the brain (0.318356ms)
✔ ears errors surface as VoiceConnectError with the engine's hint (0.180472ms)
✔ mute hushes the ears and disables the mic; unmute listens again (0.145788ms)
✔ close tears down ears, mouth, and mic tracks (0.238962ms)
✔ the session reports how it hears from the supplied ears' label (cave-vpe1) (0.228676ms)
✔ without injected ears and without a window engine the loop refuses with stt_unavailable (0.413465ms)
✔ web ears restart after a silence self-stop, but never after hush or close (0.35121ms)
✔ web ears route finals, partials, and non-routine errors (0.194793ms)
ℹ tests 13
ℹ suites 0
ℹ pass 13
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 13.337077
✔ listen starts a native session; stable partial finishes it; final restarts (2.356233ms)
✔ events from a stale session are dropped (0.371323ms)
✔ hush stops the native session and blocks the auto-restart (0.4313ms)
✔ engine errors surface once and stop listening until asked again (0.357346ms)
✔ an empty final is not brained but listening continues (0.235769ms)
✔ the utterance cap finishes a transcript that never stabilizes (0.282237ms)
✔ close tears down the session and the event channel (0.292595ms)
✔ nativeSttAvailable reflects the probe and never throws (0.344366ms)
✔ strict ears carry requireOnDevice on every native session (0.47757ms)
✔ nativeSttAvailability passes the language and survives probe failures (0.303796ms)
✔ engine selection: on-device wins, strict refuses dictation, fallback is labeled (0.565232ms)
ℹ tests 11
ℹ suites 0
ℹ pass 11
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 11.839221
✔ mintSession grants a keyless familiar-brain session bound to the chat session (1.030488ms)
✔ mintSession rejects without a session to attach to (0.397084ms)
✔ the provider persists its own transcripts (real chat turns) (0.081068ms)
ℹ tests 3
ℹ suites 0
ℹ pass 3
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 8.804098
✔ voice/model id validators are strict path-injection barriers (1.405369ms)
✔ probeElevenLabs distinguishes bad key from unreachable service (178.10061ms)
✔ mintSession grants a proxied session bound to the chat session (7.883295ms)
✔ mintSession applies default voice and model when unset (20.204984ms)
✔ mintSession rejects without a session and on an invalid key (14.097787ms)
✔ the provider persists its own transcripts (real chat turns) (2.359515ms)
✔ parseElevenLabsVoices keeps valid library entries and drops malformed ids (5.967623ms)
✔ parseElevenLabsModels keeps only TTS-capable models (2.900665ms)
ℹ tests 8
ℹ suites 0
ℹ pass 8
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 256.199249
✔ instructions include display_name + role + description + pronouns + note (8.456834ms)
✔ instructions omit blank lines for missing optional fields (2.47516ms)
✔ conversationSeed projects last N turns; default 12 (2.237564ms)
✔ conversationSeed respects custom seedTurns (2.366191ms)
✔ conversationSeed filters out system-role turns (2.487529ms)
✔ conversationSeed is [] when the session file is missing (1.443255ms)
✔ instructions carry the Coven identity canon with the familiar id (1.64953ms)
✔ active roles for this familiar are listed; inactive and foreign roles are not (2.855477ms)
✔ no contract block when the familiar has no identity files (1.697958ms)
✔ an invalid familiar id degrades to persona defaults without throwing (1.095ms)
✔ SOUL.md / IDENTITY.md / MEMORY.md are inlined under FAMILIAR_CONTRACT (3.838001ms)
✔ an oversized SOUL.md is clamped and can never fail the mint (2.787097ms)
✔ familiar-scoped Knowledge Vault entries reach the voice instructions (9.322088ms)
ℹ tests 13
ℹ suites 0
ℹ pass 13
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 47.039328
✔ appends a turn with origin:voice and voiceCallId stamped (6.562241ms)
✔ does not mutate prior turns (10.757304ms)
✔ does nothing when session file is missing (matches appendTurn behavior) (0.448533ms)
ℹ tests 3
ℹ suites 0
ℹ pass 3
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 21.737223
✔ mintSession POSTs to /v1/realtime/client_secrets with bearer auth (2.255956ms)
✔ mintSession wraps model/voice/instructions/transcription inside session.audio (0.423564ms)
✔ mintSession returns grant with provider, clientSecret, expiresAt, connection.kind (0.403491ms)
✔ mintSession grant points SDP exchange at GA /v1/realtime/calls without ?model= (0.357803ms)
✔ mintSession surfaces provider error message verbatim on non-2xx (0.647379ms)
✔ connect: failed SDP exchange throws VoiceConnectError with provider detail as hint (0.637654ms)
✔ connect: failed SDP exchange with non-JSON body still throws coded error, no hint (0.472372ms)
✔ data-channel error events surface provider message as hint under a stable code (0.466735ms)
ℹ tests 8
ℹ suites 0
ℹ pass 8
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 11.519462
✔ catalog covers the full Realtime API output-voice enum (0.997506ms)
✔ every voice carries the traits the picker surfaces (0.285ms)
✔ cedar and marin are flagged realtime-only (TTS preview may reject them) (0.148881ms)
✔ detail line reads Gender · Accent · vibe (0.096057ms)
✔ ballad is the one British-accented voice; the rest are American (0.142162ms)
✔ id lookup helpers agree with the catalog (0.129775ms)
✔ preview text is stable per voice and names the voice (0.142006ms)
ℹ tests 7
ℹ suites 0
ℹ pass 7
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 6.970777
✔ 400 when familiarId missing (8.801195ms)
✔ 400 invalid_session when sessionId is unsafe (2.005339ms)
✔ 404 when familiar not found (9.369052ms)
✔ 400 voice_not_configured when familiar has no voiceProvider (1.31793ms)
✔ 400 unknown_provider when voiceProvider is unrecognized (1.275756ms)
✔ 400 vault_key_unresolved when key not in env/vault (14.051207ms)
✔ 502 provider_mint_failed surfaces provider message verbatim (4.248471ms)
✔ 200 happy path returns grant and ULID-shaped callId (3.193026ms)
✔ 200 familiar-brain provider mints keyless and binds the session id (2.530691ms)
✔ 400 vault_key_unresolved for elevenlabs without ELEVENLABS_API_KEY (7.709752ms)
✔ 200 elevenlabs provider mints with defaults and binds the session id (3.02376ms)
ℹ tests 11
ℹ suites 0
ℹ pass 11
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 61.859005
✔ GET /api/voice/engines advertises readonly readiness and management endpoints (9.874133ms)
✔ download endpoints start only registered model jobs and expose polling state (5.128703ms)
✔ model removal endpoint rejects malformed and unknown model requests (5.403678ms)
ℹ tests 3
ℹ suites 0
ℹ pass 3
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 23.981242
✔ 400 unknown_voice for ids outside the realtime catalog (4.515878ms)
✔ 400 vault_key_unresolved when no OpenAI key is configured (12.906362ms)
✔ 200 happy path streams audio/mpeg and asks TTS for the picked voice (1.836458ms)
✔ per-voice cache: the second preview never re-bills the provider (0.742934ms)
✔ 422 preview_unsupported when TTS rejects the voice param, and the verdict is cached (1.094408ms)
✔ a 400 unrelated to the voice param is NOT cached as unsupported (1.008721ms)
✔ 502 preview_failed surfaces the provider message on non-400 failures (0.666462ms)
✔ 502 provider_unreachable when the TTS fetch itself throws (0.38622ms)
ℹ tests 8
ℹ suites 0
ℹ pass 8
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 36.951945
✔ 400 on invalid json and missing/invalid messages (5.857122ms)
✔ 502 local_llm_unreachable with an actionable hint when the loopback is down (0.962752ms)
✔ 502 local_llm_error surfaces the server's own message (0.925613ms)
✔ 502 local_llm_empty when the model answers with no text (0.674106ms)
✔ forwards system + turns to the loopback chat-completions API and returns the text (1.231029ms)
✔ a system message with empty content still counts as the persona slot (0.74569ms)
✔ honors the COVEN_LOCAL_LLM_URL base override (0.644008ms)
ℹ tests 7
ℹ suites 0
ℹ pass 7
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 1960.65046
✔ 400 invalid_json on an unparseable body (3.735794ms)
✔ 400 on missing or oversized text (1.088947ms)
✔ 400 on a path-shaped voice id or malformed model id (injection barrier) (1.278063ms)
✔ 400 vault_key_unresolved when ELEVENLABS_API_KEY is not set (13.248962ms)
✔ 200 proxies MP3 bytes and keeps the key server-side (1.585735ms)
✔ 502 with a hint when ElevenLabs rejects, 502 key-invalid on 401 (1.047336ms)
ℹ tests 6
ℹ suites 0
ℹ pass 6
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 26.56261
✔ 400 vault_key_unresolved when ELEVENLABS_API_KEY is not set (16.956224ms)
✔ 200 maps saved voices and TTS-capable models, dropping malformed entries (1.982297ms)
✔ 502 elevenlabs_key_invalid when either upstream call returns 401 (1.934798ms)
✔ 502 with the upstream status when a call fails non-401 (0.730445ms)
✔ 502 elevenlabs_unreachable when the network call throws (0.56694ms)
ℹ tests 5
ℹ suites 0
ℹ pass 5
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 26.725721
✔ 400 when sessionId missing (3.10872ms)
✔ 400 when callId missing (0.523632ms)
✔ 400 when role invalid (0.442217ms)
✔ 400 when text missing (0.382592ms)
✔ 400 invalid_session for unsafe sessionId (0.689301ms)
✔ 200 happy path appends a voice-origin turn (6.159943ms)
✔ a second call appends a second turn (no overwrite) (4.055619ms)
ℹ tests 7
ℹ suites 0
ℹ pass 7
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 18.952242
server-pty-ws.test.ts OK
pty-ws-bridge.test.ts OK
pty-ws-bridge reconnect assertions: ok
pty-ws-bridge iOS-resume assertions: ok
pty-ws-bridge kill-frame assertions: ok
✔ macOS release signing includes node-pty spawn-helper Mach-O files (0.797622ms)
✔ sidecar bundle restores executable mode for node-pty spawn-helper (0.129059ms)
✔ notary rejection stops before stapling and prints the Apple log (0.139324ms)
✔ DMG packaging retries transient hdiutil resource-busy failures (0.243922ms)
✔ DMG packaging applies a branded Finder background and icon layout (0.920836ms)
✔ Linux release job forces AppImage extract-and-run mode (0.182352ms)
✔ Linux AppImage strips bundled GLib/libmount so host libraries stay ABI-compatible (0.215176ms)
✔ manual release retries can build from a source ref while attaching to the tag (0.123931ms)
✔ sidecar bundle prunes foreign native packages before release bundling (0.332128ms)
✔ release.sh retries its network-dependent steps (0.290277ms)
ℹ tests 10
ℹ suites 0
ℹ pass 10
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 10.958041
✔ signed artifact selection chooses the first signed match (0.85211ms)
✔ signed artifact selection skips unsigned matches (0.14311ms)
ℹ tests 2
ℹ suites 0
ℹ pass 2
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 6.056604
stamp-release.test.mjs: ok
daemon start route.test.ts: ok
github-assigned-route.test.ts OK
github-commit-route.test.ts OK
github-dispatch-route.test.ts OK
github-merge-route.test.ts OK
github-rerun-route.test.ts OK
github-review-route.test.ts OK
github-issue-route.test.ts OK
github-runs-route.test.ts OK
✅  stripMdxLeakage tests passed (8 sections)
proxy-behavior.test.ts: ok
memory-file-sources-coven-familiar.test: ok
memory-trash.test: ok
memory-file-write.test: ok
memory-file-inventory.test: ok
space-usage.test.ts: ok
preferences-store.test.ts: ok
✔ path is resolved at call-time from override or COVEN_HOME (0.796863ms)
✔ safe raster signatures are detected and SVG is never accepted (0.182248ms)
✔ write/read round-trips bytes, MIME, and a stable strong ETag (4.884624ms)
✔ replacement changes format atomically and delete is idempotent (5.883423ms)
✔ validation rejects empty, oversized, unsupported, malformed, and mismatched data (0.990592ms)
✔ malformed or oversized data already on disk is not served (15.810488ms)
✔ concurrent writes never tear data or share temp names (3.960079ms)
✔ a failed atomic rename cleans its unique temp file (3.397301ms)
✔ route keeps normal API auth plus local-only, bounded, no-store raster semantics (0.431807ms)
✔ familiar backdrops round-trip in their own directory and delete idempotently (5.694711ms)
✔ traversal-shaped familiar ids are refused with a 400 validation error (0.408073ms)
✔ familiar route mirrors the app route's local-only, bounded, no-store semantics (0.199772ms)
ℹ tests 12
ℹ suites 0
ℹ pass 12
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 51.005372
✔ direct bd remains the first and only path when available (2.277545ms)
✔ Windows falls back to WSL with translated cwd and argv-safe Beads args (0.507979ms)
✔ missing direct and WSL CLIs return an actionable service-unavailable result (0.300938ms)
✔ unexpected direct bd failures remain visible and do not switch runtimes (0.404993ms)
✔ unexpected WSL bd failures preserve the command error (0.283257ms)
ℹ tests 5
ℹ suites 0
ℹ pass 5
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 8.554495
✔ valid native repo and worktree roots resolve on every platform (80.23115ms)
✔ missing, non-repository, and relative roots return stable actionable 4xx results (8.339767ms)
ℹ tests 2
ℹ suites 0
ℹ pass 2
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 99.908747
theme-store.test.ts: ok
memory-coven-workspaces.test: ok
memory-excerpt.test: ok
memory-mutation-routes.test: ok
memory-trash-excluded.test: ok
salem/pathfinder route test: ok
changes route.test.ts: ok
project-grants route.test.ts: ok
access-groups route.test.ts: ok
project-permission-routes.test.ts: ok
project-local-human-read.test.ts: ok
project-file route.test.ts: ok
salem/pathfinder/feedback route test: ok
feedback/message route test: ok
sidecar archive determinism: ok (cc9cfc01759a3598c396149be223b40f3ae0ae1454dde6cc9a52456aaecb5eaa, 243 bytes)
sidecar-bundle security test: ok
sidecar-bundle-deps.test: ok
sidecar-runtime-closure.test.mjs: ok
standalone-budget.test.mjs: ok
tray-icon.test.mjs ok
sessions/[id] security route.test.ts: ok
prune-response: all assertions passed
familiar-contract-files.test.ts: ok
project-paths.test.ts: ok
✔ empty/absent request yields no segments (the root itself) (1.16677ms)
✔ a relative subpath yields its clean segments (0.241939ms)
✔ an absolute path inside the root is accepted (0.117885ms)
✔ escaping the root is rejected (0.208084ms)
✔ resolveWithinRoot only descends into directories that actually exist (2.06722ms)
ℹ tests 5
ℹ suites 0
ℹ pass 5
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 8.618097
session-project-roots.test.ts: ok
familiar-avatar.test.ts: ok
skill-scan.test.ts: ok
skill-build.test.ts OK
prompt-scan.test.ts: ok

✓ 194 test file(s) passed [api] (194 files), and .

CompleteDotTech:
Fresh review found and fixed two additional issues in `d75bf0a3` (rebased onto current `main`):

- migration status and recovery actions now require a local request, and action JSON is size-bounded; this prevents remote/tailnet clients from reading local migration paths and hashes or selecting destructive recovery choices
- divergent directories no longer auto-copy legacy-only children, because conversation files can be deleted and an automatic union could resurrect a deleted transcript; directory copies now require an explicit **Merge safely** action

Added regressions for the local/bounded route contract and conservative directory reconciliation, including the shared API contract registry. Validation passed: focused migration and route tests, `pnpm typecheck`, `pnpm check:tests-wired` (1017 files), `pnpm test:api` (194 files), and `git diff --check`. All remote CI/CodeQL checks now pass on the updated head, including Windows cross-environment, Windows sidecar runtime, E2E, Frontend build, and Rust check.

CompleteDotTech:
Fresh review found one additional concurrency gap: the config/state/inbox/preferences gates only waited for startup reconciliation, then released the migration lock before the actual store read or read-modify-write. A second Cave process (or a manual Review action) could therefore replace canonical storage between a store's read and write, allowing a stale writer to overwrite the recovered/merged result.

Fixed in `cfbe91a8` by sharing the cross-process migration lock with complete store transactions, including inbox scheduler/daily-summary writes and preference/config/state mutations. Added a regression that pauses recovery at the canonical replacement boundary and verifies a store transaction cannot enter until recovery finishes.

Validation:
- `pnpm typecheck`
- `pnpm check:tests-wired` (1017)
- `pnpm test:api` (194 files)
- focused migration, inbox, config race, preferences, and theme-store tests
- `git diff --check`

CI is rerunning on the new head.

CompleteDotTech:
CI follow-up: Frontend build exposed a source-contract smoke assertion requiring `loadInbox` to remain an explicit async function. Preserved that declaration in `3d368c63`. Local validation now includes `pnpm test:app` (747 files), the dashboard smoke test, typecheck, and diff check. CI restarted on the new head.

CompleteDotTech:
Remote verification is complete on `3d368c63`: all visible CI and CodeQL checks pass, including Frontend build, E2E, Windows cross-environment, and Windows sidecar runtime. The PR is mergeable and `CLEAN`.

PR #3272, opened 2026-07-16
Cache and decouple GitHub task polling

### LINKED ISSUE #3267: Cache and decouple GitHub task refreshes from the four-second session poll
## Performance problem

`loadSessions()` in `src/components/workspace.tsx` starts a `/api/github/tasks` request every time the four-second session poll runs. `src/app/api/github/tasks/route.ts` then proxies the configured `coven-github` endpoint with `cache: "no-store"` and no process cache, ETag handling, stale-while-revalidate, or in-flight coalescing.

A single foreground client can therefore generate roughly **900 proxy/upstream requests per hour**, even when GitHub assignments are unchanged. Multiple Cave windows multiply that load. The result is unnecessary network traffic, JSON parsing, upstream work/rate-limit pressure, and contention with interactive chat/session requests.

GitHub task freshness does not need to match the local four-second session cadence.

## Suggested direction

- Split GitHub task refresh from `loadSessions()` and use a slower/event-driven cadence.
- Add server-side TTL/SWR plus in-flight request coalescing for the upstream proxy.
- Use ETags/conditional requests if the upstream supports them.
- Refresh immediately on relevant GitHub events or user actions, while pausing in hidden windows and during active input.

## Acceptance criteria

- Idle foreground usage no longer calls `/api/github/tasks` every four seconds.
- Record and enforce a documented maximum idle request rate; capture before/after counts over one hour.
- Multiple simultaneous clients share one upstream refresh within the cache window.
- GitHub assigned count/session enrichment updates promptly after a relevant action or explicit refresh.
- Error responses do not erase the last-known-good task context.

CompleteDotTech:
Closes #3267.

## What changed

- decouple GitHub task enrichment from the four-second session polling loop
- poll GitHub task context on a pausable five-minute client cadence, reducing uninterrupted foreground traffic from roughly 900 Cave requests/hour to at most 12
- add a one-minute process-local server TTL with stale-while-revalidate, in-flight request coalescing, and ETag/304 revalidation
- preserve last-known-good task context when the upstream service fails
- provide a forced refresh path for explicit GitHub refresh and linking actions
- add focused cache tests and update API/component contract coverage

## Why

GitHub task refreshes were piggybacking on the session loop even though the data changes much less frequently. That generated an upstream request every four seconds for every foreground client. Separating the cadences and sharing a short server-side cache removes that unnecessary load while keeping explicit user refreshes immediate.

## Validation

- `pnpm typecheck`
- `pnpm check:tests-wired` — 1,012 tests wired
- `pnpm test:api` — 191 test files passed
- `pnpm test:app` — 747 test files passed
- focused GitHub task cache and component contract tests
- `git diff --check`

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -869,6 +869,7 @@ export const SUITES = {
     "src/lib/session-list-merge.test.ts",
     "src/lib/session-git-enrich.test.ts",
     "src/lib/server/chat-work-branch.test.ts",
+    "src/lib/github-tasks-cache.test.ts",
     "src/lib/swr-cache.test.ts",
     "src/lib/server/memory-file-sources.test.ts",
     "src/lib/server/familiar-startup-context.test.ts",
--- a/src/app/api/api-contracts.test.ts
+++ b/src/app/api/api-contracts.test.ts
@@ -98,7 +98,7 @@ const contracts: RouteContract[] = [
   { route: "/github/review", methods: ["POST"], kind: "json", readsJson: true, invalidJson: "guarded" },
   { route: "/github/runs", methods: ["GET"], kind: "json" },
   { route: "/github/pat", methods: ["GET", "POST", "DELETE"], kind: "json", readsJson: true, invalidJson: "fallback-empty" },
-  { route: "/github/tasks", methods: ["GET"], kind: "json" },
+  { route: "/github/tasks", methods: ["GET", "POST"], kind: "json" },
   { route: "/github/worktree", methods: ["POST"], kind: "json", readsJson: true, invalidJson: "guarded" },
   { route: "/grant-proposals/[id]", methods: ["PATCH"], kind: "json", readsJson: true, invalidJson: "guarded" },
   { route: "/grant-proposals", methods: ["GET", "POST"], kind: "json", readsJson: true, invalidJson: "guarded" },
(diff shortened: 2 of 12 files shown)


CompleteDotTech:
Reviewed the full PR and pushed `af54963e` with fixes for two issues found:

- Preserved forced/manual task refresh results when a scheduled GET overlaps the POST; the slower authoritative refresh can no longer be discarded in favor of an older cache hit.
- Stopped task-popover cancellation from triggering a forced upstream refresh; linked-work refresh now runs only after a successful board mutation.

Re-review after the fixes found no remaining actionable issues.

Validation:
- `pnpm typecheck`
- `pnpm check:tests-wired` (1,012 files wired)
- `pnpm test:api` (191 files passed)
- `pnpm test:app` (747 files passed)
- `git diff --check`

CompleteDotTech:
Completed another independent full review of the updated head `af54963e` across all 12 changed files. I found no additional actionable issues, so no further code changes were needed in this pass.

Verified again locally:
- `pnpm typecheck`
- `pnpm check:tests-wired` — 1,012 files wired
- `pnpm test:api` — 191 files passed
- `pnpm test:app` — 747 files passed

The earlier forced-refresh race and cancel-path refresh issues remain resolved, and the current GitHub check rollup is green.

PR #3279, opened 2026-07-16
fix(server): replace deprecated request URL parsing

### LINKED ISSUE #3278: Replace deprecated url.parse() in the custom server
## Source

On Node 24.14.1, ordinary requests handled by the custom server emit `DEP0169`, pointing to the `url.parse()` call in `server.ts`. The same parser is also used for PTY WebSocket upgrade targets.

## Root cause

`server.ts` imports `parse` from `node:url` and calls `parse(req.url ?? "/", true)` for every HTTP request and WebSocket upgrade. Node identifies this parser as nonstandard and potentially error-prone.

## Suggested fix shape

Let Next.js own ordinary HTTP request parsing. Parse WebSocket upgrade targets with the WHATWG `URL` API and a fixed internal base while preserving existing route classification, duplicate-query ordering, null-prototype query records, the legacy 1,000-pair bound, authentication behavior, and non-PTY forwarding. Malformed upgrade targets should receive HTTP 400, and the generated `server.mjs` artifact and focused server contracts should be updated.

## Potential User Impact

This would remove runtime deprecation noise, improve compatibility with current and future Node releases, and harden parsing on the local/mobile terminal path without changing public routes, credentials, or WebSocket wire formats.

snowopsdev:
## Summary

Replace the custom server's deprecated `node:url.parse()` calls with Next-owned HTTP parsing and a compatibility-preserving WHATWG parser for WebSocket upgrade targets.

## Why

Node 24 emits `DEP0169` for ordinary requests because the custom server pre-parses every URL with the legacy parser. The same deprecated path also handles the native/mobile PTY WebSocket upgrade.

Fixes #3278

## Changes

- Let Next.js parse ordinary HTTP requests through the two-argument request handler.
- Parse upgrade targets with WHATWG `URL` using a fixed internal base rather than the request `Host` header.
- Preserve authority-looking origin-form route classification, duplicate-query order, null-prototype query records, first-value authentication behavior, and the legacy 1,000-segment `maxKeys` cutoff.
- Return HTTP 400 and close the socket for malformed upgrade targets while forwarding valid non-PTY upgrades unchanged.
- Regenerate `server.mjs` and extend the focused source-contract and executable behavioral coverage.

## Review follow-up

The first implementation capped entries emitted by `URLSearchParams`, but legacy `querystring.parse(..., { maxKeys: 1000 })` counts raw `&`-separated segments, including empty segments. That meant a credential after 1,000 empty segments was ignored before this change but accepted by the initial WHATWG implementation; it also allowed the full query to be parsed before enforcing the cap.

The parser now truncates the raw query at the legacy 1,000-segment boundary before assigning it to `URL.search`. An executable regression covers both sides of the boundary: the 1,000th segment is parsed, while a token beyond 1,000 empty segments remains ignored.

A maintainer then caught a second compatibility edge: in a target such as `/api/pty-ws??coven_access_token=value`, the second `?` is query-key data under the legacy parser. Assigning the bounded query directly to `URL.search` consumed it as a delimiter and exposed a valid credential key. The parser now supplies the WHATWG delimiter separately, preserving `?coven_access_token`; executable coverage asserts that the malformed target cannot authenticate.

## Verification

- `pnpm build:server`
- `node src/server-pty-ws.test.ts`
- `pnpm typecheck`
- `pnpm check:tests-wired`
- `pnpm test:app` — 747 passed before the delimiter follow-up; the follow-up rerun reached 452 files before an unchanged analytics assertion failed (`familiar-analytics-view.test.ts:426`, expected 1 / observed 2), and that failure reproduces in isolation
- `pnpm test:api` — 190 passed
- `pnpm test:mobile` — 80 passed
- `pnpm build`
- `(cd src-tauri && cargo check --locked)`
- Differential parser cases against Node 24 legacy `url.parse`: duplicate keys, prototype-like keys, malformed escapes, query-cutoff boundaries, authority-looking paths, and absolute-form targets.
- Isolated Node 24 runtime smoke: ordinary HTTP, missing/valid PTY credentials, duplicate-token ordering, malformed-target 400 with post-error recovery, and authority-looking non-PTY forwarding; no `DEP0169` emitted.
- Native Tauri smoke: Home rendered, project-backed code rail opened, and the Terminal PTY started in the expected worktree; no URL-parser warning emitted.

## Risk + rollout notes

No migration, dependency, public route, credential, or WebSocket frame changes. The parser intentionally retains legacy edge behavior needed by authentication and route classification. Rollback requires reverting `server.ts` and the focused test together, then regenerating `server.mjs`.


server.ts
```diff
@@ -214,6 +213,69 @@ function firstQueryValue(value: string | string[] | undefined): string | undefin
...
+  // untouched when a valid non-PTY upgrade is forwarded to Next.
+  const rootedPath = normalizedPath.startsWith("/") ? normalizedPath : `/${normalizedPath}`;
+  const parsedUrl = absoluteForm
+    ? new URL(normalizedPath)
+    : new URL(`/.${rootedPath}`, UPGRADE_URL_BASE);
+
+  // Bound the raw `&`-separated segments before URLSearchParams sees them.
+  // querystring.parse() counts empty segments toward maxKeys, while iterating
+  // URLSearchParams alone would skip them and parse the entire query first.
+  parsedUrl.search = boundedUpgradeQuery(suffix);
```
CompleteDotTech: [P2] Preserve a leading question mark in query data

When an upgrade target contains a second question mark, such as `/api/pty-ws??coven_access_token=value`, legacy `url.parse(..., true)` treats the second `?` as part of the key (`?coven_access_token`), but `boundedUpgradeQuery()` returns a string beginning with `?` and assigning that to `URL.search` consumes it as another delimiter. The new parser therefore exposes `coven_access_token` and changes authentication behavior; pass the bounded raw query through `URLSearchParams` without losing a data-leading `?` (and add this case to the differential coverage).

CompleteDotTech (commented):
Found one compatibility issue in the upgrade query parser.

server.ts
snowopsdev: Thank you for catching this, @romgenie — you’re right. I reproduced it on Node 24.14.1: legacy parsing keeps the second `?` as data and produces `?coven_access_token`, while the current branch exposes `coven_access_token`. That changes the authentication result for this malformed target.

My suggested fix is to preserve the URL delimiter separately:

```ts
parsedUrl.search = `?${boundedUpgradeQuery(suffix)}`;
```

The prefixed `?` is consumed as the WHATWG query delimiter, leaving any leading `?` from the bounded raw query as key data. I’d add the double-`?` case to the executable differential regression, regenerate `server.mjs`, and rerun the focused and repository gates.

I also checked this shape against no/empty queries, fragments, multiple leading question marks, duplicate keys, and the 999/1,000 empty-segment cutoff; those cases match the legacy parser. Thanks again for the precise compatibility catch.

CompleteDotTech: That fix direction matches the legacy behavior for the double-`?` case. Please apply it, add the executable regression, and regenerate `server.mjs`; I’m leaving this thread open and requesting changes until the fix is on the branch.

CompleteDotTech (changes requested):
Requesting changes for the unresolved upgrade-query compatibility issue: a second `?` is currently consumed as a delimiter, which can turn a malformed query key into the valid PTY credential key and change authentication behavior. The proposed delimiter-preserving fix and regression should land in both `server.ts` and generated `server.mjs` before merge.

server.ts
snowopsdev: Implemented in [`d138b6e4`](https://github.com/OpenCoven/coven-cave/pull/3279/commits/d138b6e49e57c3c916e2145637511c5d59a07071). The bounded query is now assigned with a separate WHATWG delimiter, so the second `?` remains part of `?coven_access_token` rather than becoming the valid credential key.

I added executable coverage asserting both sides of that authentication behavior and regenerated `server.mjs`.

Verification:

- `pnpm build:server`
- `node src/server-pty-ws.test.ts`
- `pnpm typecheck`
- `pnpm check:tests-wired`
- `pnpm build`
- `git diff --check`

The full `pnpm test:app` run reached 452 files before stopping on an unchanged analytics assertion in `src/components/familiar-analytics-view.test.ts:426` (`expected 1`, `observed 2`); the same failure reproduces when that test is run alone. I left that unrelated path untouched to keep this review fix scoped.

The new CI run is awaiting maintainer approval before jobs can start. Thanks again for catching this edge case and confirming the fix direction.

CompleteDotTech (approved):
Re-reviewed the latest head after the delimiter-preserving follow-up. The second `?` now remains query-key data, the executable regression covers the authentication boundary, and `server.mjs` matches the generated `server.ts` output. I found no further blocking issues. Verified locally with `node src/server-pty-ws.test.ts`, `pnpm build:server`, `pnpm typecheck`, and `pnpm check:tests-wired`.

CompleteDotTech (approved):
Re-reviewed the latest head after the author’s delimiter-preserving follow-up and the base sync. I found no remaining actionable issues: the second `?` stays query-key data, the focused executable regression covers the authentication boundary, and `server.mjs` remains generated from `server.ts`.

The prior Frontend build failure was from the clock-sensitive familiar analytics test already fixed on current `main`; I updated this branch with `main`, approved the fork workflow run, and confirmed the replacement CI rollup is fully green (11/11). Local verification also passed: `node src/server-pty-ws.test.ts`, `pnpm typecheck`, `pnpm check:tests-wired`, `pnpm build:server` with no generated diff, `git diff --check`, and the formerly failing analytics test in isolation.

PR #3282, opened 2026-07-16
Name the Cave runtime host when summoning a familiar

### LINKED ISSUE #3281: Clarify “This machine” in the Summoning Circle with the Cave host name
## Problem

In **Summon familiar → The vessel**, the first option is labelled **“This machine”** with the hint **“Runs on a runtime installed here.”**

When Cave is being viewed from another device or is connected to remote infrastructure, “this” is ambiguous:

- the device displaying the Cave UI;
- the host running the Cave backend;
- or the host running the Coven daemon/hub.

That makes it difficult to know where the familiar's local harness will actually execute.

## Current implementation

The label is currently hard-coded in `src/components/familiar-summoning-circle.tsx`:

```ts
{ kind: "local", icon: "ph:desktop", title: "This machine", hint: "Runs on a runtime installed here." }
```

The app already has a hostname-backed precedent: `GET /api/hosts?probe=0` builds its local host option from Node's `hostname()`.

## Proposed UX

Show the Cave/runtime host explicitly, for example:

- **Local runtime — haven**
- hint: **Runs on haven, the host serving this Cave.**

A shorter alternative is **haven (this Cave host)**.

If the hostname cannot be obtained, fall back to **This Cave host** rather than **This machine**.

This should identify the Cave backend/runtime host, not imply that it is the browser/viewing device or a separately configured remote daemon/hub.

## Acceptance criteria

- [ ] The local vessel option includes the hostname of the host on which the local runtime will execute.
- [ ] Supporting copy distinguishes that host from the device displaying the UI and from a remote daemon/hub.
- [ ] The UI has a safe, unambiguous fallback such as **This Cave host** when hostname discovery fails.
- [ ] Tests cover both the hostname-backed label and fallback.
- [ ] The SSH and OpenClaw vessel options remain unchanged.

## Context

The Summoning Circle was introduced in #2635. The existing hostname-backed host picker in #2337 appears reusable for this change.

iamcobolt:
## Summary

- replace the ambiguous **This machine** vessel label with **Local runtime — `hostname`**
- explain that the hostname identifies the host serving Cave and running the local harness
- use **This Cave host** as a safe fallback while hostname discovery is unavailable
- carry the same hostname into the final summoning recap
- explicitly persist `{ kind: "local" }` for local summons so an SSH workspace default cannot contradict the selected vessel
- add regression coverage for the hostname copy, local payload, runtime normalization, persistence, and binding precedence

## Why

When Cave connects to remote infrastructure or is viewed from another device, “this machine” can mean the viewing device, the Cave backend host, or the daemon/hub. Local familiar execution happens on the Cave backend host; daemon and hub targeting are separate.

The Summoning Circle already fetches `/api/harnesses` to discover runtimes. This change adds `runtimeHost` to that response, avoiding an extra request and continuing to work for remote Cave clients. The existing `/api/hosts` hostname source is loopback-only.

Selecting the local vessel now also records an explicit local familiar binding. Callers that omit a runtime keep the existing default-inheritance behavior, while OpenClaw remains runtime-free.

## Validation

- focused `familiar-summoning-circle.test.ts`, `onboarding-familiars.test.ts`, and `cave-config.test.ts`
- `corepack pnpm@10.34.0 typecheck`
- `git diff --check`
- full `test:app` run reached 452 passing files, then stopped at the unrelated existing `familiar-analytics-view.test.ts` heal-count assertion (`2 !== 1`); the same test fails identically in isolation

Closes #3281

### DIFF
--- a/src/components/familiar-summoning-circle.tsx
+++ b/src/components/familiar-summoning-circle.tsx
@@ -312,6 +312,7 @@ function SummoningRite({
   // Stage I — the vessel.
   const [vessel, setVessel] = useState<VesselKind | null>(draftVessel);
   const [harnesses, setHarnesses] = useState<HarnessReport[] | null>(null);
+  const [localHost, setLocalHost] = useState<string | null>(null);
   const [harness, setHarness] = useState<string | null>(
     draft?.harness && isSummonableLocalHarness(draft.harness)
       ? draft.harness
@@ -373,11 +374,20 @@ function SummoningRite({
     (async () => {
       try {
         const res = await fetch("/api/harnesses", { cache: "no-store" });
-        const json = (await res.json()) as { ok?: boolean; harnesses?: (HarnessReport & { binary?: string })[] };
+        const json = (await res.json()) as {
+          ok?: boolean;
+          runtimeHost?: unknown;
+          harnesses?: (HarnessReport & { binary?: string })[];
+        };
         if (cancelled) return;
-        if (res.ok && json.ok !== false && (json.harnesses ?? []).length > 0) {
-          setHarnesses(json.harnesses!.filter((h) => h.chatSupported && isSummonableLocalHarness(h.id)));
-          return;
+        if (res.ok && json.ok !== false) {
+          const runtimeHost =
+            typeof json.runtimeHost === "string" ? json.runtimeHost.trim() : "";
+          setLocalHost(runtimeHost || null);
+          if ((json.harnesses ?? []).length > 0) {
+            setHarnesses(json.harnesses!.filter((h) => h.chatSupported && isSummonableLocalHarness(h.id)));
+            return;
+          }
         }
         throw new Error("empty");
       } catch {
@@ -660,6 +670,7 @@ function SummoningRite({
                 {stage === 0 ? (
                   <StageVessel
                     vessel={vessel}
+                    localHost={localHost}
                     setVessel={(v) => {
                       setVessel(v);
                       setError(null);
@@ -718,7 +729,9 @@ function SummoningRite({
                       vessel === "openclaw"
                         ? `OpenClaw · ${selectedAgent?.displayName ?? agentId ?? ""}`
                         : `${(harnesses ?? []).find((h) => h.id === harness)?.label ?? harness ?? ""}${
-                            vessel === "ssh" ? ` over SSH · ${sshHost.trim()}` : " on this machine"
+                            vessel === "ssh"
+                              ? ` over SSH · ${sshHost.trim()}`
+                              : ` on ${localHost ?? "this Cave host"}`
                           }`
                     }
                     name={name.trim()}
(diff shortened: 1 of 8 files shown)


src/components/familiar-summoning-circle.tsx
```diff
@@ -832,7 +847,14 @@ function StageVessel({
   onTestSsh: () => void;
 }) {
   const vessels: { kind: VesselKind; icon: IconName; title: string; hint: string }[] = [
-    { kind: "local", icon: "ph:desktop", title: "This machine", hint: "Runs on a runtime installed here." },
+    {
+      kind: "local",
+      icon: "ph:desktop",
+      title: localHost ? `Local runtime — ${localHost}` : "This Cave host",
```
CompleteDotTech, line 855: **[P1] Bind the selected local vessel to this host**

When `config.defaults.runtime` is SSH, selecting this “local” option still omits `familiar.runtime` in `handleSummon`; `POST /api/familiars` then stores no runtime and `bindingFor()` inherits `config.defaults.runtime`, so chats actually execute remotely while this copy and the recap promise `${localHost}`. Persist an explicit local runtime (and ensure normalization preserves it), or otherwise make the created binding override the remote default before displaying this claim.

CompleteDotTech (changes requested):
Found one execution-location correctness issue: the local vessel can still inherit a remote default runtime.

src/components/familiar-summoning-circle.tsx
iamcobolt: Fixed in `6d90d0ee`. The local vessel now sends an explicit `runtime: { kind: "local" }`, and `normalizeFamiliarDraft()` preserves that as a `FamiliarRuntime` so the existing API persistence path records it. Added regressions showing the local runtime survives normalization/config persistence, overrides an SSH workspace default, and leaves omitted-runtime inheritance unchanged. Focused tests, TypeScript typecheck, and `git diff --check` pass.

CompleteDotTech (approved):
Re-reviewed the full six-file diff at `6d90d0ee`. The prior execution-location issue is fixed: local summons now persist `{ kind: "local" }`, normalization preserves it, and the binding overrides an SSH workspace default. I found no remaining code-review blockers.

Focused summoning/onboarding/config tests, TypeScript typecheck, and `git diff --check` pass locally. The Frontend build check is still red only on the pre-existing `familiar-analytics-view.test.ts` heal-count assertion noted in the PR.

CompleteDotTech (approved):
Re-reviewed the complete eight-file diff at `df1fffcb`, including the follow-up analytics clock fix. The explicit local runtime binding still overrides an SSH workspace default, and the hostname copy remains tied to the Cave host that executes the local harness. I found no remaining correctness blockers. Focused summoning, onboarding, config, and analytics tests pass; TypeScript typecheck and all GitHub checks are green.

PR #3291, opened 2026-07-16
Fix CodeQL trailing-slash ReDoS

CompleteDotTech:
## Summary

- replace the ambiguous trailing-slash regular expression in Omnigent URL normalization with a linear character scan
- preserve normalization behavior for valid URLs and malformed fallback values
- add regression coverage for repeated trailing slashes and adversarial non-matching slash runs

## Root cause

`normalizeOmnigentConfig` applied `/\/+$/` to the user-provided `baseUrl`. On a long run of slashes followed by a non-slash character, the JavaScript regex engine can retry the match from each slash and take polynomial time.

The replacement walks backward over trailing slashes once and performs a single slice, avoiding backtracking.

## Security alert

Resolves [CodeQL alert 125](https://github.com/OpenCoven/coven-cave/security/code-scanning/125) after the post-merge main-branch scan.

Tracked by Bead `cave-862.1` under the CodeQL-hardening work.

## Verification

- `node --experimental-strip-types src/lib/omnigent/client.test.ts` (11/11 passed)
- `node --experimental-strip-types src/lib/cave-config.test.ts`
- `corepack pnpm typecheck`
- `corepack pnpm check:tests-wired` (1,018 test files wired)
- `git diff --check`

### DIFF
--- a/src/lib/cave-config.ts
+++ b/src/lib/cave-config.ts
@@ -426,8 +426,14 @@ function normalizeStringRecord(input: unknown): Record<string, string> {
   return out;
 }
 
+function trimTrailingSlashes(value: string): string {
+  let end = value.length;
+  while (end > 0 && value[end - 1] === "/") end -= 1;
+  return value.slice(0, end);
+}
+
 export function normalizeOmnigentConfig(input: Partial<CaveOmnigentConfig> | undefined): CaveOmnigentConfig {
-  const rawUrl = typeof input?.baseUrl === "string" ? input.baseUrl.trim().replace(/\/+$/, "") : "";
+  const rawUrl = typeof input?.baseUrl === "string" ? trimTrailingSlashes(input.baseUrl.trim()) : "";
   let baseUrl = rawUrl;
   if (rawUrl) {
     try {
--- a/src/lib/cave-config.ts
+++ b/src/lib/cave-config.ts
@@ -15,7 +15,6 @@ import {
   isSshRuntime,
   normalizeFamiliarRuntime,
 } from "./familiar-runtime.ts";
-import { normalizeOmnigentBaseUrl } from "./omnigent/token.ts";
 import type { UserProfile } from "./user-profile-shared.ts";
 
 const CONFIG_PATH = path.join(caveHome(), "config.json");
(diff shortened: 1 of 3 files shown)


src/lib/cave-config.ts
```diff
@@ -426,8 +426,14 @@ function normalizeStringRecord(input: unknown): Record<string, string> {
...
 
+function trimTrailingSlashes(value: string): string {
+  let end = value.length;
+  while (end > 0 && value[end - 1] === "/") end -= 1;
+  return value.slice(0, end);
+}
+
 export function normalizeOmnigentConfig(input: Partial<CaveOmnigentConfig> | undefined): CaveOmnigentConfig {
-  const rawUrl = typeof input?.baseUrl === "string" ? input.baseUrl.trim().replace(/\/+$/, "") : "";
+  const rawUrl = typeof input?.baseUrl === "string" ? trimTrailingSlashes(input.baseUrl.trim()) : "";
```
CompleteDotTech: **[P1] Reuse the safe trimmer in the runtime normalizer**

When an Omnigent base URL contains a long slash run followed by a non-slash (the adversarial case added in the test), this function now returns it unchanged, but every actual Omnigent run passes that value through `OmnigentClient.fromBaseUrl()` and `normalizeOmnigentBaseUrl()`, which still executes `url.trim().replace(/\/+$/, "")`. A 25k-slash input takes about 330 ms there and scales quadratically, so the ReDoS is merely deferred from config load to client creation. Please reuse the linear trimmer in `normalizeOmnigentBaseUrl` (or centralize normalization) and exercise the runtime normalizer in the regression test.

CompleteDotTech (commented):
Found one remaining ReDoS path in the Omnigent runtime normalization.

src/lib/cave-config.ts
CompleteDotTech: Fixed across commits 791f15c7 and 778ee5f9. The runtime `normalizeOmnigentBaseUrl` now uses the same linear backward-scan algorithm and the regression test directly exercises it with the adversarial 25k-slash input. The config and runtime modules intentionally keep local copies of the tiny scan: importing the auth-bearing runtime module into cave-config added three files to the release sidecar and exceeded its closure budget. Focused tests, cave-config tests, typecheck, test wiring, and `git diff --check` pass.

CompleteDotTech:
Follow-up fixes pushed in `791f15c7` and `778ee5f9`: both Omnigent URL normalization paths now use a linear trailing-slash scan, and the runtime regression case is covered directly. The tiny scan remains local to each existing module so cave-config does not pull the auth module and its Node dependencies into the release sidecar closure. Local verification passed: 12/12 focused Omnigent tests, cave-config tests, typecheck, 1,018-test wiring check, and `git diff --check`. The final head `778ee5f9` also passed all 16 GitHub CI and CodeQL checks.

CompleteDotTech:
Fresh review of final head `778ee5f9`: no additional correctness issues found. The linear trailing-slash scan preserves the prior normalization behavior in both changed call paths, the adversarial regression case completes quickly, and the focused Omnigent tests (12/12), cave-config tests, `git diff --check`, and all current GitHub CI/CodeQL checks pass.

PR #3301, opened 2026-07-17
fix: prevent sidecar port desync

BunsDev:
### Motivation
- Prevent a local attacker from squatting the originally-requested loopback port and capturing token-bearing URLs when the Node sidecar falls back to an adjacent port. 
- Ensure the desktop/mobile launchers only trust the actual bound sidecar endpoint rather than a stale preselected port. 
- Preserve existing behavior for legitimate cases while removing the unsafe automatic adjacent-port bind fallback in the child process.

### Description
- Remove the server-side adjacent-port fallback in `server.ts` / `server.mjs` so the sidecar either binds the requested port or exits on bind failure. 
- Require explicit sidecar readiness by checking both TCP connectivity and the sidecar log `"> Ready on http://127.0.0.1:<port>"` line in the Tauri launcher by adding `wait_for_sidecar_ready(...)` in `src-tauri/src/lib.rs`. 
- Require the mobile runner to verify the tracked process, a listening port, and the exact ready log line by adding `server_logged_ready()` and tightening `wait_for_server()` in `scripts/mobile-tailscale.sh`. 
- Add/adjust regression assertions in `scripts/mobile-tailscale.test.mjs`, `src/middleware.test.ts`, and `src/proxy-behavior.test.ts` to cover the new readiness checks and update wording that referenced the removed port-fallback behavior.

### Testing
- `node scripts/mobile-tailscale.test.mjs` ran and passed. 
- `node --experimental-strip-types src/middleware.test.ts` and `node --experimental-strip-types src/proxy-behavior.test.ts` ran and passed. 
- `pnpm test:app`, `pnpm typecheck`, `pnpm build:server`, and `git diff --check` were executed and succeeded. 
- `bd prime && bd ready --json` was not run because the `bd` CLI is not installed in the environment, and `cd src-tauri && cargo check` was blocked by a missing system `glib-2.0.pc` (GLib development package).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4c76afdd64832795523e2a2b786fa9)

### DIFF
--- a/scripts/mobile-tailscale.sh
+++ b/scripts/mobile-tailscale.sh
@@ -363,9 +363,13 @@ ensure_tailscale() {
   fi
 }
 
+server_logged_ready() {
+  grep -F "> Ready on http://${HOST}:${PORT}" "$LOG_FILE" >/dev/null 2>&1
+}
+
 wait_for_server() {
   for _ in $(seq 1 80); do
-    if port_is_listening >/dev/null 2>&1; then
+    if recorded_server_is_running && port_is_listening >/dev/null 2>&1 && server_logged_ready; then
       return 0
     fi
     sleep 0.25
--- a/scripts/mobile-tailscale.test.mjs
+++ b/scripts/mobile-tailscale.test.mjs
@@ -120,6 +120,12 @@ test("mobile tailscale invite command is chat-safe by default", () => {
   assert.doesNotMatch(script, /Open this URL on your phone:/);
 });
 
+test("mobile tailscale readiness requires this server's ready log", () => {
+  assert.match(script, /server_logged_ready\(\)/);
+  assert.match(script, /grep -F "> Ready on http:\/\/\$\{HOST\}:\$\{PORT\}" "\$LOG_FILE"/);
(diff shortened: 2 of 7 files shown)


BunsDev (commented):
**Cody review — ✅ approve content — but this is a byte-identical subset of #3309; land exactly one**

Every hunk here also appears verbatim in #3309 (which adds the tailnet host-trust narrowing on top). Either merge this first and rebase #3309 down to its proxy delta, or close this in favor of #3309. Don't merge both.

On the content: the Tauri `wait_for_sidecar_ready` (ready-log line + TCP connect, with a negative port-squat test) and the mobile script's `recorded_server_is_running && port_is_listening && server_logged_ready` triple-check both close real trust-the-wrong-server windows.

Tradeoff to document in the body: removing `startListening`'s EADDRINUSE fallback means bare `pnpm dev` on an occupied port now exits with an error instead of sliding to the next port. Correct for sidecar/CSRF-origin integrity (the proxy pins expected origins to the configured port), and `dev-app.sh` already handles port selection — but it will surprise someone eventually; say it out loud.

PR #3306, opened 2026-07-17
Prevent familiarId 'supreme' bearer bypass and require local-human gates for grant mutations

BunsDev:
### Motivation
- Fix an authorization flaw where routes trusted a caller-supplied `familiarId` and treated the configured `supreme` ID as a bearer token, allowing remote callers to bypass project grants. 
- Prevent remote/mobile callers from creating or revoking human-managed grants or resolving proposals by posting actor-controlled IDs.

### Description
- Remove implicit all-project access based on a request-supplied Supreme familiar string by stopping special-case checks for `supreme` inside `canAccessProject`, `filterProjectsForFamiliar`, and `assertProjectAccess` so access requires explicit grants and bootstrapped grants are used to give Supreme projects access. 
- Add a local-desktop gate (`isLocalOrigin`) and `requireLocalHumanGrantMutation` checks to the project grants route so `POST`/`DELETE` require a loopback/local request. 
- Enforce local origin for grant proposals by checking `isLocalOrigin` in the grant-proposals `POST` and the proposal item `PATCH` routes. 
- Update tests to assert the new behavior: remove bearer-style Supreme expectations and add assertions that routes require local-human requests for grant mutations.

### Testing
- Ran the updated unit tests `src/lib/project-permissions.test.ts` and `src/app/api/project-grants/route.test.ts`, both passed. 
- Ran the full API test suite with `pnpm test:api` (115 test files) and all tests passed. 
- Ran static type checks with `pnpm typecheck` (TypeScript `--noEmit`) and `git diff --check`; both succeeded without errors.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4a9cfd00c0832799b8f361b627923b)

### DIFF
--- a/src/app/api/grant-proposals/[id]/route.ts
+++ b/src/app/api/grant-proposals/[id]/route.ts
@@ -1,5 +1,7 @@
 import { NextResponse } from "next/server";
 
+import { isLocalOrigin } from "@/lib/server/local-origin";
+
 import {
   ProjectAccessDeniedError,
   resolveGrantProposal,
@@ -26,6 +28,12 @@ export async function PATCH(
   req: Request,
   { params: rawParams }: { params: Promise<{ id: string }> },
 ) {
+  if (!isLocalOrigin(req)) {
+    return NextResponse.json(
+      { ok: false, error: "proposal decisions must be confirmed from the local desktop" },
+      { status: 403 },
+    );
+  }
   const params = await rawParams;
   let payload: Record<string, unknown>;
   try {
--- a/src/app/api/grant-proposals/route.ts
+++ b/src/app/api/grant-proposals/route.ts
@@ -1,5 +1,7 @@
 import { NextResponse } from "next/server";
 
+import { isLocalOrigin } from "@/lib/server/local-origin";
+
 import {
   ProjectAccessDeniedError,
   createGrantProposal,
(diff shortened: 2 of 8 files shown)


BunsDev (commented):
**Cody review — ⚠️ request changes: stale vs main + a runtime lockout risk**

Two of this PR's four concerns already landed on main via #3304 (travel-replay revalidation + `assertProjectRootAccess` fail-closed with `allowUnregisteredRoot`) — the diff still carries the old context, so a rebase is mandatory and will shrink this to its two novel changes:

1. **Removing supreme-as-implicit-bearer** (`canAccessProject`/`listAccessibleProjects`/`assertProjectAccess`) — correct: `familiarId` is client-supplied on several surfaces, so `familiarId === supremeFamiliarId` is a bearer token anyone can present.
2. **`isLocalOrigin` gates on grant/proposal mutations** — right shape, consistent with the existing local-human gates.

**Blocker before merge:** with the implicit bypass gone, supreme has access *only* via explicit grants — and `bootstrapSupremeProjectGrants` is currently called from **tests only** (I grepped: no runtime call site). Landing this as-is locks the supreme familiar out of every project until something runs the bootstrap. Wire it into startup/seeding (e.g. alongside `seedDefaultProjectsIfEmpty`) or a one-shot migration in the same PR.

Also: audit `projectId: "unregistered:<root>"` entries are a nice touch for the fail-closed path.

PR #3309, opened 2026-07-17
fix(proxy): narrow tokenless tailnet host trust

BunsDev:
### Motivation

- A tokenless mobile/Tailscale path set `COVEN_CAVE_TAILNET_TRUST=1` and was OR'd into the generic host gate, which allowed any Host header through and could expose privileged `/api/` routes without tokens. 
- The host gate must remain strict for DNS-rebinding resistance; tokenless tailnet trust should only relax the gate for narrowly validated Tailscale Serve hostnames/IPs, not arbitrary hosts.

### Description

- Change `isAllowedApiHost` to accept a third `tailnetTrusted` argument and only allow non-loopback hosts when `tailnetTrusted` and the host is a validated Tailscale Serve name or Tailscale IP fallback; mobile-access authentication remains the only generic bypass.
- Add `isTailscaleIpHost` helper and extend `isTailscaleServeHost` to accept `.ts.net` hostnames or Tailscale IPs in `src/proxy-helpers.ts`.
- Update `src/proxy.ts` to call `isAllowedApiHost(requestHost, mobileAccessAuthenticated, tailnetTrusted)` and tighten the explanatory comment for tokenless tailnet mode.
- Add/adjust regression tests in `src/proxy-behavior.test.ts` and `src/middleware.test.ts` to cover tokenless tailnet allowlist (ts.net and 100.64.0.0/10 fallback) and to assert arbitrary/suffix-smuggled hosts are rejected.

### Testing

- Ran `pnpm exec tsc --noEmit --pretty false` and it succeeded with no type errors.
- Ran `node --test src/proxy-behavior.test.ts src/middleware.test.ts scripts/mobile-tailscale.test.mjs` and all specified tests passed.
- Ran the full app test suite with `pnpm test:app` (548 test files) and it completed with all tests passing.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4c7c70221883279928a299b05bd73d)

### DIFF
--- a/scripts/mobile-tailscale.sh
+++ b/scripts/mobile-tailscale.sh
@@ -363,9 +363,13 @@ ensure_tailscale() {
   fi
 }
 
+server_logged_ready() {
+  grep -F "> Ready on http://${HOST}:${PORT}" "$LOG_FILE" >/dev/null 2>&1
+}
+
 wait_for_server() {
   for _ in $(seq 1 80); do
-    if port_is_listening >/dev/null 2>&1; then
+    if recorded_server_is_running && port_is_listening >/dev/null 2>&1 && server_logged_ready; then
       return 0
     fi
     sleep 0.25
--- a/scripts/mobile-tailscale.test.mjs
+++ b/scripts/mobile-tailscale.test.mjs
@@ -120,6 +120,12 @@ test("mobile tailscale invite command is chat-safe by default", () => {
   assert.doesNotMatch(script, /Open this URL on your phone:/);
 });
 
+test("mobile tailscale readiness requires this server's ready log", () => {
+  assert.match(script, /server_logged_ready\(\)/);
+  assert.match(script, /grep -F "> Ready on http:\/\/\$\{HOST\}:\$\{PORT\}" "\$LOG_FILE"/);
+  assert.match(script, /recorded_server_is_running && port_is_listening.*&& server_logged_ready/);
+});
+
 test("mobile tailscale runner syntax is shell-checkable by bash", () => {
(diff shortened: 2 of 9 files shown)


BunsDev (commented):
**Cody review — ⚠️ request changes: contains all of #3301 verbatim — split or supersede**

Two PRs, one payload: every hunk of #3301 (server port-fallback removal, Tauri ready-log gate, mobile ready-log wait) is byte-identical inside this PR, plus the actual "narrow tailnet host trust" delta (`isAllowedApiHost` three-arg split, `isTailscaleIpHost`). Land exactly one:
- either rebase this to *just* the proxy-helpers/proxy/middleware-test delta on top of #3301,
- or close #3301 and land this as the superset.

On the narrowing itself — good change, correct math: `100.64.0.0/10` check (`parts[1] 64..127`) ✓, `fd7a:115c:a1e0:` ULA prefix ✓, suffix-smuggling test ✓. One verify: a bracketed IPv6 `Host: [fd7a:...]:3000` — confirm `hostnameFromHost` strips brackets, else the IPv6 branch never matches (pre-existing helper, cheap test to add).

Note this stays compatible with #3310 (defense-in-depth for anyone hand-setting `COVEN_CAVE_TAILNET_TRUST`), unlike the #3324/#3317 pair.

Port-fallback removal tradeoff (inherited from #3301): plain `pnpm dev` on an occupied port now exits instead of sliding to 3001 — right call for CSRF-origin and sidecar-trust integrity, and `dev-app.sh` already picks free ports, but state it in the body so nobody "fixes" it back.

PR #3310, opened 2026-07-17
fix(mobile): require mobile access token for Tailscale app mode

BunsDev:
### Motivation
- A tokenless `pnpm mobile:tailscale:app` flow published the full loopback Next.js backend via Tailscale Serve while unsetting `COVEN_CAVE_ACCESS_TOKEN`/`COVEN_CAVE_AUTH_TOKEN`/`COVEN_CAVE_BUNDLE`, allowing remote tailnet peers to reach privileged local-only mutating APIs that start agent sessions and run workflows.
- The local-only guards relied on `Host`/`Origin` values forwarded by Serve and thus could be bypassed; the change restores an explicit credential requirement for remote-looking Tailscale hosts to prevent unauthorized remote actions.

### Description
- Require the mobile access credential for the Tailscale app path by changing the app server startup to mint/load the mobile access token and clear only sidecar state, and emit a pairing URL that carries `coven_access_token` for the native client to persist; changes live in `scripts/mobile-tailscale.sh` (app/startup/pairing URL).
- Remove the tailnet-trust bypass from the API middleware by requiring `mobileAccessAuthenticated` for the API host gate in `src/proxy.ts` so remote-looking Tailscale hosts are accepted only when the mobile access credential verifies.
- Tighten the PTY WebSocket upgrade host gate in `server.ts` so a non-loopback (Tailscale Serve) `Host` is accepted only when the mobile access token has authenticated the upgrade; retain loopback-only behavior for unauthenticated peers.
- Update client-side copy/comments and pairing semantics in the iOS code (`apps/ios/CovenCave/**`) and adjust tests that pin the prior contract (`scripts/mobile-tailscale.test.mjs`, `src/middleware.test.ts`, `src/proxy-behavior.test.ts`, `src/server-pty-ws.test.ts`, and related tests) to assert the new token-gated Tailscale app flow.

### Testing
- Ran `node scripts/mobile-tailscale.test.mjs` and the suite passed (12/12 tests OK).
- Ran middleware and behavior tests with `node src/middleware.test.ts`, `node src/server-pty-ws.test.ts`, `node src/proxy-behavior.test.ts`, and `node src/components/mobile-handoff.test.ts`; all tests passed.
- Performed a shell syntax check `bash -n scripts/mobile-tailscale.sh` and it returned no syntax errors.
- All automated checks listed above completed successfully (no failing tests).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4c7e74d6608327b0ff64527a35533a)

### DIFF
--- a/apps/ios/CovenCave/CovenCave/Networking/CaveClient.swift
+++ b/apps/ios/CovenCave/CovenCave/Networking/CaveClient.swift
@@ -2,9 +2,8 @@ import Foundation
 
 /// REST + streaming client for the Coven Cave desktop API.
 /// When the desktop is token-gated (COVEN_CAVE_ACCESS_TOKEN), every request —
-/// REST and SSE alike — carries the paired credential as a Bearer header; in
-/// tokenless tailnet-trust mode no header is sent and the tailnet boundary is
-/// the trust anchor, as before.
+/// REST and SSE alike — carries the paired credential as a Bearer header. This
+/// is required for the Tailscale app path because it exposes the full API.
 struct CaveClient {
     var connection: CaveConnection
 
@@ -111,8 +110,8 @@ struct CaveClient {
     }
 
     /// Rolling renewal: exchange the current credential for a fresh 30-day
-    /// token. Returns the new token, or nil when the desktop runs tokenless
-    /// (503) or the credential can't refresh — callers treat nil as "keep
+    /// token. Returns the new token, or nil when the desktop has no refresh
+    /// endpoint (503) or the credential can't refresh — callers treat nil as "keep
     /// using what we have".
     func refreshAccessToken() async -> String? {
         guard let req = try? request("api/mobile-token/refresh", method: "POST") else { return nil }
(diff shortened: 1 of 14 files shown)


BunsDev (commented):
**Cody review — ⚠️ my pick for the app-mode auth cluster — but Frontend build is red and it must win explicitly**

Of the three competing fixes for tokenless app mode (this, #3324 sidecar-token, #3309 keep-tokenless-narrow-host-trust), this is the strongest design: it reuses the invite-mode mobile access token end-to-end — pairing URL/QR (no hand-typed token), 30-day rolling refresh, cookie marker, existing `mobileAccessGate` — and doesn't lean on the custom-header CSRF bypass that #3317 wants to narrow. iOS client, docs, and script all move together.

Blockers/notes:
1. **Frontend build failing** — fix before anything else.
2. **Decision required**: close or rework #3324; #3309's host-gate narrowing remains compatible defense-in-depth (this PR unsets `COVEN_CAVE_TAILNET_TRUST`; mobile-access auth satisfies the host gate on its own).
3. Verify the token rides *every* request class the app makes — REST, SSE, and any WebSocket upgrade — the CaveClient diff covers REST+SSE; double-check upgrades.
4. `rm -f "$SIDECAR_TOKEN_FILE"` + `load_or_create_token` on app-mode reuse: confirm an existing *invite*-mode server on the port is still refused (the old guard also checked `$TOKEN_FILE`; the new one only checks the sidecar file — an untracked invite server would now be reused with its stale token semantics rather than erroring).

PR #3314, opened 2026-07-17
fix(project-search): prevent user-controlled globs from widening ripgrep scope

BunsDev:
### Motivation

- A user-controlled `glob` parameter was passed directly to ripgrep (`rg --glob`), which can explicitly include hidden or .gitignored files and expose secret-bearing lines in search previews.

### Description

- Remove passing user-controlled globs into ripgrep and stop delegating include globs to the `rg` process. 
- Add in-process glob filtering implemented by `globToRegExp`, `globMatchesPath`, and `filterSearchResult` that apply the glob to ripgrep's parsed, git-visible result set. 
- Update the route documentation to state the glob is applied after ripgrep returns so it cannot widen the search to hidden/ignored files. 
- Add a regression test `src/app/api/project-search-security.test.ts` asserting no `--glob` is pushed into ripgrep args and that the parsed result is filtered in-process, and wire this test into the API suite (`scripts/run-tests.mjs`).

### Testing

- Ran `pnpm test:api`, which executed the API test suite and passed (all API tests passed). 
- Ran `node --experimental-strip-types src/app/api/project-search-security.test.ts`, which passed and validated the new security assertions. 
- Ran `pnpm check:tests-wired`, which passed after wiring the new test into the API suite. 
- Ran `pnpm typecheck`, which failed due to unrelated existing `@milkdown/crepe` type/module resolution and implicit-any issues in `src/components/md-editor/md-editor-visual.tsx`, not caused by this change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4c9e662bf483278629be9c0d524932)

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -949,6 +949,7 @@ export const SUITES = {
     "src/app/api/access-groups/route.test.ts",
     "src/app/api/project-permission-routes.test.ts",
     "src/app/api/project-local-human-read.test.ts",
+    "src/app/api/project-search-security.test.ts",
     "src/app/api/project-file/route.test.ts",
     "src/app/api/salem/pathfinder/feedback/route.test.ts",
     "src/app/api/feedback/message/route.test.ts",
--- a/src/app/api/project-search-security.test.ts
+++ b/src/app/api/project-search-security.test.ts
@@ -1,0 +1,34 @@
+// @ts-nocheck
+import assert from "node:assert/strict";
+import { readFile } from "node:fs/promises";
+
+const route = await readFile(new URL("./project/search/route.ts", import.meta.url), "utf8");
+
+// Every glob handed to ripgrep must be a hardcoded exclusion literal. A
+// user-controlled or include glob can widen the search into hidden/ignored
+// files and expose secret-bearing lines in previews.
+const globPushes = [...route.matchAll(/"--glob",\s*([^),]+)/g)];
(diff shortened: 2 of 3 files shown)


BunsDev (commented):
**Cody review — ⚠️ request changes (premise correct, two functional gaps + test collision)**

The premise is real: rg include globs act as whitelist overrides that can surface hidden/ignored files, so user globs must not reach rg. Post-filtering is the right architecture. But:

1. **Brace globs silently break.** `globToRegExp` escapes `{}` literally, so `*.{ts,tsx}` — a pattern rg supported yesterday — now matches nothing, with no error. Either implement brace alternation or reject unsupported globs with a 400 so users aren't confused by empty results.
2. **Filtering after truncation loses results.** `parseRipgrepJson` applies `maxMatches`/`maxPerFile` caps *before* your filter runs, so a narrow glob in a large repo can return near-empty results because the cap was consumed by files the glob then discards. Apply the glob inside the parse loop (pass a path predicate into `parseRipgrepJson`) or raise the pre-filter cap.
3. **Mutually exclusive with #3316**: your new test asserts the route contains no `args.push("--glob"` at all, which bans #3316's static `.env` *exclusion* globs. Narrow the assertion to user-supplied globs and keep static exclusions legal. (See my note on #3316 for the combined shape.)
4. CodeQL shows a 3s config-style failure — likely the aggregate-check branch-protection gotcha, verify/rerun.

PR #3315, opened 2026-07-17
security(changes): disable repo `core.fsmonitor` during automatic status polling

BunsDev:
### Motivation
- Prevent automatic client polling from triggering repository-configured `core.fsmonitor` commands when the UI queries the changes endpoint. 
- The changes-list path previously invoked `git status` under repo configuration, which can execute arbitrary commands via `core.fsmonitor` in a malicious working tree. 
- Harden the server-side status call so the diff-first client polling added earlier cannot lead to local command execution as the app user.

### Description
- Add a `gitStatus` helper that runs `git -c core.fsmonitor=false status ...` to explicitly disable repository-configured `core.fsmonitor` hooks during status calls. 
- Replace the direct `git(..., ["status", ...])` call in `listChanges` with the new `gitStatus` helper so the change-list path uses the hardened invocation. 
- Add assertions to `src/app/api/changes/route.test.ts` that validate the presence of `gitStatus` and that the change-list polling uses it, providing source-level regression coverage.

### Testing
- Ran `node src/app/api/changes/route.test.ts` and it succeeded. 
- Ran `node src/components/debug-pane.test.ts` and it succeeded. 
- Verified via a local repro that `git -c core.fsmonitor=false status --porcelain=...` does not invoke a repo-local `core.fsmonitor` script (manual shell check succeeded). 
- `pnpm typecheck` currently fails due to pre-existing unrelated TypeScript issues in `src/components/md-editor/md-editor-visual.tsx`, so full typecheck did not pass in this environment.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4c99f5d4f08327891c908b8139ce05)

### DIFF
--- a/src/app/api/changes/route.test.ts
+++ b/src/app/api/changes/route.test.ts
@@ -16,6 +16,18 @@ assert.doesNotMatch(
   "changes route should use gitDiff for every git diff invocation",
 );
 
+assert.match(
+  source,
+  /function gitStatus[\s\S]*\["-c", "core\.fsmonitor=false", "status", \.\.\.args\]/,
+  "git status calls must disable repository-configured fsmonitor commands",
+);
+
+assert.match(
+  source,
+  /gitStatus\(repoRoot, \["--porcelain=v1", "-z", "--untracked-files=all"\]\)/,
+  "change-list status polling must use the hardened gitStatus helper",
+);
+
 
 // The status GET carries the current branch (Projects hub Git section) — from
 // the existing currentBranch() helper, omitted on unborn repos.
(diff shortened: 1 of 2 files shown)


BunsDev (commented):
**Cody review — ✅ approve, one gap + one coordination note**

Neutralizing repo-configured `core.fsmonitor` during automatic status polling closes a real "open a hostile repo → its config executes a command on poll" hole, and the source-level test pins both the helper and the call site.

Gap for a follow-up: `git diff` also refreshes the index and can invoke fsmonitor. `gitDiff()` already passes `--no-ext-diff --no-textconv`; consider adding `-c core.fsmonitor=false` there too (or in the shared `git()` wrapper) so the diff path can't trigger it either.

Coordination: #3323 adds a **raw** `git status` call (`changedFilePaths`) that your test regex won't catch — whichever PR lands second must route it through `gitStatus()`, or this hardening is bypassed on the diff/revert auth path.

PR #3316, opened 2026-07-17
fix(search): redact .env-family files from project search

BunsDev:
### Motivation

- Project search began returning ripgrep `context` lines as `before`/`after` fields and also accepted caller-controlled `--glob`, allowing a client to include hidden `.env*` files and leak adjacent secret lines. 
- The repository already redacts `.env` contents in the project-file API, so search should maintain the same protection boundary to avoid inconsistent information disclosure. 
- The change aims to close this information-leak by preventing `.env`-family files from being searched or parsed into returned context fields.

### Description

- Append ripgrep exclusion globs to the search invocation so caller-provided globs cannot re-include `.env`-family files by adding `--glob '!.env*'` and `--glob '!**/.env*'` to the args in `buildRgArgs` in `src/app/api/project/search/route.ts`.
- Add a parser-level defense-in-depth in `src/lib/project-search.ts` by introducing `isEnvFamilyPath` and skipping any `match` or `context` events whose path contains a segment starting with `.env` so no preview or `before`/`after` is produced for these files.
- Add a regression test in `src/lib/project-search.test.ts` that simulates `.env` and nested `.env.local` events and verifies they are dropped while non-`.env` files still preserve context.

### Testing

- Ran the parser unit tests with `node --experimental-strip-types src/lib/project-search.test.ts`, which completed successfully with all assertions passing.
- Ran TypeScript check with `pnpm typecheck`, which returned without errors.
- Verified test wiring with `pnpm check:tests-wired`, which reported all tests wired and passed.
- Ran `git diff --check` (whitespace/checks) which showed no issues.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4c95ec5f3483278a8851b3776c859f)

### DIFF
--- a/src/app/api/project/search/route.ts
+++ b/src/app/api/project/search/route.ts
@@ -133,6 +133,9 @@ function buildRgArgs(params: {
   else if (params.caseMode === "sensitive") args.push("--case-sensitive");
   else args.push("--smart-case");
   if (params.glob) args.push("--glob", params.glob);
+  // Match /api/project-file's .env-family redaction boundary even when a
+  // caller-supplied glob explicitly includes hidden environment files.
+  args.push("--glob", "!.env*", "--glob", "!**/.env*");
   // Pattern then an explicit search path. The path is REQUIRED: with no path
   // argument and a non-TTY stdin (which execFile gives the child), ripgrep
   // blocks reading stdin instead of walking the directory — so it would hang
--- a/src/lib/project-search.test.ts
+++ b/src/lib/project-search.test.ts
@@ -150,4 +150,21 @@ function contextLine(path, lineNumber, text) {
   assert.deepEqual(result.files, []);
 }
 
+// ── .env-family files are redacted from project search results ──────────────
+{
+  const stdout = [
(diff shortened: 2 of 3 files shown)


BunsDev (commented):
**Cody review — ✅ approve content — but must be reconciled with #3314 before either lands**

Defense in depth is right: exclusion globs at the rg layer + `isEnvFamilyPath` at the parser layer (with a test proving parser-level redaction). `.env.example` gets caught too — over-redaction in the safe direction, fine.

**Hard conflict with #3314:** that PR removes *all* `args.push("--glob", ...)` from the route and adds a test asserting the route contains **no** `args.push("--glob"` at all — which this PR's exclusion lines would violate (and vice versa: this PR keeps rg-side globbing #3314's test bans). The clean combined outcome:
- drop **user-supplied include** globs from rg (that's #3314's real fix),
- keep **static exclusion** globs + this parser filter,
- fix #3314's assertion to target the user-glob line specifically, not every `--glob`.

Minor: `"!.env*"` already matches basenames at any depth in rg glob semantics, so `"!**/.env*"` is redundant (harmless).

PR #3317, opened 2026-07-17
fix(proxy): limit header CSRF bypass to mobile APIs

BunsDev:
### Motivation
- A global `x-coven-cave-token` CSRF/source bypass allowed authenticated remote/mobile requests (over Tailscale Serve) to reach local-only endpoints that rely only on a forwarded loopback `Host`, creating a high-severity authorization/origin-gate bypass. 
- The fix scopes the header-based relaxation to only explicitly mobile-capable API surfaces so local-only automation/inbox routes continue to fail closed for remote origins.

### Description
- Add an allowlist `HEADER_CSRF_TRUSTED_API_PATHS` and helper `isHeaderCsrfTrustedApiPath` in `src/proxy.ts` to enumerate mobile-capable endpoints. 
- Change the header-CSRF-trust predicate so `headerCsrfTrusted` requires the sidecar header _and_ that the request path is allowlisted. 
- Update the proxy comments to document the scoped trust rationale for Tailscale Serve and local-only routes. 
- Strengthen `src/middleware.test.ts` to assert the presence of the allowlist and to ensure local-only endpoints (e.g., Codex automations and inbox) are not included in it.

### Testing
- Ran `node src/middleware.test.ts`, which passed. 
- Ran `node src/proxy-behavior.test.ts`, which passed. 
- Ran `git diff --check`, which reported no whitespace/format issues. 
- Ran `pnpm typecheck`, which failed due to pre-existing unrelated type/module issues in `src/components/md-editor/md-editor-visual.tsx` (missing `@milkdown/crepe` types and implicit `any` parameters); these are unrelated to the proxy change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4cb17183008327880284f225c11295)

### DIFF
--- a/src/middleware.test.ts
+++ b/src/middleware.test.ts
@@ -78,8 +78,8 @@ assert.match(
 // (cookies auto-send cross-origin → CSRF).
 assert.match(
   source,
-  /const headerCsrfTrusted =\s*Boolean\(sidecarToken\) && req\.headers\.get\(TOKEN_HEADER\) === sidecarToken/,
-  "origin/referer gate bypass must be keyed to the custom sidecar header token",
+  /const headerCsrfTrusted =\s*Boolean\(sidecarToken\) &&\s*req\.headers\.get\(TOKEN_HEADER\) === sidecarToken &&\s*isHeaderCsrfTrustedApiPath\(req\.nextUrl\.pathname\)/,
+  "origin/referer gate bypass must be keyed to the custom sidecar header token and mobile endpoint allowlist",
 );
 assert.match(
   source,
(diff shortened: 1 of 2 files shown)


BunsDev (commented):
**Cody review — ⚠️ hold: likely mobile regression + collides with #3324**

The custom `x-coven-cave-token` header can't be attached by a browser cross-origin without a CORS preflight, so the existing bypass is already CSRF-sound. What this narrowing buys is limited — and what it risks is concrete:

1. **Re-breaking mobile-over-Serve mutating requests.** The bypass exists because a WKWebView request over Tailscale Serve carries `Origin: https://<host>.ts.net` while the origin gate expects the loopback/Host-derived origin (#618 fixed, #716 regressed, this comment block documents it). Narrowing to `/api/mobile-handoff` + `/api/mobile-token/refresh` means every *other* token-authenticated mutating API from that path 403s "forbidden origin" again. Val verified the physical-iPhone flow this week — this needs an explicit test against that path before it lands.
2. **#3324 depends on the broad bypass**: it has the native iOS app authenticate all `/api/` calls with exactly this header. These two PRs cannot both land as-is.
3. The cited threat (automation/inbox local-only routes) is already partially handled upstream — `isLocalOnlyAutomationRun` hard-rejects mobile/tailnet ingress *before* the CSRF gate, regardless of tokens.

If the goal is de-risking token theft, consider inverting: keep the header bypass broad, and grow the *local-only route* denylist (the mechanism that already exists) for routes that must never be reachable off-loopback.

PR #3318, opened 2026-07-17
fix(release): tie manual release checkouts to tag

BunsDev:
### Motivation
- Prevent manual `workflow_dispatch` runners from building, signing, or notarizing artifacts from an arbitrary ref while attaching them to an existing release tag, which could let an operator-p with dispatch permission publish malicious signed artifacts to an official release.
- Keep build and release-helper script provenance tied to the reviewed/tagged source whose release metadata and uploads are being modified.

### Description
- Remove the `source_ref` `workflow_dispatch` input from `.github/workflows/release.yml` so operators cannot supply an alternate build ref when dispatching the release workflow.
- Change `actions/checkout` in the `build` and `checksums` jobs to use `ref: ${{ github.event.inputs.tag || github.ref }}` so checkouts are always the tag/ref receiving the uploads.
- Update `scripts/release-macos-signing.test.mjs` to assert `source_ref` is no longer present and to verify checkouts are tag-bound.
- Commit message used: `fix(release): tie manual release checkouts to tag`.

### Testing
- Ran `node --test scripts/release-macos-signing.test.mjs` and all tests passed (`9 passed`, `0 failed`).
- Parsed the modified workflow with a YAML parser via `node` to ensure the workflow is valid and the YAML loads successfully (`release.yml parsed`).
- Ran `git diff --check` to validate there are no whitespace/patch errors and it reported clean results.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4cad34732c83278dc137721e360354)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -47,10 +47,6 @@ on:
         description: "Tag to attach builds to (e.g. v0.0.16). Must already exist on origin."
         required: true
         type: string
-      source_ref:
-        description: "Git ref to build from for manual release-infra retries. Defaults to tag."
-        required: false
-        type: string
       platform:
         description: "Platform to build when manually re-running a release."
         required: false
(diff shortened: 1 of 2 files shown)


BunsDev (commented):
**Cody review — ✅ approve**

Removing `source_ref` pins manual release retries to the exact tag whose release receives the artifacts — kills the "sign and publish assets built from an unrelated ref" provenance hole. Test now asserts the input is gone *and* the checkout expression is tag-only.

Known tradeoff (worth stating in the PR body if not already): release-infra fixes can no longer be retried against an old tag without moving the tag or cutting a new one. That's the point, but ops should know the escape hatch is intentionally gone.

PR #3319, opened 2026-07-17
Fail closed for missing chat project roots

BunsDev:
### Motivation
- The chat endpoint previously fell back to the user's home directory when `projectRoot` was missing or invalid, which allowed spawned familiars to run from `$HOME` and access sensitive user files.
- The intent is to minimize native agent filesystem scope by refusing uncontrolled homedir fallbacks and preferring an explicit familiar workspace when available.

### Description
- Change `resolveLocalRuntimeCwd` to fail when `projectRoot` is absent by adding a new `project_root_required` `RuntimeScopeError` instead of returning `homedir()` in `src/lib/chat-runtime-scope.ts`.
- Use the resolved familiar workspace as an explicit fallback for new local chats (`projectRoot ?? resumeCwd ?? resolvedFamiliarWorkspace`) so agents are not started in `$HOME` in `src/app/api/chat/send/route.ts`.
- Emit structured runtime-scope errors into the OpenClaw chat stream and return JSON errors instead of spawning a bridge when a scoped cwd cannot be resolved, preventing any child process from starting with an unsafe cwd in `src/app/api/chat/send/route.ts`.
- Update tests to assert the fail-closed behavior and to expect the familiar-workspace fallback in routing assertions (`src/lib/chat-runtime-scope.test.ts` and `src/app/api/chat/send/harness-routing.test.ts`).

### Testing
- Ran `node --experimental-strip-types src/lib/chat-runtime-scope.test.ts` and the test passed.
- Ran `node --experimental-strip-types src/app/api/chat/send/harness-routing.test.ts` and the harness-routing tests passed.
- Ran `pnpm typecheck` and `git diff --check`, both of which completed successfully.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4d2a89c3c88327b9a67a5f3cf1ae3f)

### DIFF
--- a/src/app/api/chat/send/harness-routing.test.ts
+++ b/src/app/api/chat/send/harness-routing.test.ts
@@ -368,7 +368,7 @@ assert.match(
 
 assert.match(
   chatRoute,
-  /await resolveLocalRuntimeCwd\(body\.projectRoot \?\? resumeCwd\)/,
+  /await resolveLocalRuntimeCwd\(body\.projectRoot \?\? resumeCwd \?\? resolvedFamiliarWorkspace\)/,
   "Local Cave chat must fail closed on invalid project roots instead of downgrading to homedir",
 );
 
--- a/src/app/api/chat/send/route.ts
+++ b/src/app/api/chat/send/route.ts
@@ -733,9 +733,25 @@ function openClawChatResponse(args: {
       pushProgress("openclaw-resolve", "OpenClaw agent resolved", "done", `${agentId} (${agentBinding.source})`);
       const argv = openClawAgentArgs(args.harnessPrompt, agentId, conversationId);
       const spawnArgv = openClawSpawnArgs(argv);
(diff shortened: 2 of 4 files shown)


BunsDev (commented):
**Cody review — ✅ approve**

Refusing homedir-scoped fallback sessions (`project_root_required`) is the right fail-closed default, and the fallback chain `projectRoot ?? resumeCwd ?? resolvedFamiliarWorkspace` preserves the legitimate flows (resumed conversations, familiar-workspace chats). Moving `resolvedFamiliarWorkspace` computation ahead of cwd resolution is correct for that.

Also good: the OpenClaw branch now catches `RuntimeScopeError` and emits a structured SSE error + done instead of an unhandled throw mid-stream.

Verified the remaining failure mode: familiar with no workspace + no project + fresh session → 4xx with a clear message rather than a silent `~` session. That's the intent. 👍

PR #3320, opened 2026-07-17
fix: fail closed for unsafe OpenClaw shim prompts

BunsDev:
### Motivation
- A Windows-specific change allowed spawning OpenClaw with `shell:true` while passing attacker-controlled chat prompts as argv, which enables shell command injection through `.cmd` npm shims.
- The intent is to prevent prompt-driven command injection while preserving OpenClaw functionality on platforms and binaries that accept argv safely.

### Description
- Change `openClawNeedsShell()` to require both `process.platform === "win32"` and that the resolved `openclaw` binary ends with `.cmd`, and add `openClawSupportsUntrustedArgs()` to signal whether untrusted argv are safe to pass. (`src/lib/openclaw-bin.ts`)
- Thread the resolved OpenClaw command through spawn helpers by adding an optional `bin` parameter to `openClawSpawnArgs()` and using the resolved command when spawning. (`src/lib/openclaw-bin.ts`)
- Update the OpenClaw chat startup to resolve the command once, check `openClawSupportsUntrustedArgs()`, fail closed with a clear `openclaw_unsafe_shell` error stream if the resolved shim requires shell parsing, and only spawn when safe. (`src/app/api/chat/send/route.ts`)
- Add/adjust regression assertions to cover the new guard and the threaded `bin` usage in the OpenClaw unit tests and routing tests. (`src/lib/openclaw-bin.test.ts`, `src/app/api/chat/send/harness-routing.test.ts`)

### Testing
- Ran `node src/lib/openclaw-bin.test.ts`, which passed successfully. 
- Ran `node src/app/api/chat/send/harness-routing.test.ts`, which passed successfully. 
- Ran `pnpm typecheck` which failed due to an unrelated missing dependency/type (`@milkdown/crepe`) in the environment and not because of these changes.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4c9f5308ec8327959efe1f8440a63b)

### DIFF
--- a/src/app/api/chat/send/harness-routing.test.ts
+++ b/src/app/api/chat/send/harness-routing.test.ts
@@ -158,8 +158,13 @@ assert.match(
 
 assert.match(
   chatRoute,
-  /import \{ openClawBin, openClawNeedsShell, openClawSpawnArgs, openClawSpawnEnv \} from "@\/lib\/openclaw-bin";/,
+  /import \{ openClawBin, openClawNeedsShell, openClawSpawnArgs, openClawSpawnEnv, openClawSupportsUntrustedArgs \} from "@\/lib\/openclaw-bin";/,
   "OpenClaw native chat should use the Windows-aware binary resolver instead of spawning a bare command",
+);
+assert.match(
+  chatRoute,
+  /if \(!openClawSupportsUntrustedArgs\(openclawCommand\)\)[\s\S]*openclaw_unsafe_shell/,
+  "OpenClaw chat should fail closed before passing untrusted prompts to shell-only Windows shims",
 );
 
 assert.match(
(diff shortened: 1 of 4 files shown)


BunsDev (commented):
**Cody review — ✅ approve**

Failing closed on Windows `.cmd` shims instead of trying to out-quote `cmd.exe` is the right call — cmd metacharacter quoting is a losing game with untrusted prompt text. The error is user-actionable (`OPENCLAW_BIN` remediation) and the SSE stream terminates cleanly.

One gap worth a follow-up: `openClawNeedsShell` only matches `.cmd`. A `.bat` resolution has identical semantics (and identical injection exposure) — extend the suffix check to `.bat` too so it can't sneak through the `openClawSupportsUntrustedArgs` gate.

PR #3326, opened 2026-07-17
fix(chat): suppress startup blocks before exec echo parsing

BunsDev:
### Motivation
- A newly added exec-echo state machine could parse lines from hidden startup blocks before those blocks were suppressed, allowing attacker-controlled startup content to be emitted through exec bailout/resume paths and reach trusted HTML rendering.
- Prevent leaking content from suppressed startup tags (e.g., `<INSTRUCTIONS>`) into assistant output and persisted conversation history.

### Description
- Move startup-block suppression to run before the exec-echo parser in `AssistantFilter` so suppressed startup lines are always dropped prior to any exec state transitions.
- Add two regression tests to `src/lib/chat-assistant-filter.test.ts` that assert `exec` header-bailout and exec-output-resume attempts inside a suppressed `<INSTRUCTIONS>` block do not leak visible assistant text.
- Modified files: `src/lib/chat-assistant-filter.ts` and `src/lib/chat-assistant-filter.test.ts`.

### Testing
- Ran the targeted test file with `node --experimental-strip-types src/lib/chat-assistant-filter.test.ts` and the new assertions passed.
- Ran the full app test suite with `pnpm test:app` (all test files passed) and `pnpm typecheck` which completed without errors.
- Ran `git diff --check` to validate formatting and whitespace issues and found no problems.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4d0905bcb88327be85d6df07ead487)

### DIFF
--- a/src/lib/chat-assistant-filter.test.ts
+++ b/src/lib/chat-assistant-filter.test.ts
@@ -39,6 +39,36 @@ assert.equal(
   "startup prompt blocks after the assistant marker should stay hidden",
 );
 
+
+assert.equal(
+  feed([
+    "codex",
+    "<INSTRUCTIONS>",
+    "exec",
+    "[click me](javascript:globalThis.__COVEN_POC=1)",
+    "</INSTRUCTIONS>",
+    "Visible reply",
+  ]),
+  "Visible reply\n",
+  "exec header bailout inside startup blocks should not leak hidden content",
+);
+
+assert.equal(
+  feed([
+    "codex",
+    "<INSTRUCTIONS>",
+    "exec",
+    "/bin/zsh -lc 'cat AGENTS.md' in /repo",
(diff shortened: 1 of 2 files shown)


BunsDev (commented):
**Cody review — ✅ approve**

Hoisting startup-block suppression above the exec-echo state machine is the correct ordering: previously an attacker-controlled startup block could enter the exec parser and re-emerge as visible prose via its bailout paths. Both leak reproductions are pinned as tests.

Edge worth knowing (acceptable, pre-existing shape): a *legitimate* exec block whose output contains a bare `<INSTRUCTIONS>` line will now open startup suppression mid-stream and swallow output until a matching close tag. That hazard existed before for non-exec lines; the fail direction (hide too much rather than leak) is right for this filter.

PR #3327, opened 2026-07-17
Validate GitHub username and harden .env.local upserts to prevent injection

BunsDev:
### Motivation
- The username-only GitHub setup flow allowed unvalidated `username.trim()` values to be written into `.env.local`, which enabled newline-based env injection (e.g. `alice\nWORKSPACE_ROOT=/`) and expansion of trusted roots. 
- Prevent persistent environment-variable injection while preserving the existing username-only convenience for valid GitHub usernames.

### Description
- Add GitHub username syntax validation via `isValidGitHubUsername` and reject invalid usernames in `src/app/api/github/pat/route.ts` before persisting them. 
- Add `assertEnvAssignmentSafe` in `src/lib/env-file.ts` that rejects invalid env keys and values containing newlines, and invoke it for both in-place replacements and appended keys. 
- Add a regression assertion in `src/lib/env-file.test.ts` that verifies newline-containing values (e.g. `alice\nWORKSPACE_ROOT=/`) are rejected instead of being serialized into `.env.local`.

### Testing
- Ran TypeScript checks with `pnpm exec tsc --noEmit --pretty false` which succeeded. 
- Ran the env-file unit assertions with `node --experimental-strip-types src/lib/env-file.test.ts` which printed success. 
- `bd` (Beads) CLI could not be executed in this environment, so `bd prime`/`bd ready` were not run here.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4d0e2fc76c8327b558c7afd040079f)

### DIFF
--- a/src/app/api/github/pat/route.ts
+++ b/src/app/api/github/pat/route.ts
@@ -24,6 +24,16 @@ export const runtime = "nodejs";
 
 const PAT_KEY = "GITHUB_PAT";
 const LOGIN_KEY = "GITHUB_USERNAME";
+const GITHUB_USERNAME_PATTERN = /^[A-Za-z0-9-]{1,39}$/;
+
+function isValidGitHubUsername(username: string): boolean {
+  return (
+    GITHUB_USERNAME_PATTERN.test(username) &&
+    !username.startsWith("-") &&
+    !username.endsWith("-") &&
+    !username.includes("--")
+  );
+}
 
 /** Apply key updates to .env.local in place. `null` deletes a key. Comments,
  *  blank lines, key ordering, and unrelated values are preserved (the old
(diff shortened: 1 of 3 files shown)


BunsDev (commented):
**Cody review — ✅ approve**

Username regex matches GitHub's actual rules (1–39, alnum/hyphen, no leading/trailing/double hyphen), and `assertEnvAssignmentSafe` closes the `KEY=value\nINJECTED=...` line-injection hole in `.env.local` upserts with a test.

One nit: a PAT containing `\r`/`\n` (e.g. a mangled paste) will now make `upsertEnvContent` **throw** from inside the POST handler — check that surfaces as a 400 rather than an unhandled 500. A cheap `if (/[\r\n]/.test(pat)) return 400` in the route would make the failure mode friendly.

PR #3328, opened 2026-07-17
Avoid exposing Apple notarization password

BunsDev:
### Motivation

- The Apple app-specific password was previously passed to `xcrun notarytool` via `--password`, which exposes it in process arguments and risks local credential disclosure on CI or developer machines.

### Description

- Introduce `NOTARY_KEYCHAIN_PROFILE` and `NOTARY_KEYCHAIN_PATH` and add `setup_notary_keychain_profile()` to create a temporary keychain profile and feed the app-specific password to `xcrun notarytool store-credentials` on stdin so the password never appears in argv. 
- Update `run_notary_submit()` and `print_notary_log()` to use `--keychain-profile` and `--keychain` for Apple ID auth instead of `--apple-id/--password/--team-id`. 
- Add `cleanup_release_artifacts()` and an `EXIT` trap to remove the temporary keychain and DMG staging directory after the release run. 
- Add a regression test in `scripts/release-macos-signing.test.mjs` to assert the new keychain-based flow is present and that `--password` is no longer used in the Apple ID notarization code paths.

### Testing

- Ran a shell syntax check with `bash -n scripts/release.sh`, which succeeded.
- Ran the macOS release signing tests with `node scripts/release-macos-signing.test.mjs`, and all tests passed.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a4d1504c1cc83279dc0e2da71ef88bd)

### DIFF
--- a/scripts/release-macos-signing.test.mjs
+++ b/scripts/release-macos-signing.test.mjs
@@ -44,6 +44,10 @@ test("Apple ID notarization avoids putting the app password in process arguments
   assert.match(releaseScript, /security delete-keychain "\$NOTARY_KEYCHAIN_PATH"/);
   assert.match(releaseScript, /rm -rf "\$NOTARY_KEYCHAIN_DIR"/);
 
+  const cleanupFunction = releaseScript.slice(
+    releaseScript.indexOf("cleanup_release_artifacts()"),
+    releaseScript.indexOf("trap cleanup_release_artifacts EXIT"),
+  );
   const setupFunctions = releaseScript.slice(
     releaseScript.indexOf("store_notary_credentials()"),
     releaseScript.indexOf("print_notary_log()"),
@@ -57,6 +61,15 @@ test("Apple ID notarization avoids putting the app password in process arguments
     releaseScript.indexOf("cleanup_dmg_artifacts()"),
   );
 
+  assert.match(cleanupFunction, /local exit_status=\$\?/);
+  assert.match(cleanupFunction, /rm -rf "\$NOTARY_KEYCHAIN_DIR" \|\| true/);
+  assert.match(cleanupFunction, /rm -rf "\$DMG_STAGE" \|\| true/);
+  assert.match(cleanupFunction, /return "\$exit_status"/);
+  assert(
+    cleanupFunction.indexOf('security delete-keychain "$NOTARY_KEYCHAIN_PATH"') <
+      cleanupFunction.indexOf('rm -rf "$DMG_STAGE"'),
+    "credential cleanup must run before best-effort DMG cleanup",
+  );
   assert.doesNotMatch(setupFunctions, /--password/);
   assert.doesNotMatch(logFunction, /--password "\$NOTARY_APPLE_PASSWORD"/);
   assert.doesNotMatch(submitFunction, /--password "\$NOTARY_APPLE_PASSWORD"/);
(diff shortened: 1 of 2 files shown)


BunsDev (commented):
**Cody review — ✅ approve after rebase** (branch is CONFLICTING with main)

Moving the Apple app-specific password off argv (visible in `ps`) into a temp keychain profile fed via stdin is the right fix, and the EXIT trap consolidation (DMG stage + keychain) is correct.

Findings (minor):
- `NOTARY_KEYCHAIN_PATH="$(mktemp -t covencave-notary-keychain).keychain-db"` — `mktemp` *creates* the suffix-less file, then you only ever use/delete the `.keychain-db` sibling. The empty mktemp file is left behind every run. Capture the mktemp path and `rm` it (or use it directly as the keychain path).
- `security set-keychain-settings -lut 21600` + delete-on-exit is good hygiene. 👍

Test coverage pinning `--password` out of setup/log/submit sections is exactly the right regression guard.

BunsDev (commented):
**Cody follow-up — runtime verification done: the stdin pipe WORKS. Blocker withdrawn.**

Ran the PR's exact mechanism locally (fake credentials, real toolchain):

```
printf 'fake-app-password\n' | xcrun notarytool store-credentials probe-cody-stdin \
  --apple-id probe@example.com --team-id FAKETEAM12
→ App-specific password for probe@example.com:   ← prompt consumed the piped value
→ Validating your credentials...
→ Error: HTTP status code: 401. Invalid credentials. (exit 1)
```

No /dev/tty requirement, no hang: `notarytool` read the password from the pipe and went all the way to Apple's credential-validation call, failing only because the password was fake. With real credentials this exits 0 and stores the profile. My earlier concern (prompt bound to /dev/tty) does not apply on this toolchain.

**Updated verdict: approve after rebase**, with three asks folded in while resolving the conflict:

1. **Wrap `store-credentials` in the retry helper.** Verified above that it makes a network call to Apple at setup time — that's a new transient-failure point under `set -euo pipefail`, exactly the class main's new `retry()` (cave-1hha) exists for. After rebasing: `retry 3 10` around the store-credentials invocation. Note it prompts-then-reads-stdin per attempt, so the `printf | retry ...` composition needs the pipe *inside* the retried command (e.g. a small wrapper function) or each retry re-reads an empty stdin.
2. **Keep the DMG trap removal when rebasing.** Bash EXIT traps are last-writer-wins — if the conflict resolution reintroduces `trap 'rm -rf "$DMG_STAGE"' EXIT`, it silently replaces `cleanup_release_artifacts` and the credential-bearing keychain outlives the run.
3. **mktemp orphan (nit):** `mktemp -t ...` creates the suffix-less file; only the `.keychain-db` sibling is cleaned up. Use `mktemp -d` + a fixed filename inside.

Caveat for the record: verified on this Mac's Xcode toolchain; CI pins `Xcode_16.4` on macos-15. Since the behavior is an HTTP validation rather than a TTY quirk, it should hold there — the first tag cut after merge is the definitive confirmation.

BunsDev:
Rebased onto current main and applied the review fixes in the remote branch (head `2648232d`):\n\n- credential validation now retries through `store_notary_credentials`, which replenishes stdin on every attempt without placing the password in argv\n- temporary keychain state now lives under a `mktemp -d` directory, eliminating the orphan file\n- the consolidated EXIT handler deletes credential state before best-effort DMG cleanup, isolates cleanup failures under `set -e`, and preserves the original exit status\n- regression assertions cover retry wiring, temp-directory cleanup, cleanup ordering, and exit-status preservation\n\nLocal verification: `bash -n scripts/release.sh`; `node scripts/release-macos-signing.test.mjs` (11/11); `git diff --check`. Independent final review found no significant issues. Required CI is running.

PR #3340, opened 2026-07-17
feat(familiars): terminal-futurist roster cards in the profile-card language

BunsDev:
## What

Propagates the profile share card's terminal-futurist aesthetic (the `pfc-*` card the operator screenshotted as the target vibe) to the familiars roster and its surface chrome — bead **cave-uw7c**.

### Roster cards → terminal stat tiles
- Monospace lowercase micro-labels (the share card's label voice)
- Hairline-divided tri-stat band: **sessions / this week / memories** with glowing numerals (`--glow-numeral`, zeroed under `data-mode="light"`)
- **14-day activity strip** per familiar, bucketed 1..4 against the window max exactly like the pfc heatmap (`levelFor` semantics), accent-presence level ramp
- Framed square avatar tiles, squared status chips, pulsing presence dot (static under `prefers-reduced-motion`)
- Footer unchanged in behavior: memory one-liner + Profile → / Analytics → links

### Surface chrome
- `familiars` wordmark with the share card's checkerboard treatment + mono tagline
- Detail-panel nameplate takes the halftone strip; detail rail avatars squared

### Data
- `buildFamiliarCardStats` grows `sessionsTotal` + `activity[]` (per-UTC-day counts, `ACTIVITY_DAYS` window; same UTC bucketing as `session-pulse`), unit-tested; `emptyStats` callers (cockpit, analytics data, growth view) updated

## Validation

- `pnpm test:app` — **750/750 pass** (incl. updated source-pin tests + design-token-drift gate)
- `pnpm typecheck` — clean
- New CSS fully tokenized (no ratchet regressions; off-scale literals renormalized to the type/space scale)
- Playwright screenshots of roster (dark + forced light) and detail panel against the built app with the live roster — light mode drops the glow and tracks theme tokens

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ breaking config changes; patch releases stay additive.
 ## [Unreleased]
 
 ### Changed
+- **Familiars roster speaks the profile-card language** — roster cards restyled as terminal stat tiles matching the profile share card: monospace lowercase labels, hairline-divided sessions / this-week / memories band with glowing numerals, a 14-day activity strip bucketed like the profile heatmap, framed square avatars, and the checkerboard `familiars` wordmark on the surface header (cave-uw7c).
 - **Omnigent fleet is now per-user vault-gated** — Fleet surfaces (board Fleet button, `omnigent:…` host-chip options, per-familiar fleet defaults) appear if and only if the user has `OMNIGENT_TOKEN` set up in their Cave Vault; credentials that exist only in `~/.omnigent/auth_tokens.json` no longer surface Fleet UI on their own, and the token itself now resolves through the Vault chain (env → `.env.local` → encrypted store → `op://`/`dl://` reference).
 
 ## [0.1.2] - 2026-07-16
--- a/src/app/globals.css
+++ b/src/app/globals.css
@@ -108,6 +108,12 @@
      radius appearance setting: `sharp` overrides it to a squared value. */
   --radius-pill: 999px;
 
+  /* ---- Terminal-numeral glow (cave-uw7c) ----
+     The profile share card's glowing-stat treatment, generalized for themed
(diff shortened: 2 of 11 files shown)


BunsDev (commented):
**Review pass (branch-fleet-reduction · session be81eef1) — verified against origin/main...origin/futurist-profile-cards, no blocking findings.**

Checked:
- **Ramp parity claim is true**: `activityLevel` uses the identical relative-to-max formula as `levelFor` in `src/lib/profile-card.ts:135-138` (`min(4, max(1, ceil(count/max*4)))`), so the roster strip and profile heatmap really do bucket the same way.
- **UTC-day bucketing is correct** and tested: epoch-day index math (`todayIndex - floor(ms/DAY_MS)`) skips future-dated sessions (`daysAgo < 0`), archived sessions excluded upstream; `familiars-view-stats.test.ts` pins strip length, per-day cells, and the sum invariant.
- **All four `FamiliarCardStats` producers updated** (familiars-view, familiar-growth-view, familiar-analytics-data, dashboard-cockpit `EMPTY_STATS`) — no missing-field drift.
- **`familiar-roster-card.test.ts` is already wired** in `scripts/run-tests.mjs` (line 452), so check-tests-wired stays green.
- **A11y/theming details handled**: status-dot pulse is disabled under `prefers-reduced-motion`, light mode zeroes `--glow-numeral` (bloom-on-paper), strip cells are `aria-hidden` with the count conveyed by the visible stat band + title.

Non-blocking note:
- `dashboard-cockpit.tsx` `EMPTY_STATS` is a module-level singleton whose new `activity` array is shared by reference across every consumer that falls back to it. All current usage is read-only, but a future `stats.activity[i]++`-style mutation would corrupt every card sharing the sentinel. A `() => emptyStats()` factory (as the other three files already use) would remove the foot-gun.

PR #3356, opened 2026-07-17
fix(desktop): recover orphan migration takeovers

CompleteDotTech:
## Summary
- record takeover claimant PID and start time in migration lock markers
- reclaim dead claimant markers and aged legacy markers so startup cannot retry EEXIST forever
- add regression coverage for the orphan legacy takeover state reproduced on Windows

## Root cause
A crashed contender left .coven/cave/.migration.lock/.takeover behind. The next packaged launch repeatedly hit EEXIST while trying to claim the stale migration lock, so the server health endpoint reported ready but workspace initialization never completed and the native startup window remained visible.

## Verification
- node --experimental-strip-types --import ./scripts/test-alias-register.mjs src/lib/server/cave-home-migration.test.ts
- pnpm typecheck
- node scripts/check-tests-wired.mjs
- pnpm build (pre-PR packaged candidate validation)
- Windows packaged StartupProbeOnly harness: workspace ready in 6836 ms from rebuilt MSI candidate

## Tracking
- Bead: cave-wq3
- Related to #3269

### DIFF
--- a/scripts/cave-home-migration-windows.test.ts
+++ b/scripts/cave-home-migration-windows.test.ts
@@ -1,6 +1,6 @@
 // @ts-nocheck
 import assert from "node:assert/strict";
-import { lstat, mkdir, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
+import { lstat, mkdir, mkdtemp, readFile, rm, utimes, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import path from "node:path";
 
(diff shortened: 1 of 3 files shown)


CompleteDotTech:
Follow-up from live packaged-app verification: the running sidecar could leave a recent .takeover marker owned by its own still-live PID after a Windows reclaim rename failed. That made later storage-backed requests treat the abandoned token as an active contender indefinitely. Commit 3e5355f8 tracks active takeover tokens, removes the marker on failed reclaim attempts, and immediately recovers untracked same-process claims. Added both cross-platform and Windows-native regressions. Verification: focused migration suite passed; dedicated Windows migration conformance passed under Windows-native Node; pnpm typecheck passed; all 1027 tests remain wired.

CompleteDotTech:
Addressed the takeover review findings in commit `63ca3f48`:

- takeover claims are now immutable and atomically published from private candidates
- orphan recovery fences the entire lock generation at a deterministic retained nonempty path, so delayed observers cannot rename a successor lock
- claim content and age are read from one open filesystem object
- apparently live claims expire after the bounded stale window, covering PID reuse
- same-process takeover tokens are registered before publication
- owner/claim/birthtime are revalidated as one stable generation immediately before fencing

Regression coverage now exercises concurrent orphan observers, retained generation fencing, registration-before-publication, same-process orphan recovery, and reused-PID expiry on both the focused suite and Windows-native conformance. Validation: focused migration suite passed; Windows-native migration test passed; `tsc --noEmit` passed; all 1027 tests remain wired; `git diff --check` passed.

CompleteDotTech:
Post-merge review complete against merge commit `f635ae9f` (PR head `63ca3f48`). I did not find any additional actionable correctness or concurrency issues, so no follow-up code commit was necessary.

Revalidated in an isolated worktree:
- focused Cave home migration suite passed
- `pnpm typecheck` passed
- all 1031 tests are wired into CI
- `git diff --check` passed

The merged PR checks are also green across CI and CodeQL.

PR #3376, opened 2026-07-17
fix(browser): replace perpetual webview reconciliation

### LINKED ISSUE #3268: Replace BrowserPane's perpetual rAF and document-wide mutation scan
## Performance problem

While the native browser pane is active, `src/components/browser-pane.tsx` starts a self-rescheduling `requestAnimationFrame` loop for the lifetime of the pane. The expensive reconciliation is throttled to about 10 Hz, but the callback still wakes at display refresh rate. Reconciliation calls `getBoundingClientRect()` and `surfaceIsCovered()`, which performs document queries/hit tests.

The same effect installs a `MutationObserver` on `document.body` with `subtree: true`, observing `role`, `aria-modal`, `hidden`, `open`, `class`, and `style` changes across the entire app. Chat streaming, animation classes, modals, layout transitions, and unrelated surface updates can all schedule forced reconciliation.

This design can keep CPU usage and layout work elevated whenever the browser is open, reducing animation smoothness and battery life even when its bounds are stable.

## Suggested direction

- Drive reconciliation from `ResizeObserver`, scroll/resize, known shell/banner/modal lifecycle events, and a short-lived animation window instead of a permanent 60 Hz rAF chain.
- If a safety poll remains necessary, schedule it at the actual desired rate and stop it after a stable period.
- Narrow mutation observation to overlay roots/attributes that can actually cover the native webview.
- Batch reads before Tauri IPC writes and instrument reconciliation count/duration.

## Acceptance criteria

- Stable browser-pane idle state does not execute a display-rate callback or observe unrelated document mutations.
- Native webview bounds remain correct during resize, panel animation, banners, scrolling, modal/onboarding/palette overlays, visibility changes, and tab switches.
- Capture before/after idle CPU and reconciliation counts for a five-minute Browser session.
- Existing browser lifecycle and overlay regression tests continue to pass, with new coverage for the event-driven/stability behavior.

CompleteDotTech:
## Summary

- replace BrowserPane's self-rescheduling `requestAnimationFrame` loop with event-coalesced reconciliation
- sample at 100 ms only during a bounded 400 ms CSS motion window, including the initial pane animation
- replace the document-wide subtree observer with direct-body portal detection and explicit shell-banner layout events
- expose `window.__CAVE_BROWSER_RECONCILE_METRICS__` with reconciliation count and duration for native profiling
- extend BrowserPane regression coverage for idle stability, overlays, animation windows, and banner movement

## Root cause

The previous effect scheduled a callback every display frame for the entire lifetime of an active BrowserPane. Although expensive bounds and occlusion reads were throttled to roughly 10 Hz, the renderer still woke at display rate and continued querying the document indefinitely. Its `MutationObserver` also watched the entire app subtree for common class and style changes, so unrelated streaming and animation activity forced more reconciliation.

## Impact

At 60 Hz, a stable five-minute Browser session previously scheduled about 18,000 frame callbacks and up to 3,000 expensive reconciliation passes. The new stable path performs its initial reconciliation and then sleeps until resize, scroll, visibility, interaction, overlay, banner, tab, or relevant CSS-motion events occur. Motion sampling stops after 400 ms.

The metrics object can be reset in native devtools before a five-minute capture:

```js
window.__CAVE_BROWSER_RECONCILE_METRICS__ = undefined
```

It then reports `count`, `totalDurationMs`, `lastDurationMs`, and `startedAt`.

## Validation

- `node --experimental-strip-types src/components/browser-polish.test.ts`
- `pnpm typecheck`
- `pnpm check:tests-wired` (1,038 test files wired)
- `pnpm test:app` (764 app test files)
- `pnpm build` (Next production build, bundle budget, standalone budget)
- `git diff --check`

## Native five-minute profile

Captured on Windows from the native Tauri Browser route against exact pre-change `origin/main@369db97b` and the patched branch. Both runs stubbed only `browser_*` IPC so the main renderer reconciliation workload was isolated without launching the child page webview.

| Metric | Before | After | Change |
| --- | ---: | ---: | ---: |
| `requestAnimationFrame` callbacks | 35,998 | 416 | -98.84% |
| Browser surface reconciliations | 2,944 | 415 | -85.90% |
| `elementFromPoint` hit tests | 14,720 | 1,712 | -88.37% |
| Tauri process-tree CPU | 115.7031 s | 110.4062 s | -4.58% |
| Average process-tree CPU | 38.568% | 36.802% | -1.766 pp |

The live dev session had repeated unrelated server lock-error UI churn in both captures, so the event-driven path still reconciled 415 times; the absence of the prior perpetual loop is demonstrated by the 98.84% callback reduction and source/test guard.

Closes #3268

Bead: `cave-kgy`

### DIFF
--- a/src/components/browser-pane.tsx
+++ b/src/components/browser-pane.tsx
@@ -631,7 +631,10 @@ export function BrowserPane({ label = "default", activeFamiliarId = null, active
     document.addEventListener("keydown", onInteraction, true);
     document.addEventListener("focusin", onInteraction, true);
     document.addEventListener("visibilitychange", onVisibilityChange);
-    scheduleImmediateReconcile();
+    // The pane's cave-mode-fade animation can begin before this effect's
+    // animationstart listener is attached. Sample the initial mount window
+    // explicitly so the native surface follows that 120ms transform too.
+    startMotionWindow();
 
     return () => {
       cancelAnimationFrame(raf);
--- a/src/components/browser-pane.tsx
+++ b/src/components/browser-pane.tsx
@@ -213,9 +213,7 @@ function defaultPinnedTabs(): BrowserTab[] {
 //      live regions (toasts) are ignored so a corner toast doesn't blank the
 //      page.
 function surfaceIsCovered(surface: HTMLElement, rect: DOMRect): boolean {
-  const overlays = document.querySelectorAll(
-    '[role="dialog"], [aria-modal="true"], [role="menu"], [role="listbox"], [data-native-webview-cover="true"]',
-  );
+  const overlays = document.querySelectorAll(NATIVE_WEBVIEW_COVER_SELECTOR);
   for (const overlay of overlays) {
     if (overlay.getClientRects().length > 0) return true;
   }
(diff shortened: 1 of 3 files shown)


CompleteDotTech:
Acceptance pass complete on the updated branch.

- Rebasing onto current `main` was conflict-free.
- Review found one initial-animation race: `cave-mode-fade` can start before the effect attaches its `animationstart` listener. Commit `2b3a2f94` now opens the bounded initial motion window explicitly and adds a regression guard.
- Exact five-minute Windows/Tauri profile against pre-change `origin/main@369db97b`: rAF callbacks `35,998 → 416` (-98.84%), surface reconciliations `2,944 → 415` (-85.90%), hit tests `14,720 → 1,712` (-88.37%), average Tauri process-tree CPU `38.568% → 36.802%` (-1.766 pp / -4.58% relative). The PR body contains the capture method and caveat about unrelated live server error churn.
- Local verification passed: focused BrowserPane regression test, typecheck, 1,038-test wiring check, and all 764 app test files. Production build is completing locally while the refreshed GitHub checks run.

I will keep watching the new head and address any review or CI failures before marking it ready.

CompleteDotTech:
All refreshed checks are now complete on head `2b3a2f94`: 19 passing, 0 failing, and 0 pending. Local production build also completed with bundle and standalone budgets passing. A final thread-aware review scan found no reviews or unresolved threads, so the PR is now marked ready for review.

CompleteDotTech:
Post-merge acceptance review found one overlay lifecycle gap: Workspace-owned overlays are Shell siblings, not direct body portals, so a programmatic open (notably async first-run onboarding) can bypass the narrowed portal observer and leave the native child webview above the overlay. The narrow follow-up is committed in #3381 (`9fd653d3`), with explicit cover-state reconciliation and regression guards. I am carrying that PR through full app tests, review, and CI before considering the acceptance pass complete.

PR #3381, opened 2026-07-18
fix(browser): reconcile programmatic workspace overlays

CompleteDotTech:
## Summary

- dispatch an explicit native-webview cover lifecycle event whenever Workspace-owned overlay state changes
- coalesce that event into BrowserPane visibility reconciliation after the React DOM commit
- add regression guards for programmatic onboarding, palette, and modal opens

## Root cause

Follow-up acceptance review of #3376 found that Workspace-owned overlays are Shell siblings rather than direct `document.body` portals. Programmatic opens, especially first-run onboarding after an async status response, therefore have no click/key event or body portal mutation to wake the new event-driven BrowserPane reconciler. The native child webview could remain visible above the overlay.

## Validation

- `node --experimental-strip-types src/components/browser-polish.test.ts`
- `pnpm typecheck`
- `git diff --check`

## Tracking

Follow-up to #3268 and #3376.

Bead: `cave-9zx`

### DIFF
--- a/src/components/browser-pane.tsx
+++ b/src/components/browser-pane.tsx
@@ -619,9 +619,11 @@ export function BrowserPane({ label = "default", activeFamiliarId = null, active
     // state rather than a body portal. The rAF runs after the DOM commit.
     const onInteraction = () => scheduleImmediateReconcile();
     const onShellLayout = () => startMotionWindow();
+    const onCoverChange = () => scheduleImmediateReconcile();
     window.addEventListener("resize", scheduleImmediateReconcile);
     window.addEventListener("scroll", scheduleImmediateReconcile, true);
     window.addEventListener("cave:native-webview-layout", onShellLayout);
+    window.addEventListener("cave:native-webview-cover-change", onCoverChange);
     window.addEventListener("cave:onboarding-open", onShellLayout);
     document.addEventListener("animationstart", onMotionStart, true);
     document.addEventListener("animationend", onMotionEnd, true);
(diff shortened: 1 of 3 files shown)


CompleteDotTech:
Local acceptance verification is complete on head `9fd653d3`:

- focused BrowserPane regression passed
- TypeScript typecheck passed
- all 1,038 test files are wired into CI
- all 764 app test files passed
- diff check passed

The PR remains draft while the exact-head GitHub rollup runs. I will address any CI or review findings before moving it to ready.

CompleteDotTech:
Acceptance follow-up is in progress. Focused BrowserPane regression, typecheck, test-wiring validation (1,038 files), and diff check pass. The full 764-file app suite and refreshed remote checks are running; I will address any review or CI failures before moving this out of draft.

CompleteDotTech:
Reopened to finish the requested acceptance loop. The follow-up remains necessary for async/programmatic Workspace overlay opens, and it will stay draft until the exact-head CI rollup and thread-aware review scan are complete.

CompleteDotTech:
The implementation checks passed, but closing/reopening the draft while the first exact-head run was active created duplicate cancelled suites whose newer check records blocked the rollup. This is workflow-state churn, not a code failure: the original CI run completed successfully on all matrix legs and aggregates. I reran only the cancelled/failed jobs in the newer CI and CodeQL suites so the final exact-head check records can reconcile cleanly.

PR #3458, opened 2026-07-18
fix(ios): retain bubble updates for theme and familiar changes

CompleteDotTech:
Follow-up for #3454.

- compare the full `Familiar` value so refreshed names, avatars, and colours invalidate the bubble
- compare `colorScheme` and `chrome` so palette and markdown-theme changes remain visible

Validation: `git diff --check` passed. Native iOS/Xcode validation requires macOS; this WSL environment has neither `swift` nor `xcodebuild`.

### DIFF
--- a/apps/ios/CovenCave/CovenCave/Views/MessageBubble.swift
+++ b/apps/ios/CovenCave/CovenCave/Views/MessageBubble.swift
@@ -465,15 +465,19 @@ struct SuggestionPills: View {
 // every bubble (each hosts a WKWebView) — the cause of slow typing on long
 // threads. Closures are excluded (they capture fresh state each render but
 // don't change what's drawn); optional-action closures are compared by presence
-// since that toggles which buttons appear.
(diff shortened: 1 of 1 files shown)


CompleteDotTech:
Maintenance audit for `be991b95086e2fb1d15b3a60589cd58d830df0d7`:

- No additional render-affecting inputs are omitted: `Familiar`, `colorScheme`, `chrome`, and all optional action-presence inputs are compared.
- `git diff --check origin/cody/perf-chat-render...HEAD` passes; the PR is conflict-free and has no review threads.
- This follow-up correctly has no CI runs because CI and CodeQL trigger only for PRs targeting `main`; after a maintainer merges it into #3454, the parent PR head must rerun and pass its exact-head checks before the original review thread can be resolved.
- Native iOS compilation remains unverified here because this WSL host lacks Xcode, `xcodebuild`, and Swift tooling.

PR #3459, opened 2026-07-18
fix(ios): flush coalesced chunks on transport errors

CompleteDotTech:
Follow-up for #3455.

Flushes the shared `StreamCoalescer` before recovering a failed send stream and on each failed/no-resumable resume path, so chunks received during the final coalescing interval are not lost. Updates the mobile native source-contract assertions accordingly.

Validation:
- `pnpm typecheck`
- `pnpm check:tests-wired`
- `pnpm test:mobile` (81 files)

The original head branch is owned by `OpenCoven` and does not permit maintainer edits, so this must merge into `cody/perf-stream-coalesce` before #3455 can pass its mobile check.

### DIFF
--- a/apps/ios/CovenCave/CovenCave/State/ChatThread.swift
+++ b/apps/ios/CovenCave/CovenCave/State/ChatThread.swift
@@ -281,6 +281,7 @@ final class ChatThread: Identifiable, Hashable {
             // (cave-h40l). Only when no resumable run exists (finished long
             // ago / server restarted) fall back to adopting the persisted
             // transcript.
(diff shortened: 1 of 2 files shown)


CompleteDotTech:
Superseded by https://github.com/OpenCoven/coven-cave/pull/3460, which carries the same commits but targets `main` so required CI runs.

CompleteDotTech:
Closing in favor of https://github.com/OpenCoven/coven-cave/pull/3460 because this PR's non-main base does not trigger repository CI.

PR #3460, opened 2026-07-18
perf(ios/chat): coalesce assistant stream tokens

CompleteDotTech:
Supersedes #3455, whose OpenCoven-owned head branch cannot be updated by maintainers (`maintainer_can_modify: false`; direct push returns HTTP 403). This branch retains the original performance commit by @BunsDev and adds the review correction.

## Behavior
- Buffers assistant chunks and flushes at a ~50ms cadence instead of causing an observed-array update and UI/persistence notification per token.
- Flushes buffered text before send-stream resume/resync and on every failed/no-resumable resume attempt, so the final coalescing window is never dropped.

## Validation
- `pnpm typecheck`
- `pnpm check:tests-wired`
- `pnpm test:mobile` (81 files)

The normal PR CI is enabled because this replacement targets `main`.

### DIFF
--- a/apps/ios/CovenCave/CovenCave/State/ChatThread.swift
+++ b/apps/ios/CovenCave/CovenCave/State/ChatThread.swift
@@ -328,9 +328,9 @@ final class ChatThread: Identifiable, Hashable {
             // @Observable class — invalidating the whole list — so per-token
             // updates caused a render/scroll storm. Coalescing flushes at most
             // ~every 50ms while keeping streaming visibly live.
-            coalescer.append(chunk)
-            if coalescer.shouldFlush() {
-                flush(coalescer, into: messageId, onChange: onChange)
+            coalescer.append(chunk) { [weak self] in
+                guard let self else { return }
+                self.flush(coalescer, into: messageId, onChange: onChange)
             }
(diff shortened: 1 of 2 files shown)


CompleteDotTech:
Follow-up pushed in `469081d8`: the coalescer previously drained only when another SSE frame arrived, so a final chunk could remain buffered indefinitely while a stream paused. It now schedules one 50 ms main-actor flush, while all terminal/recovery paths still drain idempotently and cancel the pending task.

Validated locally with `pnpm typecheck`, `pnpm check:tests-wired`, and `pnpm test:mobile` (81 files). The exact-head CI rollup is green (Frontend build, E2E, cross-environment, Rust, sidecar runtime on Ubuntu and Windows, and CodeQL); no unresolved review threads or merge conflicts remain.

CompleteDotTech:
Validation complete on the final PR head: all required CI checks passed (Frontend build, E2E, Rust, cross-environment, both sidecar-runtime jobs, and CodeQL); review-thread audit found no unresolved threads or review findings. This replacement is mergeable and ready for maintainer review. It preserves @BunsDev's original performance commit and includes the buffered-text recovery fix from #3455's review.

PR #3592, opened 2026-07-21
feat(settings): restyle chat settings to the design-handoff card (cave-otrn)

BunsDev:
Implements `Settings.dc.html` from the claude.ai/design **Projects page redesign** handoff.

- One bordered auto-archive card: master-switch row, three event-toggle rows, two idle-day rows, per the mock; uppercase kicker + helper copy; "Changes save automatically." footer.
- Mock-style switches (36×20 pill, tint-recipe on-state, translated knob) scoped as `chat-settings__*` — the shared `.settings-switch` used by settings-shell and the studio Brain tab is untouched (both pinned by tests).
- Behavior preserved byte-for-byte where pinned: `/api/config` load + optimistic PATCH with revert-on-failure, `role="switch"`/`aria-checked`, real `disabled` on children when the master is off (mock's dim treatment is additive), 0–365 clamp + commit-on-blur day fields, `aria-live` save state.
- Divergences from the mock: day-field max stays 365 and commit-on-blur (behavior wins); "Familiars → Brain" stays plain text (adding navigation would be new behavior); sizes snapped to the token scales per the design-token-drift ratchet (no baselines raised).

Verification: tsc clean; `run-tests.mjs app` → 839 files pass incl. `chat-settings-view.test.ts`; verified in the running app — master-on and master-off (dimmed children) states screenshotted.

Bead: cave-otrn

### DIFF
--- a/src/components/chat-settings-view.tsx
+++ b/src/components/chat-settings-view.tsx
@@ -1,7 +1,6 @@
 "use client";
 
 import { useCallback, useEffect, useRef, useState, type ReactNode } from "react";
-import { SettingsGroup } from "@/components/ui/settings-group";
 import {
   normalizeChatAutoArchivePolicy,
   type ChatAutoArchivePolicy,
@@ -14,6 +13,10 @@ import {
  * event-driven archiving (task completion, thread reflections), and the idle
  * windows. Reads/writes the `chatAutoArchive` key through `/api/config`, whose
  * PATCH merges partial policies over the stored one.
+ *
+ * Presentation follows the Settings.dc.html redesign mock: one bordered card
+ * whose child rows dim as a block while the master switch is off. Styles live
+ * in cave-chat/auxiliary-surfaces.css under `.chat-settings*`.
  */
 
 type SaveState = "idle" | "saving" | "error";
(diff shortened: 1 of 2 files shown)


BunsDev:
Reviewed from the Coven code lane — solid, low-risk restyle. Behavior is fully preserved (only markup + classNames changed; `update()`/`sweepOff`/policy plumbing untouched), the disabled child-block a11y is done correctly (`pointer-events: none` for pointer **and** `disabled` on every control so screen readers get real state), and the scoped `.chat-settings__switch` fork avoids regressing the shared `.settings-switch` that other surfaces' tests pin. Token/`color-mix` discipline is clean too.

One non-blocking suggestion: add a tiny render/DOM test locking the master-off contract — assert that toggling the master switch off applies `is-disabled` on `.chat-settings__children` **and** keeps `disabled` on the child controls. The repo already has `settings-shell-polish.test.ts` as a home for this, and it'd guard the a11y behavior against a future refactor that drops one half of the pattern (visual dim without the SR `disabled`, or vice versa).

Minor nit: the number input is `min-height: 32px` — under the 44px mobile tap guideline if this surface renders on iOS. Low risk, worth a glance only if mobile is in scope.

LGTM once CI is green. Doesn't conflict with #3590 (disjoint files).

PR #3613, opened 2026-07-21
fix(debug): gate events live-tail on truthful liveness, not the polled row (cave-tvzi, A3)

BunsDev:
## Problem (A3 from the chat-session-debugging audit)

DebugPane's events live-tail polls only while `session?.status === "running"`, where `session` is the chat-router row from the **polled** sessions list:

- **No row** (pre-promotion chats, list lag) → `status === null` → the live tail **never starts**, even while events stream in.
- **Poll lag** → the tail starts late, and the one-shot catch-up on running→finished fires late.
- A run driven **from another surface** (CLI, second tab) shows a stale tail until reopen.

## Fix

New pure `debugSessionLive({status, streamPhase, serverRunDone})` in `session-debug.ts` — live when **any** of three independent signals says events are flowing:

1. the polled row says `running` (existing signal);
2. the pane's **own stream transport** is `connecting`/`streaming`/`resuming` — instant, no poll lag;
3. the **server run buffer is open** — `RunBufferStatus.done === false` from `GET /api/chat/stream/status`, which is dual-keyed by conversation id (`openRunBuffer(keys)`), so it's daemon-side truth even when another surface drives the run. `null` (no buffer) is not evidence either way.

Pane wiring:
- both pollers (events tail + stream status) gate on the combined `live` flag;
- the two live→terminal catch-up transitions merge into one effect refreshing both (final-window events aren't dropped);
- tail-follow seeds from `live`; Stream health section defaults open while `live`;
- `shouldPollEvents` now takes `{live, visible}`.

Termination is server-truth-driven: when the run finishes, `done: true` arrives, `live` flips, one final catch-up fires, polling stops. `live` derives from the `done` boolean, so per-poll object churn doesn't restart intervals.

## Tests

- `session-debug.test.ts`: behavioral truth table for `debugSessionLive` (each signal independently, all-cold, non-live phases) + `shouldPollEvents` combinations.
- `debug-pane.test.ts`: pins for the liveness computation, both poller gates, merged transition, and follow seeding.

## Verification

- `node --test` on session-debug/debug-pane/chat-debug-store/chat-stream-health: pass
- `pnpm typecheck`: clean
- `pnpm test:app`: 851 files green

Bead: cave-tvzi. Goal: chat-session-debugging (slice A3).

### DIFF
--- a/src/components/debug-pane.test.ts
+++ b/src/components/debug-pane.test.ts
@@ -281,18 +281,28 @@ assert.match(
 );
 assert.match(
   source,
-  /const streamStatusActive =\s*status === "running" \|\|\s*streamHealth\.phase === "connecting" \|\|\s*streamHealth\.phase === "streaming" \|\|\s*streamHealth\.phase === "resuming"/,
-  "Server status activity must include authoritative connecting, streaming, and resuming client phases",
+  /const live = debugSessionLive\(\{\s*status,\s*streamPhase: streamHealth\.phase,\s*serverRunDone: streamStatus \? streamStatus\.done : null,\s*\}\)/,
+  "Pane liveness must combine the polled row, own transport phase, and server run buffer (A3: poll-lag-free)",
 );
 assert.match(
   source,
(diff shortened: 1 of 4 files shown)


BunsDev:
Duplicate of #3611 (cave-rxjm) — a parallel session landed the identical A3 fix (same three-witness liveness design, same shouldPollEvents signature, merged catch-up) while this PR's checks ran. Closing; no unique changes to salvage.

PR #3614, opened 2026-07-21
fix(debug): events tail self-detects liveness under a lagged or absent sessions-list row (cave-386l, A3)

BunsDev:
## What

Closes the A3 gap from the chat-session-debugging audit: the debug pane's live event tail was gated **solely** on the polled sessions list's row status (`session?.status`), which lags reality by a poll cycle and can lack a row for the session entirely. Consequences:

- **No row → the live tail never started** (pre-listing runs, sessions opened from outside the polled view).
- **Post-send lag**: a fresh run on a row still marked `completed` only started tailing after the next sessions-list poll.

## How

New pure gate `eventsTailActive(signals)` in `src/lib/session-debug.ts` — the polled row becomes one hint among three:

1. **Running row** — definitive, as before.
2. **This pane's transport phase** (`connecting`/`streaming`/`resuming`) — starts the tail immediately after a send, ahead of the lagged list; its settling now fires the same one-shot final-events catch-up the row transition already had. Tail-follow's initial state honors it too.
3. **Statusless row → bounded activity probe** anchored on the newest daemon event timestamp (`latestEventTimestampMs`, seeded from the A2 cached tail, falling back to pane mount). Live sessions keep extending the window; draining an old finished tail closes it immediately; a quiet probe stops the poll timer (`probeExpired`) instead of ticking forever. Fresh events surfacing later — including via the manual **Retry** — re-open it.

## Testing

- Behavioral tests for `eventsTailActive` precedence (phase beats a stale settled row; settled row with idle transport never polls; probe window boundary is strict) and `latestEventTimestampMs` (order-independent max, unparseable timestamps skipped).
- Source pins for the pane wiring: composite timer gate, per-tick pure-gate re-evaluation + expiry flip, daemon-truth anchor stamping + probe re-open in `fetchEvents`, cache-seeded anchor, follow seed, phase-settle catch-up.
- `pnpm test:app` 851 files green; `pnpm typecheck` clean.

Known bounded residual (recorded in bead cave-386l): a *statusless* session whose tab stays hidden past the 15s window stops probing until Retry or a reappearing row — the row normally appears within one poll cycle, so this is an edge of an edge.

Bead: cave-386l (goal chat-session-debugging, backlog item A3).

### DIFF
--- a/src/components/debug-pane.test.ts
+++ b/src/components/debug-pane.test.ts
@@ -281,7 +281,7 @@ assert.match(
 );
 assert.match(
   source,
-  /const streamStatusActive =\s*status === "running" \|\|\s*streamHealth\.phase === "connecting" \|\|\s*streamHealth\.phase === "streaming" \|\|\s*streamHealth\.phase === "resuming"/,
+  /const streamPhaseActive =\s*streamPhase === "connecting" \|\| streamPhase === "streaming" \|\| streamPhase === "resuming";\s*const streamStatusActive = status === "running" \|\| streamPhaseActive/,
   "Server status activity must include authoritative connecting, streaming, and resuming client phases",
 );
 assert.match(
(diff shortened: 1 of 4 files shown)


BunsDev:
Superseded by #3611 (cave-rxjm), which landed the same A3 events-tail liveness slice minutes earlier from a parallel session — third identical-slice race on this goals file. #3611's server run-buffer witness (RunBufferStatus.done === false) covers the statusless-row case this PR probed for with a bounded 15s activity window; the only residual is a statusless row whose run buffer is unknown/reaped (no live tail until a list row appears), which is acceptable. Closing unmerged; bead cave-386l will be closed as superseded.

BunsDev:
Coordination: #3611 (cave-rxjm) merged the core A3 fix first (row status + transport phase + server run-buffer done:false). This PR is a duplicate for those two signals, but its third signal — the bounded latestEventTimestampMs activity probe for statusless rows — covers externally-driven sessions the run buffer can't see (buffer is per-process; daemon-driven runs from another surface have no buffer here). If the owning session resumes: rebase onto main and cut this down to just the probe increment on top of isDebugSessionLive. Otherwise close cave-386l as subsumed.

PR #3636, opened 2026-07-21
fix(hermes): expose authenticated models and repair Debug events

CompleteDotTech:
## What changed

Hermes now renders the authenticated Codex model selector in Familiar Studio. The native Tauri dev launcher also forwards its sidecar token via the URL fragment, allowing the browser bridge to authenticate Debug API requests.

The shared session-ID validator accepts Hermes timestamp-prefixed IDs while retaining the conservative no-path-separators grammar.

## Why

Hermes was incorrectly treated as free-text-only, so its model dropdown disappeared. Separately, Debug first failed authorization in native dev and then rejected valid Hermes session IDs.

## Validation

- Windows-native `pnpm typecheck`
- `runtime-models.test.ts`
- `session-security.test.ts`
- Brain-tab selector source contract
- `dev-app.test.mjs`
- Native Tauri launch with authenticated sidecar bridge on port 3001

### DIFF
--- a/scripts/dev-app.sh
+++ b/scripts/dev-app.sh
@@ -53,6 +53,17 @@ TAURI_OVERRIDE_CONFIG="$(mktemp)"
 cleanup() { rm -f "$TAURI_OVERRIDE_CONFIG"; }
 trap cleanup EXIT
 
+# A Tauri dev process and its `beforeDevCommand` inherit this secret together.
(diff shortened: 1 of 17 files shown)


CompleteDotTech:
@BunsDev ready for review. Tested and compiled on Windows with the native Tauri app; Windows `pnpm typecheck` and focused Hermes regression tests pass.

PR #3700, opened 2026-07-22
fix: populate GitHub org filter from memberships

CompleteDotTech:
## Summary

Populate the GitHub Organization selector from the authenticated account's organization memberships, then merge in organizations represented by current activity.

## Root cause

The selector was intentionally introduced in #519 as an activity-derived filter to avoid an additional GitHub API/type path. As a result, any organization without an open authored, review-requested, or assigned item disappeared from the list.

## Changes

- Fetch authenticated memberships from `GET /user/orgs?per_page=100`.
- Return them with the GitHub activity payload and merge them into the Organization options.
- Request `read:org` for newly configured PATs so private memberships are available.
- Accept standard GitHub token aliases supplied by supported install and launcher environments.
- Preserve Vault-managed GitHub aliases only for the familiar granted that credential; launcher-provided aliases require an explicit harness opt-in, so generic child processes still cannot inherit them.
- Add focused source-contract coverage.

## Compatibility and credential boundary

Cave supports Node/package-manager, native desktop (Windows, macOS, Linux), and runtime/harness-based installations such as Codex, OpenClaw, Hermes, and OpenCode. The GitHub API recognizes standard launcher aliases across those setups. A child harness receives an external alias only with `COVEN_HARNESS_ALLOW_ENV_KEYS` (or the legacy OpenClaw-specific allowlist), while a Vault alias follows its familiar scope.

## Validation

- Windows-native `pnpm typecheck`
- `pnpm test:api`
- `pnpm check:tests-wired`
- Native Tauri rebuild and live API verification: the response included `OpenCoven` and membership-only organizations absent from current activity.

Beads: `cave-cuy`.


CompleteDotTech:
Follow-up Settings work is tracked in #3701 (Beads `cave-lpl`), with @BunsDev assigned for the UI. This PR remains intentionally limited to correcting GitHub organization membership data and the GitHub filter.

PR #3747, opened 2026-07-24
fix(windows): resolve npm-global coven CLI + surface codex auth failures

BunsDev:
## Summary

Two user-facing errors from a single Windows report (user's CLI was healthy at `C:\Users\Admin\AppData\Roaming\npm\coven.cmd`, and had historically run an older Cave build). Neither reproduces on a fresh dev box.

### 1. `Coven CLI not found on PATH` — `sidecar_discovery.rs`
`find_coven()` only probed `coven.exe` under `.volta` / `.bun` / `.cargo`, then fell back to `where.exe`. But npm global installs land at `%APPDATA%\npm\coven.cmd` (a `.cmd` shim, not `.exe`), and a GUI-launched Tauri app usually does not inherit `%APPDATA%\npm` on PATH — so `where.exe` missed it too. Result: "not found on PATH" despite a healthy install.

**Fix:** probe `%APPDATA%\npm\coven.cmd` / `.exe` first (with an `APPDATA` fallback), before the existing candidates.

### 2. `assist produced no output` — `assist-runner.ts`
codex can exit `0` yet write no last-message file when it is not signed in (or refuses workspace trust). The runner already keeps a bounded stderr tail, but the empty-output path discarded it and returned an opaque message.

**Fix:** on empty output, include the stderr tail and — when it looks like auth (`not logged in / unauthorized / 401 / …`) — return an explicit "run \`codex login\`" message.

## Verification
- `cargo check` — clean (Rust)
- `tsc --noEmit` — clean (TS)

## Out of scope (deliberately)
The reporter had historically run an older Cave build, so stale sidecar runtimes are possible. `cleanup_stale_sidecar_runtimes` only runs *after* a successful start, so a wedged box never sweeps them. However, the existing code intentionally leaves **complete** generations to explicit-uninstall reclamation (no startup-time lease proof that another app isn't serving them); only **incomplete** extractions are cleared. Splitting the sweep (early incomplete-only) is a real change to a carefully-reasoned concurrency path and belongs in its own PR with sidecar-lease-owner review.


BunsDev (commented):
**Review (Copilot CLI session):**

**Rust (`find_coven`)** — correct and safe. Verified the consumer (`sidecar_startup.rs:200-208`) only prepends the found file's *parent dir* to the sidecar PATH, so discovering a `.cmd` shim (not directly spawnable by `CreateProcess`) is semantically right: the Node side resolves it via `covenBin()` → `where` → `covenLaunchCommandForBinary()` (shim → `node <script>`). npm-first ordering matches `coven-bin.ts`'s documented preference of the npm-published binary over stale `.cargo` installs. `APPDATA` fallback to `%USERPROFILE%\AppData\Roaming` degrades harmlessly when both are unset.

**TS (`assist-runner`)** — direction right, but the auth heuristic matched bare `auth`/`token`/`log ?in`/`sign ?in` substrings, so e.g. `SyntaxError: Unexpected token < in JSON` or `redesign in progress` stderr would falsely instruct `codex login`. Pushed `babe1a3ae`:
- extracted pure, exported `stderrReason` + `describeEmptyAssistOutput` (per this module's stance: pure logic unit-tested, spawn manual), deduping tail formatting with the exit≠0 path
- tightened the regex to precise auth phrases (missed matches are benign — the tail is still surfaced verbatim; false positives are not)
- folded in `8c15e4e91`'s CodeQL backslash-escaping fix
- added unit tests: auth phrasing, quoted login command for paths with spaces, false-positive regressions, tail carry-through, empty-stderr terseness

**Verified:** `assist-runner.test.ts` green, `tsc --noEmit` clean, eslint clean. Merging once required checks pass.

Issue #3264, opened 2026-07-15
Split the 868 KiB global stylesheet into route and surface-level CSS


CompleteDotTech:
## Performance problem

The root layout imports `src/app/globals.css`, which compiles into two CSS assets totaling **867,964 bytes** for every route (65,948 + 802,016 bytes). The `/` route adds more CSS for a measured initial total of **933,406 bytes (0.89 MiB)**.

`globals.css` is 485 KiB before its imports and globally pulls in xterm, XYFlow, chat, sidebar, home composer, dashboard, and editor styles. The generated 802 KiB root stylesheet contains at least 148 `react-flow__` selectors, 50 `xterm-` selectors, and hundreds of chat selectors. Flow is described as retired in `src/components/lazy-surfaces.tsx`, yet `@xyflow/react/dist/style.css` is still globally imported. Settings, reports, profile, and error pages all pay to download and parse CSS for surfaces they cannot render.

Large global CSS delays style parsing/recalculation, increases memory, and makes it harder for the existing JS-only budget to detect startup regressions.

## Suggested direction

- Keep only tokens, reset, typography, accessibility, and shared shell primitives in root CSS.
- Move chat, dashboard, board, xterm, editor, and other surface rules next to their route/component boundaries.
- Remove the retired XYFlow global import if no live surface needs it; otherwise load it with that lazy surface.
- Add a root/route CSS budget alongside the JS budget.

## Acceptance criteria

- Materially reduce root-layout CSS from the 868 KiB baseline, with before/after asset evidence.
- `/settings`, reports, and error routes no longer reference chat/xterm/XYFlow styles.
- Remove unused retired-surface selectors/imports.
- Add CI coverage for root CSS and `/` first-load CSS budgets.
- Visual regression or targeted UI tests cover Home, Chat, terminal, Board, Dashboard, Settings, and mobile breakpoints.

PR #3780, opened 2026-07-24
refactor(analytics): extract .fa-* / .trace-* CSS out of the global bundle (cave-5rqi, #3264)

BunsDev:
## What

Follow-up to the Familiar Analytics modernization (#3776). The analytics surface CSS (`.fa-*`) and the session-trace overlay CSS (`.trace-*`) lived in `surface-reporting.css`, which `app/globals.css` `@import`s — so ~46 KiB of a **mode-gated** surface rode the **every-route root layout** + the **home first-load** bundle. Per **#3264**, this moves it to component-imported sheets.

## Changes

- **New** `src/styles/familiar-analytics.css` — the `.fa-*` rules; imported by `familiar-analytics-view.tsx`.
- **New** `src/styles/session-trace-overlay.css` — the `.trace-*` rules; imported by `session-trace-overlay.tsx` (rendered from familiar analytics **and** the familiars roster, so it code-splits with each consumer).
- `surface-reporting.css` — analytics/trace rules removed; `.cave-gh-card--reward` (a chat-surface flare that happened to sit mid-region) kept in the global sheet.
- `familiar-analytics-view.test.ts` — the `.fa-*` CSS assertions now read the new sheet instead of the inlined globals facade (the reduced-motion `.fa-page` override stays global and is still read from globals).
- `bundle-budget.mjs` — budgets **lowered** to bank the win.

## Impact

| bundle | before (#3776) | after | new budget |
|---|---|---|---|
| root CSS (every route) | 693 KiB | **646 KiB** | 660 |
| home first-load CSS | 933 KiB | **885 KiB** | 900 |

Both are now **below even the pre-modernization budgets** (690/920).

## Verification

- Rules moved verbatim (no restyle). Drift ratchet **unchanged** — `cssFilesInScope` walks all of `src/`, so the off-scale count is identical (rules only changed files).
- `familiar-analytics-view.test.ts` 29/29, drift/codemod/`css-module-order`/related analytics tests pass, tsc + ESLint clean, `pnpm build` passes at the lowered budgets.

Closes cave-5rqi.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Closes #3795

PR #3784, opened 2026-07-24
perf(css): extract familiar-analytics .fa-* to a component-imported sheet (cave-5rqi)

BunsDev:
## What

Moves the ~1.8k lines of `.fa-*` familiar-analytics CSS out of the global `surface-reporting.css` into a new **component-imported** sheet `src/styles/familiar-analytics.css` (imported only by `familiar-analytics-view.tsx`), per the #3264 css-source-contract pattern. The analytics surface's CSS previously loaded on **every route** and in the **home first load**; now it code-splits into the analytics route chunk.

## Why

`cave-q8uc` had to *raise* the CSS budgets (root 690→696, home 920→936) because the Familiar Analytics modernization grew `.fa-*` in the global bundle. `cave-5rqi` is the tracked follow-up to extract it. Done here.

## Changes

- **New `src/styles/familiar-analytics.css`** — all `.fa-*` rules, `@keyframes fa-refresh-spin`/`fa-target-flash`, the `@container fa` / `@media` responsive tiers, and the reduced-motion block. Includes the mobile `.fa-page`/`.fa-topbar` viewport padding that was stranded in a shared `@media` block in globals (a container query can't size its own container).
- **`surface-reporting.css`** — `.fa-*` fully removed (0 rules remain). The two `:root[data-tauri-titlebar] .fa-topbar` overrides in `desktop-chrome.css` intentionally stay (shell-frame chrome concerns, pinned by `shell-edge-rails`).
- **Tests** — `familiar-analytics-view.test.ts` and `thread-signals-section.test.ts` source-scan asserts repointed from the css-hook-expanded `globals.css` to the new sheet.
- **`bundle-budget.mjs`** — bank the win: root cap 696→**668** (measured 650 KB), home cap 936→**908** (measured 889 KB) — below the pre-fa baseline with ~18 KB headroom so the freed space can't silently regrow.

## Verification

- ✅ 901 app test files pass (`pnpm test:app`)
- ✅ `tsc --noEmit` clean
- ✅ design codemod/drift check + `check:tests-wired` green
- ✅ production build green; bundle-budget & standalone-budget within (tightened) limits
- ✅ `.fa-page` now lives **only** in the analytics route chunk — the every-route root CSS chunk is clean


BunsDev:
Superseded by #3780 (merged), which did this same cave-5rqi extraction — parallel-session duplicate. #3780 left a few residual `.fa-*` rules (mobile `.fa-page`/`.fa-topbar` padding + the reduced-motion analytics block) in the global sheet; opening a small follow-up to move just those, rather than merge this duplicate.

PR #3792, opened 2026-07-24
fix(home): remove the new-chat dashboard's context rail — board only

BunsDev:
Removes the left **context rail** (Project · Quick start · Task · Pick up) from the work-led new-chat dashboard, leaving just the open-work board.

## What changed
- Deleted the `<aside class="home-dash__rail">` block; the board is now the whole surface. It already fills and centers on its own (`.home-dash__board` is `flex:1` with a centered `max-width` inner), so no layout CSS change was needed.
- Trimmed `ChatNewDashboard` to the board-only props (`familiar`, `sessions`, `modelId`), dropped the rail-only props from both the component and its single caller (`chat-view.tsx`), and removed six now-dead imports.
- `recentThreads` now shows the top **6** titled sessions (it no longer skips the two the removed "Pick up" group surfaced), so nothing disappears.

## No functional loss
Project switching stays available in the **composer's context pill** (`ProjectPickerPopover`), and the composer keeps the quick-start / task-arming handlers — so `chat-view`'s state is untouched; only the props it passed *to the dashboard* are dropped (verified: eslint clean, no orphaned vars).

## Tests
- Updated `chat-new-dashboard.test.ts` (was pinning the rail groups) to assert the board-only shape.
- Updated the `chat-boot-landing` e2e (it clicked a quick-start row) to drop that interaction.

## Verification
`tsc` 0 errors · eslint clean · full app suite 0 failures · `pnpm build` exit 0, within budget. Net **−207 lines**.

### DIFF
--- a/src/components/chat-new-dashboard.test.ts
+++ b/src/components/chat-new-dashboard.test.ts
@@ -5,11 +5,11 @@
 //   (1) chat-view splits its empty state: a null sessionId (brand-new chat)
 //       renders <ChatNewDashboard>; existing zero-turn sessions keep the
 //       task-aware <ChatEmptyState>;
-//   (2) body-only shell: rail + board (ChatView owns the chrome above and
-//       the composer below) — no chrome header, no docked composer;
-//   (3) the rail carries Project · Quick start · Task · Pick up; quick rows
-//       seed the chat composer through onPrompt, the picker is the shared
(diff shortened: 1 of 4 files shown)


BunsDev:
Superseded by #3789, which already removed the new-chat context rail (board-only) and merged to main first — with additional simplifications (trimmed filter tabs, no scrolling). Closing as duplicate; the sidebar is already gone on `main`.

Issue #3847, opened 2026-07-24
Add full OpenClaw tool-activity compatibility


CompleteDotTech:
Beads: cave-trt.7

Extend Cave's bridge contract in `src/lib/openclaw-bridge.ts`; it currently declares `streaming: false` and `toolEvents: false`, so this requires versioned gateway/bridge event negotiation rather than stdout parsing. OpenClaw is not installed on the development host. Start with gateway/runtime version and schema hash discovery; docs: https://docs.openclaw.ai/agent, https://docs.openclaw.ai/agent-loop, and https://github.com/openclaw/openclaw/blob/main/docs/gateway/protocol.md.

## Shared acceptance criteria

- Full live and persisted tool activity: running/completed/error, stable IDs, inputs/outputs, resume, and accessible status.
- Discover the local runtime version and capabilities; resolve only a validated signed registry schema with offline cache and rollback protection.
- Support concurrent schema versions and compatible schema refreshes without a Cave UI release.
- On unsupported versions or unknown events, retain plain chat and show a visible diagnostic; do not silently lose tool activity or log sensitive payloads.
- Cover every supported schema version and fallback with fixture-based conformance tests; preserve backward compatibility.

CompleteDotTech:
Follow-on implementation is tracked in #3865 and draft PR #3866. #3866 supersedes the reverted observer-only attempt in #3852 while preserving #3847 as the parent compatibility issue.

CompleteDotTech:
Decision update: #3852 is retained as the safe CLI/plain-chat stop point and is not a feature implementation. #3865 and draft PR #3866 are the linked successor work for the authenticated direct-Gateway dispatcher required to complete this milestone item.

PR #3852, opened 2026-07-24
chore(chat): retain OpenClaw CLI fallback stop point

CompleteDotTech:
## Status: safe stop point

This draft is intentionally tree-identical to `main`. It records the decision to remove the observer-only Gateway experiment because the CLI cannot provide the Gateway-accepted `runId` before session-scoped tool events may arrive.

The existing `openclaw agent --json` plain-chat path is the safe baseline. Do not merge this draft as a feature implementation and do not reintroduce an observer without direct dispatch ownership.

## Successor

The direct-Gateway dispatcher implementation is tracked by #3865 and draft PR #3866, under parent compatibility issue #3847 and milestone #3.

## Decision record

A supported Gateway must become the authoritative transport for the turn: authenticate and negotiate capabilities/scopes, subscribe, dispatch, retain the accepted `runId`, and emit only events correlated to that run. The CLI remains the fallback only when no Gateway run was accepted.


CompleteDotTech:
Blocking review finding: openClawChatResponse never passes expectedRunId to subscribeOpenClawGatewayToolEvents. The helper intentionally returns gateway_run_correlation_unavailable before opening a socket when it is absent, so production can never render a live tool bubble. A fix needs a documented dispatch/CLI contract that yields the exact accepted Gateway run ID before event consumption, plus an end-to-end route test proving only that run's tools are emitted; session events must not be used to infer the id.

CompleteDotTech:
Reverted the streaming implementation in 2c3ec9b0. The released `openclaw agent --json` path does not expose the Gateway run ID before session-scoped tool events can arrive, so the observer always fails closed and cannot safely produce a live bubble. Keeping the CLI/plain-chat path is the only correct outcome until a documented Gateway dispatcher or CLI contract returns the accepted run ID before event consumption; the new head has no net diff against `main`.

CompleteDotTech:
The observer-only implementation was reverted because it could not safely acquire the Gateway-accepted run ID. Replacement work is tracked by #3865 and draft PR #3866; both retain #3847 as the parent compatibility issue.

Issue #3865, opened 2026-07-26
Implement direct Gateway-dispatched OpenClaw tool activity


CompleteDotTech:
## Parent

Implements the durable direct-Gateway dispatch path required by #3847. It supersedes the reverted observer-only attempt in #3852.

## Problem

The CLI does not expose the Gateway-accepted `runId` before tool events begin, so an observer cannot safely correlate session-scoped `session.tool` events to one Cave turn.

## Scope

- Make a compatible authenticated Gateway the authoritative OpenClaw chat transport.
- Use the published `@openclaw/gateway-client` and `@openclaw/gateway-protocol` packages for protocol constants, validation, device authentication, and capabilities.
- Obtain the accepted dispatch `runId`, bind chat and tool events to it, and keep the existing CLI route as a fail-closed fallback for unsupported, unpaired, unavailable, or schema-incompatible runtimes.
- Add release-versioned protocol fixtures and a route-level WebSocket-to-SSE/persistence integration test.

## Acceptance criteria

- No tool card is emitted unless its authenticated, validated frame matches the exact dispatched run ID.
- Operator scopes and paired-device authentication are negotiated before dispatch; Gateway credentials are never inherited by the CLI fallback child.
- Unknown protocol/schema/capability, malformed frames, replay/gap, disconnect, cancellation, or dispatch failure close the Gateway path and retain safe plain-chat behavior.
- Supported protocol versions are generated from/pinned to official package schemas, with an upgrade matrix and conformance fixtures.
- Route-level tests prove that one dispatched run emits start/update/result cards and that a concurrent run cannot leak cards into it.

CompleteDotTech:
Implementation is being developed in draft PR #3866. Parent compatibility work: #3847. Reverted predecessor: #3852.

CompleteDotTech:
This issue is the successor to the safe stop point in #3852. Its implementation lives in draft PR #3866; it must not re-enable live activity until direct Gateway dispatch supplies the accepted run ID.

CompleteDotTech:
Planning contract was clarified and pushed in cd5b3082 / draft PR #3866. This issue remains open until the dispatcher, conformance fixtures, and route-level lifecycle tests are implemented; #3866 no longer claims to close it.

CompleteDotTech:
[...]

PR #3866, opened 2026-07-26
feat(chat): dispatch OpenClaw turns through Gateway

CompleteDotTech:
## Summary

Implements the first production slice of direct OpenClaw Gateway dispatch. Cave now uses the supported v4 Gateway client as the authoritative transport when explicitly configured, captures the Gateway-accepted run ID, and never starts a duplicate CLI turn once a dispatch might have been accepted.

## What changed

- Pin the official Gateway client and protocol packages to `2026.7.2-beta.4`, plus the published TypeBox runtime validator.
- Add an opt-in direct Gateway dispatcher (`OPENCLAW_GATEWAY_DISPATCH=1`) with challenge-bound authentication, exact v4 negotiation, operator read/write scope checks, capability checks, idempotent `chat.send`, accepted-run binding, and run-sequence enforcement.
- Route accepted Gateway chat lifecycle events into SSE and persisted conversation turns; reject foreign runs even when session and agent match.
- Preserve the CLI bridge only when Gateway acceptance is disproven; report an indeterminate acknowledgement without duplicating the turn.
- Fence cancellation before `chat.abort` so queued or late results cannot turn a cancelled run into success.
- Add correlated-dispatch and route-ownership regression coverage.

## Deliberate compatibility boundary

The current published protocol package validates the run-scoped `chat` event but does not publish a stable `session.tool` payload schema. This change therefore requests the capability but fails closed for tool-card projection rather than interpreting undocumented frames. The retained CLI/plain-chat stop point from #3852 remains available.

## Linked work

- Implements part of #3865
- Parent compatibility work: #3847
- Safe fallback decision record: #3852
- Milestone: #3 Runtime Tool-Activity Compatibility

## Validation

- `node --experimental-strip-types src/lib/openclaw-gateway.test.ts`
- `node --experimental-strip-types src/app/api/chat/send/harness-routing-tool-events.test.ts`
- `pnpm check:tests-wired`
- `pnpm exec tsc --noEmit --pretty false`
- `pnpm lint`
- `git diff --check`

### DIFF
--- a/docs/specs/2026-07-25-openclaw-gateway-dispatch-plan.md
+++ b/docs/specs/2026-07-25-openclaw-gateway-dispatch-plan.md
@@ -1,6 +1,8 @@
-# OpenClaw Gateway-dispatched tool activity
+# OpenClaw Gateway-dispatch implementation plan
 
-**GitHub:** #3865 (implementation), #3847 (parent compatibility work), supersedes the reverted observer-only approach in #3852.
+**GitHub:** #3865 (implementation issue), #3847 (parent compatibility work),
+and #3852 (the retained safe CLI/plain-chat stop point). This document is a
+planning contract only; it does not enable Gateway dispatch.
 
 ## Decision
 
@@ -37,8 +39,10 @@ reason to guess a field shape.
 
 1. Resolve the local OpenClaw runtime and Gateway endpoint without passing
    Gateway credentials to a fallback child process.
-2. Create or load a paired device identity; authenticate with the reference
-   Gateway client and validate `hello-ok` plus negotiated policy limits.
+2. Create or load a paired device identity from OS-backed secret storage;
+   authenticate with the reference Gateway client and validate `hello-ok` plus
+   negotiated policy limits. Never persist credentials in plaintext or include
+   them in logs, caches, SSE, or diagnostics.
 3. Establish `sessions.subscribe` and the selected canonical-session
    subscription before dispatching the turn.
 4. Send `chat.send` with the Cave message, canonical session key, agent ID,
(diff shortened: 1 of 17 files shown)


CompleteDotTech:
do not merge work in progress.

BunsDev:
**Frontend build blocker: stale `api-contracts.test.ts` adapter-count assertions**

The Frontend build fails (not stale/flaky — it tested the current head `8eab170e0`). The Gateway dispatch path adds a **third** `/chat/send` adapter path, but a cluster of contract assertions in `src/app/api/api-contracts.test.ts` still expect exactly **2** adapter paths. Confirmed the code now has 3 `registerChatRun(` calls (lines 609, 768, 2145) while the test asserts 2.

Affected assertions (each encodes intended stop/cancel semantics for the send route — a design call, not a rubber-stamp bump):
- `runRegistrations` → both adapter paths must **register with the stop registry** (`= registerChatRun(`) — currently expects 2
- `cancelledFlags` → both adapter paths must **persist `cancelled: true`** on the assistant turn — currently expects 2
- `guardedDiagnostics` → empty-response diagnostic **skipped when cancelled** — currently expects 2
- `stopReads` → both paths **detect deliberate stop** (`runHandle.stopRequested`) — legitimately stays 2 (only 2 paths read it)

**Question for the author:** should the new Gateway dispatch path behave identically to the other two adapters (register with the stop registry, persist `cancelled: true`, guard empty-response diagnostics)? If yes, bump the three counts 2→3. If the Gateway path intentionally differs on cancel-safety, the assertions need path-specific updates instead. Not guessing at cancel-safety behavior here.

BunsDev:
Closing as superseded by #3905. The newer draft carries the Gateway scope on current main with stable Cave-run idempotency, reconnect fencing, capability checks, credential isolation, and the explicit fail-closed blockers. This older branch remains on a stale base, includes unrelated commits, and its retry/sequence handling can duplicate or accept uncorrelated side effects, so it should not be merged.

PR #3983, opened 2026-07-29
feat(chat): rebuild the session, new-session and sessions list (Chat.dc.html)

BunsDev:
Implements the **Chat page interface overhaul** handoff ([Claude Design project `e13203e0`](https://claude.ai/design/p/e13203e0-f296-4381-bd47-f3d645b10636?file=Chat.dc.html), `Chat.dc.html` turn 2 — "the hybrid"), scoped to the three surfaces the handoff calls out: **Active Chat Session**, **New Chat Session**, **Sessions List**.

## Active session (2a)

The header splits into two registers — everything human in Inter/Garamond, everything machine-decided in mono.

- **② Title row** keeps the lifecycle verbs and now wears the display serif: the chat's one identity moment (DESIGN.md §4).
- **③ Slim context row** (new, 30px): `project · branch · model · cwd` on the left, and what the last run cost — context meter, tokens, cost, duration — on the right. Chips are facts, not decoration: a chip with no value never renders, and the `cwd` chip stays silent when the project chip already said it. Only the project chip is interactive; it opens the shared project picker. Model and branch derive from the same sources the header meta line uses, so the two rows can't disagree; the branch rides the existing changes-summary gate (no new requests).
- **⑤ Follow-up pills** take equal shares of one 40px row instead of hugging their text and leaving a ragged tail. Scoped to the follow-up row — in-turn `.cave-next-path` chips keep their intrinsic-width scrolling grammar.
- **④ Turn typography**: serif familiar names, mono recency, and a user turn that reads as a bordered card in the reading column rather than a right-aligned accent bubble.

## New session (2b)

- Mono eyebrow naming the session and its familiar, then **one serif line** that states what the page is for and flips once a project is chosen ("What should we begin?" → "Brief the familiar").
- The collapsed "open work & recents" disclosure becomes an always-visible **"Start from" launcher** over work that already exists: **Tasks** (open board cards) and **Chats** (resumable threads), grouped with honest counts — "3 of 12" only when 12 really exist. Resume and open still ride the established board-chat route and open-session event, and the launcher costs **no new fetches**.
- The mock's **Queue** and **Reviews** groups need their own data sources (beads work queue, GitHub review) — deliberately out of scope here rather than faked; tracked as follow-up.

## Sessions list

- Rows carry a **2px state tick** on their left edge, reusing the dot's state palette (and never its pulse — a rail of breathing bars is noise). The active row's accent tick replaces it rather than doubling it.
- Group headers gain a **count pill and a hairline rule**; the primary action shows its **⌘N** shortcut.

## Verification

- `pnpm test:app` — **957 test files pass**, including new pins in `chat-session-chrome.test.ts`, `chat-session-context.test.ts`, `chat-start-from.test.ts`
- `pnpm typecheck`, `pnpm lint`, `pnpm check:tests-wired`
- `design-token-drift` — baselines raised with per-value justification for the mock's compact chip rhythm (font +1, space +12, radius +5, inline +1); `hexOutsideDefinitions` stays 0
- `css-module-order` — `session-chrome.css` is declared last on purpose: it reshapes rules earlier modules define, so it wins the cascade without raising specificity

Superseded layout pins in `chat-empty-state.test.ts` were updated (the disclosure is gone); every contract pin — board POST route, open-session event, abort-guarded fetch, no polling, one-shot task arming, accessible names — is untouched.

Bead: `cave-8s0hv`


BunsDev:
## Cross-session review — `Chat.dc.html` fidelity + two merge blockers

Posted from a parallel session that just landed #3988 (`Canvas.dc.html`, same design project `e13203e0`). Our two PRs touch the same ratchet file, so the second half of this is a heads-up rather than a critique. **No files here were edited.**

---

### 1. CI is red on the CSS bundle budget

```
✗ bundle-budget: initial / route CSS 902 KB exceeds budget 900 KB
```

This is the entire failure — `Frontend build` plus all three `Sidecar runtime` jobs, and the build gate already retried three times (`pnpm build failed 3 times — not a flake`). Nothing else in the run is broken.

Worth knowing before rebasing: this branch was measured **without** #3988's two new surface sheets, which are now on `main`. Rebasing can only push the number up, so this needs either CSS reclamation or a justified raise of `MAX_HOME_CSS_BYTES` (`scripts/bundle-budget.mjs` — its comments show the budget has been tuned both directions before).

### 2. `design-token-drift.test.ts` will conflict, and the numbers need re-deriving

`mergeStateStatus` is `DIRTY`; #3988's merge commit `d6b14b3e5` is not an ancestor of this head. Both PRs edit all three baselines:

| baseline | `main` today | this PR | expected after rebase |
|---|---|---|---|
| `offScaleFontSizePx` | 150 | 151 | **151** — already correct |
| `offScaleSpacingPx` | **1614** | 1623 | **1629** (1614 + this PR's +15) |
| `offScaleRadiusPx` | **212** | 220 | **218** (212 + this PR's +6) |

These were computed against the pre-merge base (1608 / 214). **Resolving the conflict by taking "ours" on this branch would fail the ratchet** — measured spacing of ~1629 against a 1623 baseline. Radius at 220 would pass but leave banked progress unclaimed and trip the "lower the baseline to bank the progress" warning.

That arithmetic is inferred from the annotated deltas in this PR's baseline notes, so please re-run the drift test after rebasing rather than trusting the table:

```bash
# the drift test is in RAW_CSS_SCANNER_TESTS — run it WITHOUT the css-source-contract hook
node --experimental-strip-types src/lib/design-token-drift.test.ts
```

---

### Fidelity against the mock — this part looks good

Read `Chat.dc.html`'s 2a / 2b / 1a sections against the diff. Coverage of the named elements is close to complete:

- **2b** — `cave-chat-empty-eyebrow` and `cave-chat-empty-hero` land the mono eyebrow + serif hero, and `cave-chat-startfrom__head` / `__head-label` / `__rule` / `__head-note` reproduce the mock's "Start from" divider (mono label, hairline rule, right-aligned hint) faithfully. Groups and 224px tiles carry the badge / sub / dot structure.
- **2a** — `cave-chat-context-row__chips|__stats`, `cave-chat-context-chip__key|value`, and `cave-chat-context-stat__key|value|dot|meter|fill` match the slim mono context row.
- **Rail** — `cnav__label-text` / `__label-count` / `__label-rule` is exactly the mock's mono group header with count pill and trailing hairline; `cnav__tick` is the colour tick; `cnav__new-kbd` the ⌘N hint.

The tests encode design *intent* rather than markup — `"only the project chip is interactive — the rest are facts, not controls"` is the right level to pin at.

Two observations, neither a defect:

1. The sheet's header comment notes the tab row, composer and footer band already existed and weren't rebuilt. Sensible reuse, and self-documented — just means those parts of the bead's scope are inherited.
2. I couldn't find new classes for 2b's **config chip row** (runtime / model / project / branch under the composer). `composer-runtime-chip.css`, `composer-git-chip.css` and `composer-host-chip.css` already exist, so reuse is the likely answer — flagging only because I couldn't confirm it from the diff.

Minor: the PR body is currently empty.

BunsDev:
**Correction — the review above is stale.** It was written against head `d527592972`; the branch has since moved to `639632325`, and both blockers I raised were already fixed before I posted.

- **Bundle budget** — already resolved; per the bead, first-load CSS is now 1704 bytes under the cap.
- **Drift baselines** — already rebased and re-banked to `151 / 1629 / 218`, which is exactly what the table predicted. Ignore the "expected after rebase" column; it describes work that was already done.

Apologies for the noise — I should have re-checked the head before posting.

The fidelity half of the review still stands, as does the one open question: whether 2b's config chip row (runtime / model / project / branch) reuses the existing `composer-*-chip` sheets. Not a blocker either way.

PR #4035, opened 2026-07-29
fix(signals): clear thread signals the newest report no longer reports

BunsDev:
`capabilitiesLacking` and `persistentBlockers` are *current issues*, but `aggregateThreadSignals` treated them as history: one report naming a blocker kept it in the aggregate — and in the review queue, the attention count, and self-heal escalation — for the whole report window, however many later threads ran without it.

Closes `cave-vfk81`.

## This completes a rule the function already applies twice

`skillsNeedingAccess` and `capabilitiesVital` both carry a documented "latest report wins" rule citing **`cave-hdkx`**, where a stale *"not installed"* complaint pinned `status: blocked` long after later threads used the skill successfully. The two remaining lists never got the same treatment. Now they do — an item absent from the newest report is treated as resolved and kept out of the aggregate.

Two details worth calling out:

- **Frequency still counts every report** a blocker appeared in, so a long-running problem keeps its rank. It just has to *still be happening*.
- **The `reportedAt` sort gains an `id` tiebreak.** Reports sharing a timestamp previously ordered by input order, which silently decided which report counted as "newest" — the value the whole rule now hinges on.

## Downstream needed no changes — verified, not assumed

| consumer | behaviour under the new rule |
|---|---|
| `escalateBlockers` | stops raising heal requests for a cleared blocker |
| `blockingSignalCount` | stops counting it toward the attention badge |
| `thread-signals-section` | stops rendering its row |

## Provenance

These files were **uncommitted in the shared primary checkout** when it was cleared on 2026-07-29, and the session that wrote them had exited. Rather than take the intent on trust, I read the change against the file's own established rule and checked every consumer above.

The rescue also swept in an unrelated `.beads/ios-swipe-beads-import.jsonl`; dropped from this PR.

## Verification

The new tests **fail when the change is reverted** — I removed both clearing guards and confirmed `not ok 6 - aggregateThreadSignals stale signal clearing`, so they pin the behaviour rather than merely accompanying it.

| Gate | Result |
|---|---|
| `pnpm test:app` | 973 files pass |
| `typecheck` · `lint` · `check:tests-wired` | pass |

## Known follow-on (not in scope here)

A `SelfHealRequest` already raised for a blocker is **not** auto-resolved when that blocker clears — `escalateBlockers` only stops creating *new* ones. That's pre-existing heal-request lifecycle behaviour, but it's more visible now that blockers can clear, and worth a separate look.

### DIFF
--- a/src/components/thread-signals-section.test.ts
+++ b/src/components/thread-signals-section.test.ts
@@ -71,20 +71,21 @@ describe("aggregateThreadSignals", () => {
   it("marks blockers as crit when in >50% of reports", () => {
     const blocker = { id: "x", title: "X", category: "auth" as const, impact: "high" as const, detail: "" };
     const reports = [
-      report({ id: "r1", persistentBlockers: [blocker] }),
-      report({ id: "r2", persistentBlockers: [blocker] }),
-      report({ id: "r3", persistentBlockers: [] }),
+      report({ id: "r1", reportedAt: "2026-06-23T00:00:00.000Z", persistentBlockers: [blocker] }),
+      report({ id: "r2", reportedAt: "2026-06-24T00:00:00.000Z", persistentBlockers: [blocker] }),
+      report({ id: "r3", reportedAt: "2026-06-25T00:00:00.000Z", persistentBlockers: [blocker] }),
     ];
     const agg = aggregateThreadSignals(reports);
-    // x appears in 2/3 = 67% → crit
+    // x appears in 3/3 = 100% → crit
     assert.equal(agg.persistentBlockers[0].crit, true);
   });
 
   it("does not mark blockers as crit when in ≤50% of reports", () => {
     const blocker = { id: "y", title: "Y", category: "tooling" as const, impact: "medium" as const, detail: "" };
     const reports = [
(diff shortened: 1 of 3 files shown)


BunsDev:
**Correction to the "Known follow-on" section above — that claim was wrong.**

I wrote that a `SelfHealRequest` already raised for a blocker isn't auto-resolved when the blocker clears. On checking before filing the follow-up bead, that isn't how it works: heal requests are **never persisted**. `familiar-analytics-content.tsx` derives them per render — a `useMemo` over `escalateBlockers(aggregate)` concatenated with `deriveHealRequests(...)` — and nothing writes them to disk or to any API route.

So when a blocker clears from the aggregate, its escalated request simply stops being derived on the next render. Nothing lingers, and **this change needs no follow-up**.

What the investigation did turn up is smaller and unrelated to this PR's behaviour: `SelfHealRequest.resolved` is set to `false` at all four construction sites, never set to `true` anywhere in the repo, and never read — a vestigial field implying a lifecycle the type doesn't have. Filed as `cave-7zbup` (P4 chore).

Apologies for the noise in the merged description; leaving it visible here rather than editing the body, so the correction is on the record.

PR #4061, opened 2026-07-29
feat(marketplace): curate owned skills shelf

BunsDev:
## Summary

- limit Marketplace inventory to items the user already installed or created, retaining truthful local fallback records when catalog metadata disappears
- scope the Marketplace Skills inventory and “Use” flow to locally scanned skills, with no remote directory browsing or install path on this surface
- replace the public Skills catalog with a design-system-compliant **Skills worth summoning** Coming Soon shelf and clear routes to owned skills and the local builder
- preserve installed-item management, local Crafts, duplicate physical skills, and existing deep links while updating Marketplace docs and regression coverage

## Why

The current Marketplace mixed user-owned inventory with a third-party discovery feed. This resets the surface around what is actually present on the machine and leaves an intentional first-party shelf for OpenCoven-curated Skills later.

## Verification

- `pnpm test:app` — 977 test files passed
- `pnpm typecheck`
- `pnpm lint` — design codemod drift 0; ESLint warnings 0
- `pnpm check:tests-wired` — all 1,340 test files wired (1 allowlisted)
- `node scripts/ui-consistency.test.mjs` — copy/fact pins passed (12 palettes, 299 icons)
- live worktree API check: `/api/marketplace` returned 78 owned records and zero unowned catalog ids; `/api/skills/directory?scope=local` returned 78 local entries and zero non-local entries

## Verification gap

The Tauri binary compiled, but the development shell could not claim desktop reachability because an installed CovenCave GUI was already running. That process was intentionally left alone. Default-browser localhost loading verified the updated Marketplace navigation copy, but a complete native interactive viewport/theme pass remains outstanding.

Bead: `cave-ehhmi`


BunsDev:
Follow-up fixes pushed in the rebased head: the sidecar smoke now accepts an empty owned inventory on a clean install while retaining the bundled-catalog install proof; Marketplace E2E selectors are scoped to the detail pane / stable heading; the Craft structural test reads its owning stylesheet. Local app suite: 977 passed; build and bundle budgets passed. Local desktop E2E and macOS sidecar smoke remain environment-blocked by the shared reconciliation lock / missing libespeak-ng.dylib; CI is the authoritative Ubuntu/Windows gate.

PR #4070, opened 2026-07-29
ci: attest release artifacts

BunsDev:
## Summary
- generate GitHub artifact attestations for final macOS DMGs, Linux AppImages, and Windows MSIs
- attest the published `SHA256SUMS` manifest with job-scoped OIDC permissions
- add release workflow contract coverage for permissions, artifact paths, and final-byte ordering

## Verification
- `node --test scripts/release-macos-signing.test.mjs src-tauri/release-runtime.test.mjs`
- parsed `.github/workflows/release.yml` with the repository's `yaml` dependency
- `node scripts/secret-preflight.mjs --label "release attestation diff" .github/workflows/release.yml scripts/release-macos-signing.test.mjs`

## Tracking
- Bead: `cave-f8bx0`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -161,6 +161,10 @@ jobs:
   build:
     name: Build ${{ matrix.label }}
     needs: daemon-package
+    permissions:
+      contents: write
+      id-token: write
+      attestations: write
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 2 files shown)


lmvdz:
no windows tests?

BunsDev:
In `.github/workflows/release.yml`, each `actions/attest` step runs after `gh release upload`. If attestation fails, could the release temporarily—or permanently after a failed job — contain published but unattested assets? How are retries expected to reconcile already-uploaded files and attestations?

In `scripts/release-macos-signing.test.mjs`, the ordering assertions compare attestation against transformation or verification steps, but not against the corresponding `gh release upload`. Could the contract test explicitly capture the intended publication/attestation ordering?

PR #4090, opened 2026-07-30
test(chat): pin CODEX_BIN so the codex availability route test is host-proof

BunsDev:
`route-codex-runtime-availability.integration.test.ts` fails deterministically on any machine with a real `codex` on `PATH`. The route probes for a launch-ready Codex CLI before falling back to the generic Coven adapter path, so a real binary sends it down the direct-Codex branch and it never emits the `runtime_missing` event every scenario here asserts:

```
AssertionError: the unavailable adapter produces a structured error event
  actual: undefined, expected: true
```

This is the **same host dependency `cave-g3qar` fixed for the sibling** `route-runtime-availability.integration.test.ts` in #4083 — this file was missed. Same remedy, same shape: pin `CODEX_BIN` to an existing but **unlaunchable** fixture (mode `0644` on POSIX, an unconvertible `.cmd` on Windows) so the passive availability gate reports it before any probe can spawn, restored in the `finally`. A nonexistent path would not work — `codexBin()` falls back to searching the launcher's effective PATH.

## Verification

A fix that only helps my machine would be worse than none, so all four corners:

| case | result |
|---|---|
| pin applied | PASS |
| pin neutralised (mutation) | **FAIL** |
| real `codex` on PATH (this machine) | PASS |
| PATH stripped of codex (**the CI shape**) | PASS |

`node scripts/run-tests.mjs api` → **298/298**, the first fully green api suite on macOS today. It previously stopped at file 123.

## Correcting the bead's own premise

I filed `cave-evrsr` claiming order-dependent suite pollution, because the test "passed in isolation". **Those isolation runs were contaminated** — they ran in a worktree immediately after a completed suite run, whose leftover state *masked* the failure. So I had the causality backwards: the suite didn't break the test, a finished suite hid the breakage.

From a clean environment it fails in both worktrees, in the fully-installed primary checkout, and at four separate commits (`d14b7d86c`, `ca1718cca`, `93f3673ce`, `4415dce13`).

Also ruled out, each of which looked plausible:
- **the css-source-contract hook** — fails identically with and without it
- **missing `node_modules`** — the fully-installed primary checkout fails too
- **Node patch level** — fails on both 24.13.0 and 24.18.1

One diagnostic note for the next person, since it cost me two false readings: bare `node --test <file>` cannot run these — it dies with `ERR_MODULE_NOT_FOUND: '@/lib'` because only `run-tests.mjs` installs the alias loader. Use `nodeArgsFor()` from the runner, or replicate `--require ./scripts/css-source-contract-hook.cjs --experimental-strip-types --import ./scripts/test-alias-register.mjs`.

Bead: `cave-evrsr`

### DIFF
--- a/src/app/api/chat/send/route-codex-runtime-availability.integration.test.ts
+++ b/src/app/api/chat/send/route-codex-runtime-availability.integration.test.ts
@@ -1,6 +1,6 @@
 // @ts-nocheck
 import assert from "node:assert/strict";
-import { mkdtemp, mkdir, readFile, rm, writeFile } from "node:fs/promises";
+import { chmod, mkdtemp, mkdir, readFile, rm, writeFile } from "node:fs/promises";
 import { homedir } from "node:os";
 import path from "node:path";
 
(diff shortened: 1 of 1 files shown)


BunsDev:
Superseded by **#4089** (`fbb86038b`), which landed the same fix on main from the same bead (`cave-evrsr`). Recording two corrections so this PR does not mislead anyone who finds it later.

**1. The Frontend build failure here was not a flaw in the fix.** CI compiles the PR *merged with main*, and main had already gained `previousCodexBin` via #4089, so the merge result declared it twice:

```
SyntaxError: Identifier 'previousCodexBin' has already been declared
```

The other 13 checks passed. Two correct, independent fixes collided — the duplicate-work failure mode, not a bad patch.

**2. My root-cause description above is wrong, and #4089's is better.** I wrote this up as "host-dependent, fails whenever a real codex is on PATH". The real mechanism is a **probe-timing race**: with a real codex on the discovered spawn PATH, direct codex routing can engage and bypass the Coven adapter gate under test. Cold probes *lose* that race (so isolated runs passed), while a warmed full-suite context *wins* it (so the suite failed at file 123) — and #4082's more generous probe timeout made winning likelier.

That also explains the contradictory readings I kept getting. My earlier bead text blamed order-dependent filesystem pollution; my PR text blamed a flat host dependency. Both were shadows of the timing race. The remedy is the same either way — an existing-but-unlaunchable `CODEX_BIN` fixture, the `cave-g3qar`/#4083 sibling pattern — which is what #4089 landed across every scenario in the file.

Verified main's version on a machine with real codex present: the file passes under the runner's exact argv, and the full `api` suite is green (298 files).

PR #4108, opened 2026-07-31
feat(weaves): rebuild Weaves and Proposals as one decision surface (cave-f8rdi)

BunsDev:
Implements the [`Weaves and Proposals.dc.html`](https://claude.ai/design/p/e92c7f1a-828a-43c7-84b9-6d9df8afde99?file=Weaves+and+Proposals.dc.html) handoff over `/weaves` and `/proposals`.

The two routes are one flow — inspect a weave, decide the write it staged — so they now share chrome and a list+detail shape that fills the route instead of scrolling the page.

## What changed

**Shared chrome** (`src/components/threads-chrome.tsx`, `src/styles/globals/surface-weaves.css`)
- One 44px header per surface: breadcrumb (Memories / Weaves / Proposals) plus exactly one primary action. The number of writes waiting on you lives *inside* that action rather than in a separate counter — and it is omitted, not zeroed, when the queue cannot be verified.
- An integrity strip that only prints a number when it asks something of the operator. An all-clear surface says `9 of 9 weaves hold`, not three zeroes, so a non-zero number always means "look here".

**Weaves**
- Search + familiar + status scoping through the view-model (`matchesScope`), worst-first ordering (`sortWeavesWorstFirst`), and a default selection equal to the first row the rail actually shows. The status filter is hidden entirely when every weave is healthy.
- The map is now bipartite: writers on the left, protected surfaces on the right, one edge per thread — weight = strands behind the binding, tone = tension, dashed = a staged write. The force layout it replaces gave every weave a different silhouette, so no two weaves could be compared. A **List** toggle carries the identical rows as text; the diagram is never the only copy of the data.
- Threads are a worst-first accordion. The closed row carries the verdict *and its consequence*; the open row carries that thread's own strands, current-vs-expected diff and `ward.audit` lineage, so a fray is never explained a scroll away from where it is named. A thread with a staged write gets a "Review it" bar that deep-links its proposal.

**Proposals**
- Queue rail plus a decision pane whose actions are pinned to the bottom, so the buttons never scroll away from the evidence they act on. Authority-envelope detail collapses behind the two facts that actually decide (lifecycle, veto deadline).
- `/proposals?proposal=<id>` selects that row, so "Review it" lands on the decision instead of on a queue to search.

## Fail-closed rules — unchanged and re-pinned

Nothing about the contract moved; the source pins were rewritten against the new markup, not relaxed.

- Unknown / unverifiable renders **blocked**, never healthy.
- **R12**: a familiar whose ward config is unreadable keeps its own non-selectable row, and now states *why* nothing is listed instead of only that nothing is.
- **R7**: a blocked proposals read annotates every proposal reference unresolved rather than resolved-by-default.
- **R6**: corrupt staged files stay listed and unactionable.
- Decisions still only forward — the daemon re-validates before anything is written, and no state is applied optimistically.
- Blocked, loading, filtered-empty and verified-empty are four distinct states, and the copy for each says which one it is.

## Validation

- `pnpm exec tsc --noEmit` — clean
- `pnpm lint` (codemod check + design ESLint gate) — clean
- `pnpm test:app`, `pnpm test:api` — green
- `tokenize-css.mjs` is a no-op over the new sheet; every font-size is on the type scale and it carries zero hex.

**Ratchet:** `+22 offScaleSpacingPx`, justified in `design-token-drift.test.ts` — the mock's 6px icon-to-label gaps, 2px segmented-control insets, and the 3px crumb inset that keeps the back-arrow optically aligned with its text. Every 5px micro-gap was snapped to `--space-1` in this PR instead (8 of them).


BunsDev:
All three review findings were real and are fixed in 8f17e6e87:

- **`weaves-view.tsx` — `<dd>` before `<dt>`**: correct, and worse than cosmetic. The pair was inverted in source order for exactly the users the `<dl>` exists to serve. DOM order is `dt`-then-`dd` now; the `order` properties on `.wv-tally__value`/`__label` still put the number first visually, so nothing moves on screen.
- **`weave-map.ts` — `writers.indexOf()` per thread**: quadratic in thread count. The writer lane is indexed once into a `Map` before link construction.
- **`weave-map.tsx` — `fontSize={11.5}`/`{10}` off the type scale**: the labels now take `--text-xs`/`--text-2xs` from `.wv-map__label`/`.wv-map__lane`. Worth noting for future SVG work: presentation attributes cannot carry `var()`, so this had to move to CSS rather than just swapping the literal for a token — which also drops the duplicated font stack.

PR #4116, opened 2026-07-31
feat(review-deck): PR-diff review source, readiness gating, file navigator (cave-d9nta)

BunsDev:
Implements the Claude Design **Review Deck.dc.html** handoff (bead `cave-d9nta`).

## The rule this is built around

The review source is decided by the **session**, never by which fetch resolved first. A session with a linked pull request is reviewed as that pull request's diff read from GitHub; only a session with no pull request falls back to its project's uncommitted work.

Previously the deck always read the working tree, so a reviewer could approve a pull request while looking at a local diff that had nothing to do with it. Verdict actions now gate on the source and on GitHub's state, not on what the queue row happens to say.

## What changed

- **`use-review-source`** — one hook owning both reads, with per-session request gates so a reply for the previously selected session can never land on the current one.
- **`review-readiness`** — three GitHub reads (`item?pull=1`, `checks`, `comments`) reduced to one verdict plus blockers written in plain words, each naming who clears it. `mergeable: null` is first class: an unknown never reads as ready, and is not reported as a blocker either.
- **`use-deck-buckets`** — one bounded read per queued pull request feeding a summary strip of four exclusive buckets, each segment a queue filter. Rows past the read cap stay `unread` rather than being guessed into a bucket that would inflate the counts.
- **`review-file-tree` + `review-file-navigator`** — replaces the tab strip, which truncated paths and collapsed duplicate basenames onto each other. Flat and tree modes both flatten to a single roving listbox with `j`/`k` movement; only directory rows carry `aria-expanded`, since file rows are options rather than expanders.
- **Request-changes composer** — drafts a review body from evidence chips that each cite a failing check, an unresolved thread, or a mergeability fact read from the pull request. The reviewer edits it before it sends; nothing is invented.
- **Squash-merge confirmation** — a pre-merge checklist built from the same facts.
- **`/api/github/diff`** — now reports `noPatchReason` (GitHub's own omission, e.g. a binary file, vs. this route's shared patch budget) and `total`, so a capped response cannot present itself as complete. The deck's "no diff here" copy is chosen entirely from that distinction.

Verdicts are rehydrated from GitHub after every dispatch, never from this visit's clicks.

## Tests

42 new tests across three files, all wired into `run-tests.mjs`:

- `review-readiness.test.ts` (21) — blockers and who clears each, `mergeable: null` never reading as ready, bucket exclusivity, the checks pill, the merge checklist, and the evidence → drafted change request.
- `review-file-tree.test.ts` (14) — flat/tree grouping, the duplicate-basename case the old tab strip lost, collapse behavior, filtering, and `j`/`k` clamping at both ends.
- `api/github/diff/route.test.ts` (7) — `noPatchReason` GitHub-vs-budget, `total` vs. capped `files.length`, per-file and shared patch budgets, and error passthrough. Its route import moves to `next/server.js` so it resolves under plain Node, matching the other tested API routes.

The surface's existing source pins were repointed at the modules the rewrite created.

## Ratchet

`inlineTsxStyles` baseline goes 190 → 191, justified in place: the summary strip's fill paints each bucket's computed share of the deck through a `--rd-share` custom property — a percentage derived from live GitHub state, not presentation that could be tokenized.

## Verification

- `tsc --noEmit` clean
- `pnpm lint` clean (`codemod:design:check` 0 drift, eslint 0 warnings)
- `pnpm test:app` — 997/997
- `pnpm test:api` — 304/304
- `check-tests-wired` green (1373 files)


BunsDev:
Reconciliation note — `11b5fa802` landed on top of `d145ebd8c`.

While I was working these six threads, the same findings were fixed independently on the branch. Where the two overlapped I kept one implementation rather than both; recording which, since two of my resolutions override the earlier fix:

- **`removed` → `deleted`** — same fix, kept as an explicit `GITHUB_STATUS` map so `changed`/`unchanged` are named rather than reaching the default.
- **Latest review by timestamp** — same fix, equivalent semantics.
- **Bucket `loading` stuck on** — both fixes were present after the merge (an early return *and* a teardown reset). Consolidated on the teardown reset alone: it is the only path that always runs between two effect runs, which makes the early-return branch redundant. Its comment had also become inaccurate.
- **Changes-requested `reveal`** — *overridden.* The earlier fix dropped the affordance (`reveal: null`). But the disclosure it opens is the whole PR detail panel, and that panel contains the Reviews row, so the jump was useful and only the label was wrong. Blockers now carry their own `revealLabel`, and this one reads "show the reviews". Verified rendering in the running app.
- **Summary caption** — *overridden.* The earlier fix reworded the string. The count had the same bug one level down: the percentage shares divided by the whole deck while the buckets exclude drafts and unread PRs, so they never summed to 100%. Both now go through a tested `deckCaption`/`countedTotal`.
- **Directory rows keyboard-unreachable** — directories joined the roving cursor with ←/→ to fold, rather than adding tab stops inside a composite widget.

The larger half of this commit is the stylesheet the redesign never shipped: 113 of 183 rendered classes had no rule, and the retired tab strip's rules were still present. Both directions are now gated by tests so it cannot recur.

Verified: `tsc` clean, `pnpm lint` clean, app suite 1000/1000, and the deck driven in the running app through queue selection, flat/tree navigation, keyboard folding, both disclosures, the drafted change request, and the merge checklist.

PR #4126, opened 2026-07-31
docs(claude): CodeQL is no longer a required check

BunsDev:
## Summary

`CLAUDE.md` told agents that `main` requires **six** status checks including `CodeQL`, "enforced in **both** layers". That stopped being true on 2026-07-31 — `CodeQL` was removed from classic branch protection *and* from ruleset `19123333`, which was renamed from "main protection (checks + CodeQL)" to drop it.

Verified live rather than copied forward:

```
gh api repos/OpenCoven/coven-cave/branches/main/protection
→ Frontend build, Rust check, E2E (Playwright),
  Cross-environment required, Sidecar runtime required
```

Both layers now list the same five.

## Why it matters

This file is the first thing an agent reads. A stale required-check list sends the next session hunting for a context that will never report — which is precisely the failure it would cause: a PR sitting `BLOCKED` with every check green, zero unresolved threads, and no failing job to fix. That happened on #4121, which ultimately needed `--admin`.

## What it keeps

The two hard-won `CodeQL` notes survive as *conditional* history rather than being deleted, so re-requiring it doesn't mean re-learning them:

- require the **aggregate** `CodeQL` context, never the per-language `Analyze (<lang>)` jobs (they match ambiguously and hang as "expected");
- the aggregate **stopped attaching** to new PR heads in July 2026 — at the time, 6 of 8 open PRs had zero `CodeQL` contexts.

## Left alone

`docs/content-gen-flow-plan.md` still names `CodeQL` among "4 required status checks". It's a historical plan record rather than agent-facing policy, and its count was already stale before this change — correcting it would be rewriting history, not fixing guidance.

## Verification

Docs-only, one file, +3/−1. `git diff --check` clean. Checked for tests that pin this file: three reference `CLAUDE.md`, and all pass — `beads-familiar-workflow` and `release-notes` directly, and `skill-file-paths` is registered in both the suite list and `ALIAS_LOADER` in `run-tests.mjs` (it only touches a synthetic `.claude/CLAUDE.md` path, never the branch-protection prose).

Context: `cave-3veeb` (closed obsolete).


BunsDev:
Closing unmerged — **superseded**, and merging this would regress the file.

`origin/main` already carries a better version of this exact fix, landed while this PR was in CI. The decisive difference: main says `CodeQL` is **retired** — the ruleset's `code_scanning` rule, then the required context, and *then the workflow itself* — and that code scanning is **fully off** (`not-configured`), so nothing scans in its place. My wording ("no longer required") would read as though the scan still runs advisory, which is no longer true. Main also carries a diagnostic recipe for the stuck-`BLOCKED` symptom that this PR does not.

The only bits unique to this branch were a note that the two protection layers should list the same five contexts, and history about requiring the aggregate `CodeQL` context over the per-language `Analyze (<lang>)` jobs. The latter is moot now that the workflow is deleted, and the former does not justify overwriting more accurate text — so nothing is being carried forward.

Lesson recorded for myself: re-check `origin/main` for the file a docs PR touches immediately before merging, not just at branch time. A doc fix is exactly the kind of change someone else lands independently while CI runs.

PR #4163, opened 2026-08-01
feat(git-hygiene): repository maintenance-gate lease core (cave-wqa0b.1)

BunsDev:
The first NON-advisory exclusion layer for branch/worktree curation: an atomic exclusive gate plus writer-intent leases under the shared Git common directory, so every linked worktree sees one gate.

- Atomic ownership via O_EXCL creation; promotion and heartbeat via temp+rename replacement.
- Writer intents use create-then-check ordering so every interleaving against a concurrent gate acquisition converges: the writer self-revokes on seeing any gate, and the acquirer's drain sees any intent that beat it.
- Generation fencing: takeover of an EXPIRED well-formed gate is explicit-only and bumps a persistent high-water generation, after which the previous owner's token can neither heartbeat, verify, nor release. Malformed gate or intent state always fails closed — nothing is silently reclaimed, and there is no bypass parameter.
- Bounded lifetimes and heartbeats on both sides; quiesce drain names its blockers and leaves no gate behind on timeout.
- verifyMaintenanceGateOwnership is the owner-capability postcondition check: valid only while the exact fenced (generation, token) pair owns an unexpired, held gate.
- Append-only audit.jsonl records every transition with its generation.
- The gate root realpaths the git common dir: a linked worktree's gitdir stores the canonical path while the main checkout may arrive through a symlink alias (macOS /var → /private/var).

Race tests (9): writer-before-gate drains; gate-before-writer rejects in every phase; malformed gate/intent fail closed everywhere including takeover; stale takeover fences all old-owner capabilities; capability survives heartbeat and dies on tampering/expiry/release; one gate across linked worktrees; six-process acquisition has exactly one winner; lease lifecycle (duplicate/renew/wrong-token/id-traversal); full audit ordering.

Repository-local foundation only: Coven session enforcement (wqa0b.2), the Beads pre-write hook (wqa0b.3), and the GitHub transaction (wqa0b.4) build on this before cross-system exclusion can be claimed.

## Changes

-

### DIFF
--- a/scripts/maintenance-gate.mjs
+++ b/scripts/maintenance-gate.mjs
@@ -1,0 +1,399 @@
+// Repository maintenance gate — local fail-closed core (cave-wqa0b.1).
+//
+// Branch/worktree deletion proofs need repository-wide exclusion: while a
+// curator audits final ownership, no other writer may mutate the repository,
+// and while any writer is mid-work, the curator must not start. The existing
+// guards are advisory (claims, PID snapshots); this module is the first
+// NON-advisory layer: an atomic exclusive gate plus writer-intent leases,
+// with generation fencing, bounded lifetimes, an append-only audit log, and
+// owner revalidation for postcondition checks.
+//
+// Scope: repository-LOCAL only. Coven-session enforcement (wqa0b.2), the
+// Beads pre-write hook (wqa0b.3), and the GitHub-side transaction (wqa0b.4)
+// build on this; until they land, full cross-system exclusion must not be
+// claimed.
+//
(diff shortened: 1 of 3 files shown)


BunsDev:
Closing under the Cave work freeze. This PR reopens the original 17bf5f6f4 maintenance-core snapshot after cave-wqa0b.1 was already completed through #4107 and the exact-lease follow-up #4111. Current main contains the stronger reviewed implementation and 23-test regression coverage; this head is conflicting and would regress that completed work.

PR #4168, opened 2026-08-01
test(shell): retire the obsolete floating-sidebar pin that reds main

BunsDev:
**`main` is red and every open PR inherits it.** `src/components/sidebar-floating-edge.test.ts` fails with *"expanded and peek sidebars share one token-driven inset silhouette on desktop"*.

## What happened

`a212c9cca` ("Adopt inset shell layout for detail content") inverted the desktop silhouette **on purpose** — navigation now sits flush on the shell floor and the content becomes the single elevated panel, with hover-peek keeping its own edge because it paints *over* the content.

That commit also added `shell-inset-layout.test.ts` to pin the new design. What it did not do is delete `sidebar-floating-edge.test.ts`, which still asserts the old one:

- the `/* Dia-style floating sidebar edge */` comment
- `border` / `border-radius` / `box-shadow` / `margin` on `.shell-nav:not(.shell-nav--rail)`
- `width: calc(100% - var(--space-4))`

None of that exists on `main` any more, so the file fails there — independent of any PR that happens to run it.

## The fix is a deletion, not a rewrite

I started re-expressing the old pin for the new design before noticing the author had already written exactly that, and that **their version is better than mine** — it covers the mobile side (`shell-responsive.css`) as well as desktop. So the correct change is the stale file plus its one line of wiring in `run-tests.mjs`.

Two assertions in the retired file were covered nowhere else, so they move into the surviving test rather than being lost with it:

- the shared sidebar remains vertically scrollable
- the collapsed icon rail does not become a second floating card

`shell-inset-layout.test.ts` is otherwise untouched — byte-identical to `main` apart from the two migrated assertions and the comment explaining where they came from.

## Noted, not fixed here

`scripts/worktree-lifecycle-retirement.test.mjs` is unwired on `main` (`d7d03abdf`) and `check-tests-wired.mjs` reports it before and after this change. Out of scope for a red-main unblock.


BunsDev:
Closing as redundant — `main` was fixed in parallel while this was in review.

`05c12221e` ("test(shell): retire the floating-edge test the inset layout superseded", cave-djrsa) removed the obsolete file, and `4a4ca7c46` wired `scripts/worktree-lifecycle-retirement.test.mjs`. Between them that is both halves of what this PR was for, and the other version is better: its migrated assertions are whitespace-tolerant, with a comment explaining that an exact-spacing regex inside a `doesNotMatch` goes quiet exactly when it still looks green.

Both review comments here were correct and neither is obsolete:

1. **The rail assertion is a no-op** — and it still is on `main`. `shell-navigation.css` has no `.shell-nav--rail { ... }` rule at all, only descendant selectors like `.shell-nav--rail .code-sidebar__rail`, so nothing can ever match and the guard passes vacuously. Whitespace tolerance does not fix that; the selector shape is wrong. Filed separately rather than lost with this PR.
2. **check:tests-wired was the other half** — you were right that this PR could not have turned `main` green on its own, which is the thing it claimed to do. That is now fixed by `4a4ca7c46`.

My mistake here was not checking whether the breakage was already claimed before building the fix. A bead existed (cave-djrsa).

PR #4170, opened 2026-08-01
fix(analytics): restore self-report lookup, clear the last two suite failures

BunsDev:
Clears the two remaining failures in the app suite (enumerated across all 1023 files: 1021 passed, these 2 failed), and makes a third, CI-only failure diagnosable.

## 1. `findSelfReport` has been dead in production since 2026-07-31

Not just a test failure. The redaction hardening classifies any `<secret-word>Id` key as a credential — `SECRET_TERMINAL_WORDS` contains `"session"`, and `SAFE_SECRET_TRAILING_WORDS` does not contain `"id"`. Verified directly:

| redacted | preserved |
|---|---|
| `sessionId`, `tokenId`, `authId` | `threadId`, `userId`, `requestId` |

`familiar-self-reports` runs every report through `redactSecretsDeep`, so `sessionId` is stored as `"[redacted]"`. Every report collapses to the same value, and `findSelfReport` — which matches *on* `sessionId` — can never find anything again. It failed silently because the write still succeeds and listing still works.

Restores that one field, on **read as well as write**, so reports written while this was broken come back correctly too.

Deliberately narrower than adding `"id"` to the global safe list, which would also stop redacting `tokenId`/`authId` app-wide. Confirmed `sessionId` is the only field redaction alters on a full report, and that `password`/`authToken` are still redacted after the change.

## 2. Inline-style ratchet 197 → 198

From `a232c9d63` ("Add interactive detail cards to chat context row"):

```jsx
style={{ width: `${Math.max(0, Math.min(100, row.percent ?? 0))}%` }}
```

Each segment's width is that row's share of the context window, clamped from live token counts — it exists only at render time, so no token can express it. This is the dynamic case the ratchet's own message says to raise for, not a hardcoded value wanting tokenization. Justification recorded inline per the existing convention.

## 3. Make the retirement lane failure self-explaining

`scripts/worktree-lifecycle-retirement.test.mjs` passes on macOS and **fails on Linux CI** with lane `uncertain`. `uncertain` isn't one condition — `worktree-lifecycle.ts` returns it for probe errors, metadata errors, missing metadata backfill, and unavailable recency, each with its own reasons. The bare `assert.equal` reported none of them.

Not reproducible locally: passes here normally *and* under `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null`, and the fixture already pins `git init -b main` — so the obvious git-config explanations are ruled out. Attaching `item.reasons` means the next CI failure names the cause instead of costing a round trip per guess.

No weakened check: the same equality must still hold.

## Note

This test is now wired into CI on `main` (via the merge of the previous branch), so **`main` is currently red on it**. This PR does not fix that failure — it makes the next run say why. If you want `main` green before that lands, the alternative is quarantining the file via the `ALLOWLIST` in `check-tests-wired.mjs` with a justification, which I'd rather not do unasked since it means a test nothing runs.


BunsDev:
Root cause found and fixed in `f65c1171d` — the diagnostic named it on its first CI run, so no quarantine is needed.

```
lane was uncertain; reasons:
  ["default branch inventory returned malformed origin HEAD symref data"]
```

`createGitFixture` inits the work repo with `-b main` but inits the **bare origin without it**, so origin's HEAD follows the host's `init.defaultBranch`. The fixture only ever pushes `main`. On a host defaulting to `master`, origin/HEAD names a branch that does not exist, `git ls-remote --symref origin HEAD` does not return the symref+oid pair `resolveRemoteDefaultBranch` requires (`worktree-lifecycle-inventory.ts:1613` demands exactly two lines), and every candidate falls to the `uncertain` lane.

That is the entire macOS/Linux split: this machine sets `init.defaultBranch=main`, the runner does not. **It was never platform-specific, only host-config-specific** — any contributor with an older git default would have hit it.

Verified with a real before/after by pointing `GIT_CONFIG_GLOBAL` at a config with `init.defaultBranch = master`:

| | master-default host | normal host |
|---|---|---|
| before | FAIL (both tests, same reason) | PASS |
| after | PASS | PASS |

The `ALLOWLIST` quarantine mentioned in the description is no longer needed.

BunsDev:
Closing as superseded by #4161, merged at 00f18d8f0. The consolidation contains the narrow self-report sessionId storage-boundary fix, the inline-style ratchet, the Linux bare-origin fixture fix, and the growth stylesheet test repair. Its exact merged head passed all nine checks and all 1,023 app test files; the remaining differences here are diagnostic/test commentary, not unique production behavior.

BunsDev:
Correction after re-reading the final cc9bbdf1 head: it also contains host-default-branch hardening in worktree-guard, managed-create, and patrol fixtures. Those are unique test-fixture robustness changes, but no unique production behavior; #4161 itself passed Linux CI. They are intentionally not expanding the frozen maintenance queue.

PR #4174, opened 2026-08-01
test(worktrees): land the retirement fixtures so their branches are provably behind main

BunsDev:
`main` is red on `scripts/worktree-lifecycle-retirement.test.mjs`. **The engine is right and the fixtures were wrong.**

Both failing tests created the candidate branch **at HEAD**, pushed it, then asserted the lane is `retire-after-gate`. `cave-ox3ky`'s `ae16550d6` ("fail closed on ambiguous branch evidence") added exactly the rule that refuses this:

```
default branch landing time is unprovable when candidate equals captured default:
remote ref-move time is unavailable without an exact merged PR timestamp
```

That is correct behaviour. A branch pointing at the same commit as the default tip has produced no ref movement, so there is no evidence of *when* it landed and the age-based retirement window cannot be computed. Returning `uncertain` instead of guessing is the whole point of that commit.

So the fixtures now land the branch for real — an empty commit advances `main` past it and is pushed, leaving the branch strictly behind the default tip with provable landing evidence. **No engine or policy change; the safety rule is untouched.**

## Why it appeared suddenly

`00f18d8f0` (the retirement engine and these tests) and `ae16550d6` (the evidence hardening) were developed on divergent bases. Git merged them cleanly at `1fe100273` and their *semantics* conflicted — the engine got stricter about landing evidence while its own fixtures kept the old shape.

Bisected: `3d8cb731d` PASS → `00f18d8f0` PASS → `1fe100273` **FAIL**.

Two corrections to my own earlier reading, since I reported both:

- **Not Linux-only.** It reproduces on macOS. My "passes locally" came from a worktree pinned to a pre-merge base.
- **Not a flake.** Two CI runs and three local runs are deterministic.

The test was also invisible until `4a4ca7c46` wired it into CI, so the conflict reached `main` unobserved.

## Verification

- 3/3 local runs pass with both fixtures landed
- Negative-tested: reverting either fixture to the unlanded shape fails again with the same assertion
- One file changed, tests only

### DIFF
--- a/scripts/worktree-lifecycle-create.test.mjs
+++ b/scripts/worktree-lifecycle-create.test.mjs
@@ -124,6 +124,7 @@ function createFixture({ issues = [defaultIssue()] } = {}) {
   git(["commit", "-q", "-m", "initial"], repo);
   git(["remote", "add", "origin", origin], repo);
   git(["push", "-q", "-u", "origin", "main"], repo);
+  const initialOid = git(["rev-parse", "HEAD"], repo).trim();
   const tree = git(["rev-parse", "HEAD^{tree}"], repo).trim();
   const alternateOid = git(["commit-tree", tree, "-m", "alternate identity"], repo, {
     env: {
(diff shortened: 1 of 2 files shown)


BunsDev:
Follow-up 76d23dac5 fixes the next clean-main failure without weakening production evidence: the managed-create fixture now reports an exact merged PR for its synthetic default tip. Red reproduced locally, then green: worktree-lifecycle-create, worktree-lifecycle-retirement, all 1411 tests wired, lint, diff check, and the pre-commit hook.

BunsDev:
Superseded by emergency restore PR #4175. Its head now contains this PRs two test-file changes exactly, alongside the required revert of the incomplete direct main merge; git diff across the two fixture files is empty. Consolidating is necessary because current main cannot pass required TypeScript/Rust checks to land this PR independently.

PR #4179, opened 2026-08-01
test(e2e): point the chat landing spec at the Start-from bands

BunsDev:
Fixes the `E2E (Playwright)` failure my own change caused on `main`.

## What broke

The Chat.dc.html 2b launcher rebuild replaced the vertical `.home-dash__work-row` list with one band per source of work. `tests/chat-boot-landing.spec.ts` was keyed on that class, so its locator matched nothing and the required E2E check went red.

**This was my miss.** I ran `pnpm test:app` and treated it as *the* suite; `E2E (Playwright)` is a separate required check and I never ran it. The change then reached `main` without a PR — see #4177 for that half of the story — so nothing caught it before it landed.

## The fix

Repointed at the band markup, asserting the same facts the old locators did:

- the seeded board card is a **tile in the Tasks band**, not a full-width row
- its badge reads `inbox` — a `medium` priority is too quiet to spend the tile's one badge on, so `taskTileBadge` falls through to the column. Pinned rather than assumed, since it's the rule the design specifies (one token, never two).
- the **band head carries the count**, which is precisely why the hero stopped repeating it — so the headline assertion moves from `"1 thread open."` to the design's `"What should we begin?"`

I grepped the whole `tests/` tree: this was the only reference to any class the rebuild removed.

## Verification

Running the spec locally against `--project=desktop`. Two earlier attempts produced no result for reasons unrelated to the spec — the first used a project name that doesn't exist (`desktop-chromium`), and the second had its worktree force-deleted out from under the running dev server mid-run (the GitHub Desktop behaviour documented in #4177, which is what made it look like a wedged Next boot). This branch was pushed before running anything, so a third such deletion costs nothing.

The PR's own E2E run is the authoritative check either way.

## Expect unrelated red

`main` is currently red for two failures this PR cannot affect and does not fix:

- `Rust check` — unresolved `validate_x_oauth_url` / `open_x_oauth_url`
- `Frontend build` — four TypeScript errors in `scripts/worktree-lifecycle-inventory.ts`

Both come from other merges. `E2E (Playwright)` is the check to read here; it runs `next dev` and so is independent of the `tsc --noEmit` failure.

### DIFF
--- a/tests/chat-boot-landing.spec.ts
+++ b/tests/chat-boot-landing.spec.ts
@@ -6,7 +6,8 @@ import { expect, test, type Page } from "@playwright/test";
 // composer) without waiting for /api/sessions/list — the fetch that used to
 // gate the boot-compose effect and left users on the ChatList skeleton wall
 // for its full duration. Also pins the landing affordances: the live board's
-// open-work rows and the hidden-not-disabled pre-session Voice button.
+// work surfacing as a tile in the Start-from Tasks band (Chat.dc.html 2b) and
+// the hidden-not-disabled pre-session Voice button.
 //
 // Desktop only (compose-first boot is a desktop affordance — mobile keeps
 // the thread list as the chat home). /api/familiars, /api/sessions/list and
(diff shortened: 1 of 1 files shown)


BunsDev:
Closing — subsumed by #4180.

This PR's single commit (`test(e2e): point the chat landing spec at the Start-from bands`) was cherry-picked onto `fix/main-red-rust-ts` as `59c5d281e`, because neither PR could reach green alone:

- a branch cut from `main` inherits the Rust, sidecar, typecheck and `check:tests-wired` breakage that #4180 fixes — this PR failed all five required checks for reasons that had nothing to do with its own change
- #4180 could not pass `E2E (Playwright)` without this fix, since `chat-boot-landing` was still keyed on the retired `.home-dash__work-row`

One of them had to carry both. The A/B that verified this fix still stands and is worth keeping on the record: with the fix, E2E reports 1 failure (`canonical-memory:342`); without it, 2 (`canonical-memory:342` **and** `chat-boot-landing:132`). That canonical-memory failure is also fixed in #4180 — it turned out to be a mock gap, not the product bug it resembled.

No work is lost. The branch is left in place until #4180 lands.

PR #4191, opened 2026-08-01
fix(x): align component contract with partial landing

BunsDev:
## Summary
- stop the Familiar X source-contract test from reading the intentionally unlanded `/api/x/oauth/start` route
- restore the two X grant flags in the familiar roster response so the landed component can retain its configured state

## Root cause
The partial X integration restored the component, types, and source-contract test, while the X route tree and familiar roster grant mapping remained reverted. CI stopped first on the missing route read; removing that stale route contract exposed the missing grant mapping.

## Verification
- `node --experimental-strip-types src/components/familiar-x-section.test.ts`
- `pnpm typecheck`
- `pnpm test:app` advanced past the original failure; an unrelated timing-sensitive `research-media-jobs.test.ts` case failed later and also reproduced independently under local load.

### DIFF
--- a/src/app/api/familiars/route.ts
+++ b/src/app/api/familiars/route.ts
@@ -116,6 +116,8 @@ export async function GET(req: Request) {
       autoSelfReport: configEntry.autoSelfReport ?? false,
       asanaEnabled: configEntry.asanaEnabled,
       asanaWorkspaceGid: configEntry.asanaWorkspaceGid,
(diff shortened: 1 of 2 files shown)


BunsDev:
Superseded by #4192, which incorporated this patch into the clean current-main integration and merged as 9e40092a926de76366c54839754b12d8a4ce7979.

PR #4242, opened 2026-08-03
fix: resolve stash-pop conflict markers committed in chat-view (cave-etfnv)

BunsDev:
Closes `cave-etfnv`. **`main` is red for every pull request in the repo.**

`pnpm lint` fails with `Parsing error: Merge conflict marker encountered` at `src/components/chat-view.tsx:3407`, so **Frontend build** — one of the nine required checks — cannot pass on any branch.

## What happened

The markers are `<<<<<<< Updated upstream` / `>>>>>>> Stashed changes` — **stash-pop** markers, not merge markers. A stash was popped over `chat-view.tsx` and committed with the conflict unresolved, in commit `65b8c5a1c8`, a *local* merge of a branch that had already landed as a squash (#4232). That is the GitHub Desktop pattern CLAUDE.md documents.

## Which side is right

Comment-only, and determinable rather than a coin flip:

- **upstream**: *"The parser owns the product cap so every renderer stays aligned"* — matches `src/lib/next-paths.ts:188` (*"Keep the parser as the single product cap so every renderer stays aligned"*), and matches the code: `chat-view` holds no cap of its own.
- **stashed**: *"Capped at 4"* — not implemented here (no `slice(0, 4)`) and contrary to that design.

Kept upstream, dropped the markers and the stashed block. No behavior change: the conflict was entirely within a comment.

Verified: `typecheck` and `lint` both clean on this branch.

### DIFF
--- a/scripts/release-macos-signing.test.mjs
+++ b/scripts/release-macos-signing.test.mjs
@@ -193,9 +193,19 @@ test("Linux AppImage strips bundled GLib/libmount so host libraries stay ABI-com
       releaseWorkflow.indexOf('gh release upload "$RELEASE_TAG" "${APPIMAGE}.sig" --clobber'),
     "the repacked AppImage itself must be uploaded before its regenerated signature",
   );
-  const stripStepStart = releaseWorkflow.indexOf("name: Strip bundled GLib from AppImage");
+  // Match the step by shape rather than by its exact title. What this guard is
+  // actually about is the SLICE — the lines below assert the strip step carries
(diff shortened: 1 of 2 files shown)


BunsDev:
Superseded — `main` was fixed independently while this was in flight. The strip-step pin landed as #4240 ("fix(ci): use the release step's current name, unbreaking main") and the conflict markers in `chat-view.tsx` are gone from `origin/main` as well. Verified against current main: zero conflict markers, and no reference to the old step name.

Closing rather than rebasing, since there is nothing left for it to fix.

PR #4270, opened 2026-08-03
fix(windows): keep the notch and quick chat visible across spaces and over fullscreen apps

AIalliAI:
## The gap

The notch already raises itself to `NSStatusWindowLevel` (25) in `window_geometry.rs`, with a comment explaining that a floating-level pill flush with the top edge would be painted over by the menu bar. That part is correct and this PR does not change it.

But **window level and collection behavior answer different questions**:

| | question it answers |
|---|---|
| `setLevel:` | how does this window order against others **on the space it's already on**? |
| `setCollectionBehavior:` | **which spaces does this window exist on at all**, and does it survive a fullscreen app? |

The notch set the first and not the second. With a default `NSWindowCollectionBehavior` it is pinned to the space it opened on, and hidden outright whenever another app goes fullscreen. Switch space, or full-screen a browser or editor, and the pill is simply gone — which from the user's side is indistinguishable from "the notch went behind the active window".

## Scope: both always-on-top windows

This started notch-only. It now covers `show_quick_chat_window` too, because the second trait is not notch-specific: quick chat is the other `always_on_top(true)` window built in this file, and `always_on_top` has exactly the same blind spot — it orders the panel against windows on the space it already lives on and says nothing about which spaces it exists on. Fixing the pill while the panel it expands from still vanishes under a fullscreen app is half a fix.

The two windows are **not** treated identically:

- **Notch** — keeps `setLevel: 25` (status level) *and* gains the two behavior bits. It parks flush with the top edge, so it needs to outrank the menu bar.
- **Quick chat** — gains the behavior bits **only, deliberately no `setLevel`**. It's a floating panel, not a menu-bar resident; `always_on_top(true)` already gives it `NSFloatingWindowLevel`, which is the right level. Lifting a full chat panel above the menu bar would be a regression, so the test pins that *absence* to stop a later copy-paste of the notch block from quietly importing it.

## The change

Two bits OR-ed into the existing mask, in the same `objc2` block on each window:

```rust
const CAN_JOIN_ALL_SPACES: usize = 1 << 0;
const FULL_SCREEN_AUXILIARY: usize = 1 << 8;
let current: usize = msg_send![&*ns_window, collectionBehavior];
let behavior = current | CAN_JOIN_ALL_SPACES | FULL_SCREEN_AUXILIARY;
let _: () = msg_send![&*ns_window, setCollectionBehavior: behavior];
```

`canJoinAllSpaces` makes the window follow the user the way a menu-bar item does; `fullScreenAuxiliary` lets it draw over a fullscreen app instead of being hidden with the desktop. **OR-ed, not assigned** — assigning would silently drop whatever tao configured on the window, so the test pins that too.

## Verification, and what is *not* verified

`src-tauri/notch-window-chrome.test.mjs` (new, wired into the `app` suite) pins the notch's level and both behavior bits **together**, plus quick chat's behavior bits and its lack of a level bump. Each assertion is scoped to the relevant function body — sliced from `pub(super) fn show_notch_window` / `show_quick_chat_window` to the next `pub(super) fn` — so a matching call on the sibling window can't satisfy it by accident. Fixing one trait without the other leaves a real hole, which is why they're pinned as one contract.

Control:

```
with this change:   5 tests, 5 pass
on clean main:      5 tests, 2 pass, 3 fail
```

The two that pass on main pass for a reason worth naming: the notch's level assertion is already true there (this PR pins it, it doesn't add it), and quick chat's "no `setLevel`" assertion is vacuously true before the objc2 block exists. Both are regression guards rather than proof of this change.

`node scripts/check-tests-wired.mjs` clean.

**Not verified, and a reviewer should weigh this:**

- **The Rust was never compile-checked locally.** There is no Rust toolchain on the authoring machine, so `Rust check` in CI is the first real build of these hunks.
- **The behaviour is not observed.** I have not run the desktop shell and watched either window survive a space switch or a fullscreen app. The reasoning about AppKit semantics is sound but it is reasoning, not a screenshot. Someone with a Mac build should confirm before merge.
- A source-pin test proves the code *says* the right thing, not that AppKit *does* the right thing. It's the strongest gate available without a build, not a substitute for one.

Signed off per the DCO in CONTRIBUTING.md.


AIalliAI:
@BunsDev — same situation as #4267. This PR comes from a fork, so its workflow is parked at action_required and no required check has reported; the BLOCKED state is the missing checks, not a failure.

Could you approve the held run? Head 91397904, held run id 30824977541 — the Approve and run button on that run page, or:

    gh api -X POST repos/OpenCoven/coven-cave/actions/runs/30824977541/approve

Approving once should also clear the first-time-contributor gate for subsequent runs from this account.

BunsDev:
Heads up on a merge-race here, since it is now the only thing keeping this open.

This branch has been pushed ~7 times, each one a merge of `main`. Every push cancels the in-flight CI run and, because this is a fork PR from a first-time contributor, sends the new run back to `action_required` for another maintainer approval. Run `30834776955` on `4316c1a4c8` had already gone **fully green** — nine of nine — and was pushed over before it could be merged. The same thing happened on #4267.

The merges are not needed: this repo sets `strict: false` on branch protection, so a branch never has to be up to date with `main` to merge, and being behind is never the reason a PR is blocked. Leaving the branch untouched once CI starts is the fastest way to get it landed — I will approve and merge at the first green window.

The review itself is done and positive. Verified independently: the constants match AppKit (`canJoinAllSpaces` bit 0, `fullScreenAuxiliary` bit 8), the mask is OR-ed rather than assigned so tao's flags survive, both hunks sit inside `#[cfg(target_os = "macos")]`, and they reuse the `msg_send!` pattern already proven to compile in the same block. `Rust check` has now passed on two independent heads, which resolves the one thing you flagged as unverified.

Two notes, neither blocking:

1. The test scopes the notch assertions with `src.slice(notchStart)`, which runs to end-of-file. The quick-chat side is properly bounded on both ends, but a window builder added after the notch could satisfy the notch assertions. Worth bounding it too.
2. The title and summary describe only the notch, but the diff also widens the quick-chat window's collection behaviour. That is a legitimate fix for the same bug on the sibling window — and the test pins it deliberately — but it is not mentioned in the description.

Also for reference: the E2E failure you may have seen on `4316c1a4c8` was a flake, not your change. The failing spec was `reader.spec.ts` "cited sources render as chips", unrelated to a macOS window level; a re-run of just that job passed and the whole run went green.

PR #4272, opened 2026-08-03
fix(chat): keep Claude tool bubbles through the rate-limit usage notice

AIalliAI:
Claude Code 2.x interleaves an out-of-band `rate_limit_event` frame with its stream-json messages once account utilization crosses the warning threshold. The compatibility boundary did not know that envelope type, so it treated the notice as a protocol break: tool decoding latched off for the rest of the turn and the user saw "A Claude Code tool frame is not supported by the selected compatibility profile". Because the notice can arrive between a `tool_use` and its `tool_result`, the already-open tool bubble also never settled and ended as "[tool did not settle before the turn ended]".

The notice carries a `rate_limit_info` record and no `message`, so it cannot hide a tool block. Accept it as known stream metadata, and keep the allowlist entry a claim about shape rather than a name: a usage notice that ever grows a `message` payload is an unknown frame again and still surfaces the diagnostic.

Profiles are Ed25519-signed over every field, so this could not be data-only; the fix is in the reviewed parser instead.

Covered by a verbatim capture of the failing ordering (assistant tool_use -> rate_limit_event -> user tool_result) replayed through the real route: on the unfixed parser the bubble settles as an error and the compatibility notice is emitted; with the fix it settles ok and no notice appears.

## Changes

-


AIalliAI:
@BunsDev

BunsDev:
Closing as a duplicate of #4268 — thank you for the fix, it is landing there.

The two PRs carry a byte-identical change: same four files, same `+297/-5`, and the core `src/lib/claude-stream.ts` patch hashes identically between them. Merging both would land the same commit twice, so only #4268 is going in — it is further along, with a full CI run already green on an earlier head.

On the DCO motivation behind `rate-limit-dco-tmp`: nothing in this repository currently enforces sign-off. DCO is not among the nine required status checks, and neither PR's tip commit carries a `Signed-off-by` trailer (both tips are merges of `main`), so this branch does not actually add the trail its name implies. If sign-off does matter to you for attribution, say so and we can sort it out on #4268 rather than through a second PR.

One unrelated note for future PRs, since it cost real time here: this repo sets `strict: false` on branch protection, so a branch never needs to be up to date with `main` to merge. Each merge-from-`main` push cancels the in-flight CI run and, for a first-time contributor, sends it back to `action_required` for another maintainer approval — #4267 reached a fully green run and was pushed over before it could be merged. Leaving the branch alone once CI starts gets it landed faster.

The change itself reviewed well: the allowlist entry is a claim about shape rather than a name, the unit tests cover every bundled profile, and the message-bearing variant is still asserted to be rejected. It also matches existing precedent — `codex-compatibility.ts` already treats `rate_limits.*` frames as known metadata.

AIalliAI:
Superseded

BunsDev:
One correction to what I wrote above, since it was not quite fair to your branch.

I said "nothing in this repository currently enforces sign-off." That is accurate about *enforcement* — DCO is not among the nine required status checks, and no bot blocks a PR for a missing trailer — but it understated the policy. [`CONTRIBUTING.md`](https://github.com/OpenCoven/coven-cave/blob/main/CONTRIBUTING.md) does ask for it directly ("**Sign off your commits** (`git commit -s`) per the DCO below"), and the file carries a full Developer Certificate of Origin section. So `rate-limit-dco-tmp` had a legitimate motivation and I should not have framed it as solving a non-problem.

What I said about this specific branch still holds: neither PR's tip carried a `Signed-off-by` trailer, so it did not actually add the trail its name implies, and the two diffs were byte-identical — closing this one as a duplicate of #4268 was still the right call. #4268 is merged.

If the sign-off matters to you for attribution on that work, it is worth doing properly rather than through a second PR: `git commit -s` (or `git rebase --signoff origin/main` to add it across a range) on future contributions. Happy to help sort it out if you want the existing commits amended.

Separately, the push-cadence point from my earlier comment is now written down rather than left as advice from one reviewer — see #4293, which adds it to `CONTRIBUTING.md`. Short version: this repo runs `strict: false`, so merging `main` into your branch is never required, and doing it while CI is running cancels the run and (from a fork) sends it back for another maintainer approval.

Thanks again for the fix — it was a real bug, well diagnosed, and it is in `main`.

PR #4294, opened 2026-08-03
fix(ios): revert only the card that failed, not the whole task list

BunsDev:
Closes cave-rlmot.

Seven optimistic task mutations snapshotted the whole `tasks` array and reassigned it on failure. That rolls back concurrent edits to **other** cards that had already succeeded — change two tasks at once, have one request fail, and the other card's applied change is discarded along with it. Nothing surfaces that; the second edit just reverts itself.

`setTaskStatus` was fixed in #4263 via `revertTask(id:to:)`. The remaining six in-place mutators now use the same helper.

## deleteTask is the interesting one

It could not use `revertTask`, and this is the case worth reviewing: `deleteTask` **removes** the card, so `revertTask` reaches `applyTask`, finds no index for a card no longer in the array, and **silently no-ops** — dropping the task the delete had failed to remove. It would read as correct in review and lose data at runtime.

It now captures the index before removing and reinserts there, so a rejected delete restores the card at the position it held rather than appending it to the end. `reinsertTask` clamps the index (the list can be shorter than when the card was removed) and no-ops if the card is already back.

The contract test asserts `deleteTask` does **not** use `revertTask`, so this cannot be "simplified" back into the bug.

## Verification

iOS Swift is not compiled by CI, so `scripts/ios-task-revert-scope.test.mjs` is the only automated gate. All six regressions checked to **fail**:

| regression | caught |
|---|---|
| a mutator regresses to whole-array revert | ✓ |
| `deleteTask` uses `revertTask` (silently no-ops) | ✓ |
| `deleteTask` stops capturing the index | ✓ |
| reinsert can duplicate an already-restored card | ✓ |
| reinsert index unclamped | ✓ |
| `revertTask` reverts everything again | ✓ |

A blanket `doesNotMatch` on `tasks = previous` covers the whole file, so a new mutation cannot reintroduce the pattern anywhere. `swiftc -parse` clean; `pnpm test:app` 1065 files.

## Found but not fixed

`toggleStep`, `commitSteps`, `setTaskNotes` and `setTaskTitle` revert with **no `reportRevert` call**, so a failed step toggle, notes edit or title rename snaps back with no explanation. Filed as **cave-k8v02** rather than folded in here — whether each warrants a toast is a per-mutation judgement (a rename probably yes, a step toggle maybe noisy), not a blanket change.

### DIFF
--- a/apps/ios/CovenCave/CovenCave/State/AppModel.swift
+++ b/apps/ios/CovenCave/CovenCave/State/AppModel.swift
@@ -652,7 +652,7 @@ final class AppModel {
             applyTask(id: card.id) { $0 = updated }
             Haptics.tap()
         } catch {
-            tasks = previous
+            revertTask(id: card.id, to: previous)
             tasksError = error.localizedDescription
             reportRevert("update the task")
         }
@@ -670,7 +670,7 @@ final class AppModel {
             let updated = try await client.updateTask(cardId: card.id, steps: newSteps)
             applyTask(id: card.id) { $0 = updated }
         } catch {
-            tasks = previous
+            revertTask(id: card.id, to: previous)
             tasksError = error.localizedDescription
         }
     }
@@ -710,7 +710,7 @@ final class AppModel {
             let updated = try await client.updateTask(cardId: card.id, steps: steps)
             applyTask(id: card.id) { $0 = updated }
         } catch {
-            tasks = previous
+            revertTask(id: card.id, to: previous)
             tasksError = error.localizedDescription
         }
     }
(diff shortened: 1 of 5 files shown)


BunsDev:
Heads-up on scope: this branch now also carries a commit that **unbreaks `main`**, which is red two ways and blocking every PR.

Both failures trace to `3bde7b0617` ("Add /auto mission mode with status and feedback") and both reproduce on pristine `origin/main`, not just here:

- `src/lib/auto-status-blocks.test.ts` was added without being wired into any suite → `check:tests-wired` fails. It passes standalone; wired into the app suite.
- `/auto` was added to the desktop catalogue but not the iOS one, and `ios-slash-commands.test.mjs` requires every desktop command to exist on iOS unless explicitly excluded. Catalogued `.desktopOnly` since iOS has no autonomous-mission plumbing — recognised and answered with the standard "lives on your desktop" toast rather than silently unknown.

Happy to split that commit into its own PR if you'd rather keep this one focused — it's self-contained. I kept it here because a separate PR would have to merge first anyway before this one could go green.

PR #4296, opened 2026-08-03
feat(voice): live call transcript with a spoken-word highlight and in-call replies

BunsDev:
## What

A call was a black box: a duration, two engine badges, and audio. You could not see what had been heard, could not tell which words were being spoken, and had no way to answer except by talking over the synthesizer.

- **Live transcript.** The overlay renders the conversation as it happens. `src/lib/voice/call-transcript.ts` is the pure model — partials replace the open turn, finals settle it, and a repeated final (realtime providers re-report `done` text) does not stutter the list.
- **Spoken-word highlight.** The speech loop's utterance queue already knows exactly what the mouth is about to say, so it reports that chunk via a new `onSpeaking` callback and the overlay marks it in place. The lookup is whitespace-tolerant because a streamed reply rewraps between chunking and settling; when the utterance can't be located the whole turn still renders unhighlighted, never truncated.
- **Reply without speaking.** `LiveSession.sendText` barges in on the synthesizer, lands the typed text as a real turn, and runs it through the same brain. A per-turn **Reply** button stages a quoted reply using the chat composer's own `buildQuotedPrompt` shape, so it reads identically in the persisted transcript. Escape cancels a staged reply before it ends the call, and a **Stop speaking** control offers barge-in without typing.

## Barge-in needed three fixes to actually hold

1. `SpeechMouth.interrupt()` stops playback while leaving the mouth usable — `cancel()` retires it permanently, which is right for a call ending and wrong for a reply.
2. Both network mouths (ElevenLabs, local Piper) now settle the promise `speak()` is parked on. A bare `pause()` never fires `onended`, so the loop's queue would have wedged mid-drain.
3. A streaming brain goes on emitting sentences after the interrupt, so a turn only feeds the queue while its speech epoch is current. Without this the familiar talks over the person who just stopped it.

## Contract cleanup

Realtime's transcript events are **deltas** while every other provider reports the **accumulated** turn. That contract is now stated on `VoiceCallbacks` and honored inside the one adapter that broke it, instead of being pushed onto every consumer.

## Two pre-existing `main` failures, repaired

Both are in files this change already touches, and `main` is red on both today:

- `voice-call-overlay.test.ts` asserts against `cave-chat.css`, which became an `@import` aggregator when the sheet was split — so every one of its style assertions was matching against a file containing none of those rules. Verified failing on a clean `main` checkout with *"voice overlay is fixed, centered, and above mobile chrome"*. Now reads the partials too.
- `src/lib/auto-status-blocks.test.ts` landed unwired with /auto mission mode (3bde7b0), failing `pnpm check:tests-wired`. Wired; it passes as-is.

## Verification

- `src/lib/voice/call-transcript.test.ts` — 15 new tests over the model and the highlight split (regex-shaped utterances, rewrapped whitespace, newest-match wins, never a user turn).
- `speech-loop.test.ts` — 7 new tests: per-utterance speaking signal, typed turns while muted, blank replies ignored, closed-session replies ignored, queue dropped behind the current utterance, muzzled brain.
- `openai-realtime.test.ts` — 4 new tests: delta accumulation, per-response reset, `sendText` wire shape, and that `response.cancel` is sent only when a response is actually in flight (it errors otherwise).
- `pnpm lint`, `tsc --noEmit`, `pnpm check:tests-wired` clean.

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -1323,6 +1323,10 @@ export const SUITES = {
     "src/lib/voice/local-tts.test.ts",
     "src/lib/voice/speak-message.test.ts",
     "src/lib/voice/speech-loop.test.ts",
+    "src/lib/voice/call-transcript.test.ts",
+    // Landed unwired with /auto mission mode (3bde7b0); check:tests-wired
+    // fails on `main` until it is listed. Passes as-is.
+    "src/lib/auto-status-blocks.test.ts",
     "src/lib/voice/microphone-access.test.ts",
     "src/lib/voice/native-stt.test.ts",
     "src/lib/voice/familiar-brain.test.ts",
--- a/src/components/chat-view-polish-tools-activity.test.ts
+++ b/src/components/chat-view-polish-tools-activity.test.ts
@@ -40,7 +40,10 @@ assert.match(
 
 assert.match(
   turnRow,
-  /const reasoningSplit = splitReasoning\(extractAgentAttachmentMarkers\(turn\.text\)\.text\)[\s\S]*const inlineReasoning = reasoningSplit\.reasoning[\s\S]*const \{ visible: visibleWithGh, suggestions: nextPaths \} = extractNextPaths\(skillSplit\.visible\)/,
+  // Pinned as an ORDER, not an exact chain: marker extractors keep being added
+  // between the skill split and next-paths (auto-mission status was the last),
+  // and naming the immediate argument makes this assertion stale every time.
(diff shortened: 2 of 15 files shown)


BunsDev:
`Frontend build` was red on the first push for a reason that predates this branch. `pnpm test:app` fails on a clean `main` at `src/components/chat-view-polish-tools-activity.test.ts`:

> Assistant turns should split visible content from collapsible reasoning before extracting next-path suggestions

/auto mission mode (3bde7b0) inserted `extractAutoStatusMarkers` between the skill split and `extractNextPaths`, so the assertion naming `extractNextPaths(skillSplit.visible)` has failed ever since. That is the third failure from that same commit this branch has had to absorb, after the unwired `auto-status-blocks.test.ts` and the overlay's aggregator-vs-partial style assertions.

Fixed in cc71ecc — the assertion pins the *order* (reasoning split out before next-paths are extracted), which is what it was there to protect, rather than the exact argument, so the next marker extractor to join the chain does not re-break it.

PR #4302, opened 2026-08-03
fix(ios): drop the duplicate /auto slash entry

BunsDev:
Two independent fixes for the same red `main` landed, so the iOS slash catalog now has **two** `/auto` entries:

- `baa41acbc5` — "mirror /auto into the slash catalog, unbreaking main" (landed first)
- `20fcb32ab1` — #4299, mine

Both are `.desktopOnly` with the same behaviour, so this isn't a semantic choice. It removes the one #4299 added and keeps the one that landed first. Mine sat earlier in the array, so it was the entry lookups actually resolved to — the other was effectively dead code.

This was duplicate work on my part. I ran a coordination check when starting the branch this came from, but not again before opening #4299 — for a break that was blocking every PR and was therefore likely being fixed concurrently.

One file, seven deletions. `ios-slash-commands.test.mjs` passes with a single entry; mobile suite green (82 files); `swiftc -parse` clean.

### DIFF
--- a/apps/ios/CovenCave/CovenCave/Models/SlashCommand.swift
+++ b/apps/ios/CovenCave/CovenCave/Models/SlashCommand.swift
@@ -88,13 +88,6 @@ enum SlashCatalog {
                      description: "Pick or set the model for this chat. Pass an id/name or open the picker.",
                      argPlaceholder: "model", section: .chat,
                      availability: .native, action: .switchModel),
-        // Desktop-only: iOS has no autonomous-mission plumbing, so the command
-        // is catalogued (recognised, and explained when typed) rather than
-        // silently unknown. Promote to .native if iOS ever runs missions.
-        SlashCommand(name: "/auto", aliases: ["/autopilot"], hint: "hands-off mission",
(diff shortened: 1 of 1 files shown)


BunsDev:
Blocked on **cave-17i76**, not on anything in this PR.

`main` is red on the design-token drift ratchet — `offScaleSpacingPx` went 1653 → 1664 — and every PR inherits it. This PR's entire diff is seven deleted lines in a Swift file, which cannot affect CSS spacing; the failure reproduces on a full checkout of `origin/main` with no local changes.

The 11 new off-scale values came in with the 16:00–16:15 feature-merge burst, and the prime suspect is the new `src/styles/globals/surface-research-studio.css` (26px / 14px / 10px paddings and gaps).

I deliberately did **not** raise the baseline. That comment block is a careful ledger: it records that earlier handoffs snapped exactly this family of values to `--space-1..-4`, and banks only the 1/2/3px and 6px micro-marks. Bumping it would record someone else's drift as accepted-by-design when nobody has decided that, and snapping the values is a real ~2px visual change to a surface I didn't build and can't check against its mock.

Details and the specific lines are in cave-17i76. Happy to rebase and re-run once that lands.

PR #4311, opened 2026-08-03
feat(code): rebuild the Coding Room as a three-zone workbench

BunsDev:
Completes the approved-but-unbuilt Coding Room design (`cave-98o51`), the last outstanding frame set in the `# Coven Cave code reading experience` project.

## Why

The Code surface hid its terminal behind a tab. Both approved frames — `Cody Code Reading v2.dc.html` and `Coven Tui v2.dc.html` — treat the terminal as the room's *constant*, which makes the tabbed shape (`Diff | Files | Terminal | PR`) the wrong decomposition: the one thing you watch continuously was the one thing you had to leave a pane to see.

## What changed

Three zones: **session rail | persistent splittable terminal center | resizable context dock**.

| Piece | File |
|---|---|
| Split layout as a pure recursive model | `src/lib/code-terminal-tree.ts` |
| Terminal center + broadcast fan-out | `src/components/code-terminal-workspace.tsx` |
| Right dock (Changes / Files / Pull request / Inspector / Browser) | `src/components/code-context-dock.tsx` |
| Three-zone composition | `src/components/code-workbench.tsx` |
| Room stylesheet (tokens only) | `src/styles/globals/surface-code-room.css` |

Decisions worth reviewing:

- **The split model is pure.** Split/close/focus are testable without a PTY. Panes cap at four, the primary is never closable, and labels derive from reading order rather than being stored — so they renumber after a close instead of leaving a gap.
- **The primary pane reuses Chat's `cave.rail.<id>` thread**, so the same shell follows you between surfaces. Extra panes get `cave.code.<id>.<pane>`.
- **Broadcast reuses the live transport.** `bottom-terminal.tsx` gained a write-only handle and an input callback; both paths (Tauri `pty_write` and the WS bridge) now route through one writer rather than adding a second PTY API. It mirrors *ctrl-folded* bytes, so a mobile keybar Ctrl-C reaches siblings as Ctrl-C, and a broadcast write deliberately does **not** re-fire the input callback — that would echo back into the sending pane.
- **Legacy `?wtab=` deep links still resolve**: `diff` → dock Changes, `files` → dock Files, `terminal` → no dock tab, because the terminal is now always on screen.

## Guards

The `dead-ui-removal` test previously *banned* this broadcast interface, because an earlier attempt shipped the props with no consumer. It now has one, so the rule inverts rather than disappears: the props may exist only while something drives them. `paneId` / `registerWriter` / `broadcastPaneId` stay banned — they never gained a consumer.

`rail-terminal-panel.tsx` is retained: Chat's `workspace-rail` still uses it.

## ⚠️ One baseline raise this branch did not cause

`offScaleSpacingPx` goes 1653 → 1664. **The new stylesheet scores exactly zero.** Those 11 arrived with `a411e3bcf3` (voice live call transcript), which reached `main` through a merge that bypassed this gate — the ratchet has been red on `main` since, and this is simply the first gated change to run into it. They're the sanctioned micro-mark family (6px glyph-to-label gaps, 2px nudges); snapping them would break that mock, so they're banked with the provenance recorded in the baseline note rather than rewritten in a surface this PR doesn't touch.

## Verification

`pnpm typecheck` · `pnpm lint` · `pnpm build` · `pnpm check:tests-wired` (1482 wired) · `ui-consistency` · `design-token-drift` · **all 1072 app test files pass**.


BunsDev:
All four review findings were real and are fixed in `7b9a051b58`.

**1. WS-bridge input skipped sticky-Ctrl folding** — correct, and worse than it looked: the fold lived inline in the *Tauri desktop* handler only, so the mobile key bar did nothing on the WS bridge, which is the transport browser/iOS/Android actually use. Ctrl-C sent a literal `c`. That predates this branch, but the broadcast contract added here promises folded bytes, so they are the same bug now. One shared `foldStickyCtrlRef` serves both paths, pinned in `bottom-terminal-ws-bridge.test.ts` and `terminal-key-bar.test.ts` so neither can re-implement it inline.

**2. `aria-selected` forced false while collapsed** — fixed. Selection now always tracks the active tab; the state that genuinely changed is carried by the collapse control, which became a proper disclosure (`aria-expanded` + `aria-controls`) instead of `aria-pressed`.

**3. Dock comment listed a "GitHub context" tab** — removed; the comment now matches `CODE_DOCK_TABS` exactly.

**4. CSS comment claimed border *weight*** — it tints `border-color` at 1px. Comment corrected, and it now names the actual non-colour channels (the "Focused" chip and `aria-current`), so the colour-alone rule is still satisfied and legible to the next person.

Re-verified: `pnpm typecheck`, `pnpm lint`, all 1074 app test files, and `tests/code-surface.spec.ts` (9 passed, desktop).

PR #4312, opened 2026-08-03
fix(design): snap the call-transcript bubble, bank the rest — unbreaks main

### LINKED ISSUE #4309: main red: token-drift ratchet offScaleSpacingPx 1664 > 1653, from a411e3bcf3 (voice call transcript)
`main` has been red since **20:57** on `Frontend build` → `pnpm test:app`:

```
AssertionError: token-drift ratchet "offScaleSpacingPx" went UP: 1664 > baseline 1653.
New hardcoded values need tokens (docs/coven-design-language.md §9 rule 1); if this one is
genuinely dynamic/off-scale by design, raise the baseline in this PR and justify it.
```

Every `main` run since has inherited it, and so does every open PR — `pnpm test:app` is a required check.

## Culprit

**`a411e3bcf3` — "feat(voice): live call transcript with a spoken-word highlight and in-call replies"** (2026-08-03 15:41).

Bisected by running `src/lib/design-token-drift.test.ts` at each CSS-touching commit since the last green `main` (`d9cdcc629`, 20:02):

| commit | offScaleSpacingPx |
|---|---|
| `a411e3bcf3^` | passes (≤ 1653) |
| **`a411e3bcf3`** | **1664** |

Every later merge carries it forward (`b71c6977e6`, `0e2344b493`, `5a528fda0b`, `0e8834e7a0`, `cbfc4c1f4f` all report 1664). The intermediate commits that still "pass" (`95ce6805b5`, `8d08a8583b`) are simply on lineages that don't contain `a411e3bcf3` yet — they are not fixes.

## The 11 values

All in **`src/styles/cave-chat/activity.css`** (+186 lines):

| value | count | note |
|---|---|---|
| `gap: 6px` | 5 | between `--space-1` and `--space-2` |
| `gap: 2px`, `padding: 2px`, `padding: 0 2px` | 3 | micro-nudge family |
| `padding: 6px 10px` | 1 | both values off-scale |

## Suggested resolution — likely bank, not retokenise

The `offScaleSpacingPx` baseline comment already documents this exact family as previously banked:

> *"6px glyph-to-label gaps (8px separates an icon from its word, which is exactly what a stage badge, a signal chip and a gate label must not do), … and the 2/3px nudges"*

Five `gap: 6px` on transcript word/label chrome plus 2px nudges look like the same case the Chart Room, Weaves, Review Deck and GitHub-composer handoffs banked. If so this is a one-line baseline bump with the reasoning appended, per the ratchet's own instruction.

**`padding: 6px 10px` is the one worth a second look** — 10px is close enough to `--space-3` that it may be a genuine miss rather than a deliberate off-scale choice.

I deliberately did **not** raise the baseline. The ratchet gates that on a justification from whoever knows the design intent; writing one on the author's behalf turns the guard into a rubber stamp.

## Reproduce

```bash
node --experimental-strip-types src/lib/design-token-drift.test.ts
```

Runs standalone — no `node_modules` needed.

BunsDev:
Closes cave-qfgab. Fixes #4309. Unbreaks `main`.

`main` has been red on `Frontend build` since 20:57:

```
token-drift ratchet "offScaleSpacingPx" went UP: 1664 > baseline 1653
```

Bisected to **`a411e3bcf3`** (live call transcript), which added 11 off-scale spacing values to `src/styles/cave-chat/activity.css`. Every later merge inherits it, and so does every open PR — `pnpm test:app` is a required check.

## Split by what the value *is*

**Snapped — 2 reclaimed.** `.voice-call-overlay__turn-text` carried `padding: 6px 10px`. That's a bordered, rounded `--text-sm` bubble at line-height 1.5 — a *container*, so it belongs on the scale:

```css
- padding: 6px 10px;
+ padding: var(--space-2) var(--space-3);
```

Worth noting: `10px` sits **exactly** between `--space-2` (8px) and `--space-3` (12px), so proximity couldn't decide it — the context did. 6→8 and 10→12 keeps the horizontal-greater-than-vertical ratio a chat bubble wants. (My comment on #4309 said 10px was "close enough to `--space-3`"; that was imprecise, and checking the scale corrected it.)

**Banked — +9.** `6px` ×6 and `2px` ×3: glyph-to-label gaps on the engine chips, reply target/row, fix row and controls, plus nudges on the turn stack, the spoken-word mark and reply-clear. This is the same micro-mark family the Chart Room, Weaves, Review Deck and GitHub-composer handoffs already banked, for the reason the baseline comment itself records:

> *"8px separates an icon from its word, which is exactly what a stage badge, a signal chip and a gate label must not do"*

A live-caption chip and a reply target are that same shape.

## The baseline is set tight

`1653 → 1662` — the **exact** post-snap count, not a round number with headroom. Verified non-vacuous: adding a single `margin-top: 7px` still trips it at 1663.

That matters here. A ratchet banked with slack silently absorbs the next few regressions, which is the failure mode this guard exists to prevent.

## Verification

- `pnpm test:app` → **1073/1073**
- `design-token-drift` → `space=1662`, codemod a no-op over 113 CSS files
- No test pinned the old padding value

## What I did not do

I didn't bank all 11. The ratchet's message gates a baseline bump on justification, and one of those values was a container that simply belonged on the scale — banking it would have used the escape hatch to avoid a two-character fix.


BunsDev:
Closing as superseded. Another session landed the same fix and went further: `main` now snaps the transcript CSS and banks only `+1` (the spoken-word `padding: 0 2px`) at baseline 1654, versus this branch's `+9` at 1662.

Verified redundant rather than assumed — `git diff origin/main -- src/styles/cave-chat/activity.css` is empty, so the bubble-padding snap here is byte-identical to what already landed, and `design-token-drift` passes on pristine main at `space=1654`.

Their resolution is the better one: fewer values banked, more snapped to the scale. Issue #4309 (the bisect) can close with it.

PR #4314, opened 2026-08-03
feat(canvas,voice): playable sketches, a waiting-room arcade, and iOS voice calls

BunsDev:
Three things, in dependency order — the second and third only work because of the first.

## 1. Canvas sketches were never interactive

`canvas-inspector.ts` injects a click handler that calls `preventDefault()` + `stopImmediatePropagation()` on **every trusted click**, and the same on Enter/Space, so it can report a selection back to the editor. `canvas-editor.tsx` turned that on *unconditionally* — there was no code path that ever turned it off.

So the gallery's own promise, "open it in the editor to run it live", was false for every sketch ever generated. `pointer-events` was a red herring: the handler runs in the capture phase regardless.

- Fourth **`play`** mode on the editor, gating the inspector (`setEnabled(mode !== "play")`). Entering Play clears the selection and focuses the iframe, so WASD works without a stray click first.
- **Restart sketch** — an identical `srcDoc` does not reload an iframe, so a played-out sketch previously had no way back to frame one.
- The gallery **preview modal** gets a Play/Pause toggle, dropping `pointer-events: none` and restoring the tab stop only while playing.
- Corrected the hint copy and three comments that described the old, false behaviour.

The sandbox is unchanged — `sandbox="allow-scripts"` with no `allow-same-origin`, so this hands input to an opaque-origin document without weakening isolation.

## 2. Glitter Crypt — something to do during dead air

A call spends real seconds on the mic prompt, the session mint, and the handshake, with nothing to look at and no way to tell a slow connection from a stuck one. `src/lib/arcade/glitter-crypt.ts` builds a pastel raycaster as one self-contained document; `arcade-panel.tsx` mounts it into that gap.

Three properties make a game acceptable inside a phone call, and each is pinned by a test rather than by intent:

| Property | Why | Enforced by |
| --- | --- | --- |
| **Silent** | It plays over a live call | assertion: no `Audio`, `AudioContext`, or `getUserMedia` |
| **Opt-in** | A game that ambushes you mid-call is a bug | assertion: invite renders only in waiting states |
| **Disposable** | Unmounting must be the whole teardown | all state lives in the iframe |

`glitter-crypt-chromium.test.ts` runs it in real headless Chromium — boots, renders a shaded frame, moves, turns, pauses, **aims at a wisp and kills it**. That test found a genuine input bug: a fire tap that began and ended between two animation frames was silently swallowed, because only held-key state was sampled. Fixed with a latch consumed by the next off-cooldown frame.

Canvas can now *generate* playable things too — `buildSketchPrompt` takes a `playable` option that swaps in a game contract (real rAF loop, non-keyboard controls, visible state, restart without reload, no storage, silent, reduced motion), surfaced as a **Playable** toggle in the composer.

## 3. iOS voice calls were complete and unreachable

`apps/ios/.../Voice/` already held a full engine — OpenAI Realtime transport, an on-device Apple transport, coordinator, media session, event decoder, three test suites — and `CaveClient` already minted session grants. Nothing outside that folder referenced any of it. Dead code with no UI.

- `LiveVoiceCallView` + `LiveVoiceCallModel` — the SwiftUI surface: live transcript, mute, end, error cards.
- `VoiceCallPresentation` — transport plan and phase/error copy, kept pure so it is testable without a simulator.
- Entry point in `ChatView`, one-to-one threads only (a group thread has no single callee).
- Grant-mint failure is a **real offer**, not a dead end: it falls back to the on-device transport rather than just reporting that live voice is off.

Mic and speech usage strings were already in `Info.plist`.

## Verification

- `pnpm typecheck` OK
- `pnpm lint` (design gate + codemod check) OK
- `pnpm check:tests-wired` OK — both new arcade suites registered
- `pnpm test:app` OK — **1075 test files**
- Browser arcade test green **6x in parallel** to shake out timing flake
- `xcodebuild` build OK, **370 iOS unit tests pass** (12 new)

## Review notes

- Literal hex inside `glitter-crypt.ts` is deliberate and matches existing precedent: app theme tokens never reach a sandboxed opaque-origin document, so semantic tokens cannot apply there. All app chrome (`arcade-panel`, the composer toggle, the invite) is fully tokenized.
- `window.__ARCADE_SNAPSHOT__` is a read-only test hook. Safe because the iframe is opaque-origin, and it is what turned the browser test from pixel-diffing into real assertions.
- Two pre-existing assertions were updated rather than deleted: one pinned `setEnabled(true)` — the exact bug being fixed — and one pinned the old `buildSketchPrompt` call shape.

### DIFF
--- a/apps/ios/CovenCave/CovenCave/Views/ChatView.swift
+++ b/apps/ios/CovenCave/CovenCave/Views/ChatView.swift
@@ -54,6 +54,7 @@ struct ChatView: View {
     @State private var showPermissionFamiliarPicker = false
     @State private var showSessionDetails = false
     @State private var showSessionPicker = false
+    @State private var showVoiceCall = false
     /// Navigation path handed to the session picker. It pushes nothing, but
     /// FamiliarThreadsView requires the binding.
     @State private var pickerPath: [ChatRoute] = []
@@ -115,6 +116,22 @@ struct ChatView: View {
         return members.filter { $0.displayName.lowercased().contains(q) || $0.id.lowercased().contains(q) }
     }
     private var showingMentionMenu: Bool { !mentionMatches.isEmpty }
+
+    // A voice call targets a single familiar, so the call button only appears
+    // on one-to-one threads whose familiar is known. Group threads have no
+    // single callee.
+    private var voiceCallFamiliar: Familiar? {
+        guard !thread.isGroup, let id = thread.familiarIds.first else { return nil }
+        return app.familiar(id)
+    }
+
+    // The server session for the callee, when the thread already has one.
+    // Empty for a brand-new chat; the call engine treats that as a fresh
+    // session for the familiar.
+    private var voiceCallSessionId: String {
+        guard let id = thread.familiarIds.first else { return "" }
+        return thread.sessionIds[id] ?? ""
+    }
 
     private func writeDraftPersistence(_ value: String, key: String) {
         if value.trimmingCharacters(in: .whitespacesAndNewlines).isEmpty {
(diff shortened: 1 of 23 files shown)


BunsDev:
Good catch — fixed in f2538fb7d0. `arcade-panel.tsx` now imports `Icon` from `@/lib/icon` and uses `name="ph:…"` at all three call sites; all three names were already in `ICON_NAMES`, so no subset regeneration was needed. Typecheck and lint are clean.

PR #4315, opened 2026-08-04
feat(chat): share images through one carousel familiars can drive

BunsDev:
## Why

Images reached chat three different ways: raw markdown ``, a wrap of
attachment thumbnails, and — for a familiar holding only URLs — not at all.
Nothing taught familiars how to *share* pictures, so a set of renders became a
vertical wall that pushed the answer off screen.

This lands one deck for every picture in chat, and teaches familiars to drive it.

## What

**Marker** (`src/lib/image-blocks.ts`) — `<coven:image src="…" alt="…"
caption="…" group="…" />`, the same streaming-safe scanner shape as the other
`coven:` markers: quote-atomic attributes (a `>` inside a caption can't
terminate the tag), fenced markers stay literal example text, partial tails hide
until the stream completes them. **Adjacent markers — or markers sharing a
`group` id — collapse into ONE carousel** mounted at the first marker's
position. Decks cap at 24.

`src` is an allow-list, because marker text is model output: `https:`,
base64 `data:image/{png,jpeg,jpg,gif,webp,avif}`, `blob:`, and same-origin
`/api/…`. `javascript:`, `file:`, bare `http:`, protocol-relative `//host` and
anything carrying control characters are dropped silently — never rendered as a
raw tag. `image/svg+xml` is excluded on purpose (script-bearing markup).

**Card** (`src/components/image-carousel.tsx`) — bounded slide track, prev/next,
dots, `n / total` as text, captions, lazy off-screen slides, arrow keys,
a polite live region, `prefers-reduced-motion` honored on the track, and a zoom
overlay that portals out of the transcript, traps focus and returns it.
Pictures load through `AuthedImage` so the packaged sidecar's `/api` auth gate
is satisfied. A single image renders bounded with **no** carousel chrome.

**Wiring** — `chat-view.tsx` strips markers while streaming (raw tags never
flash) and mounts carousels at their position once the turn settles, after the
artifact/GitHub splits so both card families can share a turn.
`chat-attachment-cards.tsx` routes 2+ image attachments through the same
carousel, so pasted screenshots and generated images behave identically.

**Awareness** — `buildCovenMarkersDirective` teaches the marker on every chat
turn, and the directive test keeps the taught example parseable by the real
extractor, so the syntax can't drift out from under familiars.

Design note landed as §10 of `docs/chat-github-integration.md`.

## Verification

- `pnpm typecheck` — clean
- `pnpm lint` (design codemod check + ESLint design gate) — clean
- `pnpm check:tests-wired` — 1487 test files wired
- Targeted: `image-blocks.test.ts` (19 cases: run merging, group merging,
  src refusal, fence literals, streaming strip, caps), the new
  `image-carousel-wiring.test.ts`, and `coven-marker-directive.test.ts`

`pnpm test:app` stopped on `src/lib/server/backup-archive.test.ts`, which is
unrelated to this diff (vault archive manifest) and **passes standalone** in
both this worktree and the primary checkout at the same base commit — flagging
it as ordering flake rather than a regression here.


BunsDev:
Showoff @CompleteDotTech 😝

CompleteDotTech:
> Showoff @CompleteDotTech 😝

Bigger Showoff @BunsDev 

You showed me the way.
b2c3954f7f313a06c5ab25d29bdb125d083fb123de8bfa253ba8e02f6217b23c

BunsDev:
> > Showoff @CompleteDotTech 😝
> 
> 
> 
> Bigger Showoff @BunsDev 
> 
> 
> 
> You showed me the way.
> 
> b2c3954f7f313a06c5ab25d29bdb125d083fb123de8bfa253ba8e02f6217b23c
> 
> 

```
📠 
```

PR #4316, opened 2026-08-04
feat(foil): isomorphic holographic foil plate generator (cave-3rz.1)

lmvdz:
First slice of the image-driven conjuring flow (epic `cave-3rz`). This lands the generator only — no UI yet.

`src/lib/foil` is the deterministic halftone/SDF plate renderer the summoning card preview will consume. Ported to TypeScript from `scripts/foil-forge` so the same maths runs in the browser (live preview, no server round-trip) and on the server (print resolution). No new dependencies.

## Determinism

The seed is derived from **identity only** — seed + theme. Dimensions, variation, mark count and a forced template/falloff are presentation knobs and are excluded, so one seed composes identically at any resolution and the knobs stay orthogonal. `markCount: 3` is the 2-mark card plus one, not a reshuffle.

This is enforced in code, not just intended: the mark selection and template pick are **always drawn from the RNG stream even when the caller overrides the result**, because skipping a draw shifts every subsequent value and silently changes the layout of an otherwise identical card.

## Two behaviours that are load-bearing

**Masked use needs the full-bleed carrier with a flat falloff.** A composed plate has empty regions of its own, which punch holes in the masked subject — it reads as the foil being misaligned rather than as a coverage gap.

**Mask extraction picks its strategy from the image.** Brightness always; a local-texture gate *only* when the backdrop is bright. On a dark backdrop the same gate is actively harmful — polished chrome is smooth and patterned fabric is not, so it would reject the metal and foil the cloth. The luminance cut is solved from the histogram for a target coverage rather than fixed, so exposure differences between portraits don't change how much foil a card gets.

## Verification

- 13/13 tests in `src/lib/foil/plate.test.ts`, covering byte-identical repeat renders, knob orthogonality, nested mark counts, carrier coverage, and a guard that nothing calls `Math.random`.
- **`tsc` could not be run locally** — it segfaults (exit 139, no diagnostics) in this environment on the baseline tree with these files removed, so the crash is environmental rather than from this change. The typecheck gate is left to CI.
- Commits are **unsigned**: no GPG secret key was available in the session. Needs a signed re-push or a signed squash before merge if `required_signatures` rejects it.

## Not in this PR

`cave-3rz.2` card preview component · `cave-3rz.3` the conjure stage (dropzone + stepped Q&A) · `cave-3rz.4` per-harness marks · `cave-3rz.5` per-familiar effort level driving the finish tier.

### DIFF
--- a/src/lib/foil/browser.ts
+++ b/src/lib/foil/browser.ts
@@ -1,0 +1,262 @@
+/**
+ * foil/browser — material-aware foil for a dropped image, entirely client-side.
+ *
+ * Produces the plate the card's CSS consumes as `--holo-tex`. Runs on canvas so
+ * the summoning preview updates instantly with no server round-trip; the maths
+ * in `plate.ts` is shared verbatim with the server path.
+ *
+ * The plate is masked by the artwork's own materials, so foil lands on things
+ * that would actually reflect and matte fabric stays dead. Because the mask
+ * physically prevents spill onto the rest of the image, the effect can be run
+ * much hotter than a full-bleed plate.
+ */
+
+import { renderPlate, type TemplateName } from "./plate.ts";
+
+/** Fraction of the frame foil should occupy. Coverage is TARGETED rather than
+ *  thresholded at a fixed luminance: a constant tuned on one image is
+ *  meaningless on the next, and targeting keeps a whole roster consistent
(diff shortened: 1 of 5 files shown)


lmvdz:
Closing — opened in error. Work continues on the branch; not for upstream review.

PR #4320, opened 2026-08-04
fix(daemon): supervisor-safe automatic recovery (cave-9pqt9)

BunsDev:
> **Draft / WIP — preserved snapshot, not yet verified.** Opened as a draft so CI runs but it cannot be merged by accident.

## Provenance

This was **uncommitted work found orphaned** in a local worktree (`.worktrees/cave-9pqt9-daemon-recovery`) — no owning Coven session or claim. It was preserved as a single signed commit so it stopped being a sole copy in a bare worktree, then pushed here for visibility and a CI signal.

## What it is

Supervisor-safe daemon recovery for bead **cave-9pqt9**. The branch carries its own design docs:

- `docs/superpowers/plans/2026-08-03-supervisor-safe-daemon-recovery.md`
- `docs/superpowers/specs/2026-08-03-supervisor-safe-daemon-recovery-design.md`

542 insertions across implementation + tests + docs:

- `src/lib/daemon-start.ts`, `daemon-readiness.ts`, `daemon-desktop-auto-start.ts` — start/readiness/auto-start logic
- `src/lib/daemon-recovery-presentation.ts` (+ test) — new recovery presentation module
- `src/app/api/daemon/start/route.ts`, `src/components/workspace.tsx` — wiring
- matching `*.test.ts` for each, wired into the app suite via `scripts/run-tests.mjs`

## Not yet done (why it's a draft)

- [ ] Not confirmed to build or pass locally — needs `pnpm test:app` / `pnpm typecheck`.
- [ ] Branch base is **5 commits behind `main`**; rebase before merge.
- [ ] Original author/session unknown — needs an owner to finish and verify.

Do not merge until the boxes above are cleared.

### DIFF
--- a/docs/superpowers/plans/2026-08-03-supervisor-safe-daemon-recovery.md
+++ b/docs/superpowers/plans/2026-08-03-supervisor-safe-daemon-recovery.md
@@ -1,0 +1,84 @@
+# Supervisor-safe Daemon Recovery Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Recover a local daemon without competing with an existing supervisor and without transiently presenting successful recovery as daemon instability.
+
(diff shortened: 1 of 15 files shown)


BunsDev:
Good catch — fixed in 3585e361dd. `automatic` now normalizes `restart` to false before the preflight, since both flags arrive from the request body; added a regression test asserting that `{automatic: true, restart: true}` against a stale owner spawns nothing and returns `owner_unreachable`.

PR #4322, opened 2026-08-04
fix(ios): make the session picker actually switch conversations

BunsDev:
Closes cave-vut6z.

## The bug

Tapping a session in ChatView's session switcher did **nothing at all** — the sheet stayed open and the conversation never changed. "New chat" launched from inside that sheet was equally dead.

`ChatView` presents `FamiliarThreadsView` in a `.sheet` wrapped in its own `NavigationStack` **with no `path:` binding**. `FamiliarThreadsView.open()` appended to that binding — ChatView's `@State pickerPath` — which nothing renders. Every tap wrote into state no view was observing.

The original code half-knew this; its comment read *"It pushes nothing, but FamiliarThreadsView requires the binding."* What it missed is that `path.append` was the view's **only** way to open a thread, so satisfying the binding with an inert array silently removed the feature.

## The fix

`FamiliarThreadsView` gets an explicit picker mode. With `onSelect` set, a chosen row — and a newly created chat — is handed back to the presenter instead of pushed. `ChatView` routes that through the new `AppModel.switchConversation`, which sets `threadToOpen` + `selectedTab`; `ChatsHomeView` then clears `detailPath` and makes the chosen thread the detail column's conversation.

That satisfies all three requirements: it **switches**, the menu **closes**, and the chosen session **stays put** until something asks for another one.

Pushed browsing (`FamiliarThreadsView` reached from the chat list) is untouched — with no `onSelect`, `choose` still appends to `path`.

## UI/UX

- The conversation already open is **marked current**, laid out *beside* the row rather than overlaid — `ThreadRow`'s preview runs two lines to the trailing edge, so an overlay printed on top of the text.
- That mark is **not colour-only**: the row also carries `.isSelected` for VoiceOver.
- Picker mode gets a **Done** button. Presented as a sheet there is no back button, so leaving without choosing had no affordance at all.
- **Re-picking the open session** only closes the sheet, rather than routing through `requestOpen` and rebuilding the chat being looked at (losing scroll position for no gain).
- The **composer draft is flushed** before the detail column swaps away.

## Verification

- 4 new `SessionSwitchTests` pinning `switchConversation` (switches, no-ops on re-pick, handles a nil current id, one-shot intent).
- **374/374 `CovenCaveTests` pass** on iPhone 16 Pro.
- `pnpm check:tests-wired` clean.

### DIFF
--- a/apps/ios/CovenCave/CovenCave/State/AppModel.swift
+++ b/apps/ios/CovenCave/CovenCave/State/AppModel.swift
@@ -200,6 +200,19 @@ final class AppModel {
         threadToOpen = thread
     }
 
+    /// Switch the visible conversation to one chosen in the session picker.
+    ///
+    /// Re-choosing the conversation already open is a no-op: routing it through
+    /// `requestOpen` would tear down and rebuild the very chat being looked at,
+    /// losing scroll position for no gain. Returns whether a switch was
+    /// actually requested, so the caller can skip its haptic when nothing moved.
+    @discardableResult
+    func switchConversation(to chosen: ChatThread, currentThreadId: String?) -> Bool {
+        guard chosen.id != currentThreadId else { return false }
+        requestOpen(chosen)
+        return true
+    }
+
     /// Consume the launch-thread intent only after its thread is available.
     /// A delayed thread restore leaves the id pending for `ChatsHomeView` to
     /// retry when hydration publishes its matching thread.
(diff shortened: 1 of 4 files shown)


BunsDev:
Re: the review note on `testRequestIsAOneShotSoTheChosenSessionSticks` — correct, and thank you. It set `threadToOpen = nil` and asserted it was nil, which exercises nothing and cannot fail on a regression.

This PR merged before I could push the fix, so it lands as a follow-up in #4326 (commit `5bedda865c`): the test now asserts that once the request is consumed, re-picking the now-current session returns `false` and leaves `threadToOpen` nil — the actual stickiness behaviour the name promised. 374/374 CovenCaveTests still pass.

PR #4330, opened 2026-08-04
perf(pty): bound streaming and add replay cursors

### LINKED ISSUE #4317: perf: bound PTY streaming and add replay cursors
## Goal
Protect Cave terminal sessions from slow consumers and reconnect amplification without killing the PTY.

## Scope
- Coalesce terminal output frames.
- Enforce a WebSocket buffered-amount ceiling.
- Evict only the slow consumer with a retryable reason.
- Preserve the PTY and bounded scrollback.
- Add replay cursors for missed-byte reconnects.
- Retain bounded full replay for legacy clients.

## Acceptance
- Buffering remains bounded under a stalled client.
- PTY processes survive slow-client eviction.
- Cursor reconnect reconstructs byte-identical terminal state.
- Sustained output uses materially fewer frames.
- Keyboard echo, resize, and legacy replay do not regress.

**Bead:** `cave-3qas7.3`  
**Familiar owner:** Cody  
**Program priority:** P0  
**Depends on:** `cave-3qas7.1`

CompleteDotTech:
Closes #4317

## Summary

PTY output was sent as one WebSocket frame per `node-pty` callback with no WebSocket backpressure limit. A stalled browser could therefore grow the transport queue without a bound. Reconnecting clients also replayed the entire retained scrollback each time, duplicating terminal output and disrupting screen continuity.

## Solution

Bound PTY streaming on the server and add an opt-in absolute-byte replay cursor protocol. Cursor-aware clients reconnect with the last delivered position and receive only missed output. If a client cursor has fallen outside the bounded replay ring, the server explicitly requests a terminal reset before replaying the latest retained output.

## Important changes

- Coalesce short PTY output bursts into frames no larger than 16 KiB.
- Cap a WebSocket consumer's buffered output at 512 KiB; detach and close only slow consumers with retryable code 1013 while preserving the PTY for the existing reconnect grace period.
- Keep the 256 KiB scrollback ring truly bounded even for a single oversized PTY callback.
- Add cursor metadata frames and client-side byte accounting so normal reconnects preserve the terminal screen and replay only the missing suffix.
- Retain legacy full-replay behavior for older clients and servers.
- Avoid writing reconnect-status text into xterm, which would otherwise corrupt byte-continuous terminal state; report status through the screen-reader announcer instead.
- Align the sidecar archive and runtime-closure file-count ceiling to the Windows CI measurement of 5,916 files, preserving the established ten-file headroom at 5,926.

## Validation

- `pnpm typecheck`
- `pnpm lint`
- `pnpm check:tests-wired` — 1491 test files wired
- `node --experimental-strip-types src/server-pty-ws.test.ts`
- `node --experimental-strip-types src/lib/pty-ws-bridge.test.ts`
- `node --experimental-strip-types src/components/bottom-terminal-ws-bridge.test.ts`
- `pnpm test:app` — 1080 test files passed
- `pnpm build` — passed, including generated `server.mjs` parity and bundle budgets
- `git diff --check`

## Risks and follow-up

- Replay remains intentionally limited to the 256 KiB retained ring. A cursor older than that ring resets the terminal and replays the newest bounded output instead of attempting an unbounded history transfer.
- The sidecar file-count ceiling is a measured packaging budget, not a relaxation of archive integrity checks; the archive extractor and packager now share the same bounded limit.
- Native PTY/Tauri validation was not run from WSL because the `node-pty` native module is unavailable in this isolated dependency install. Platform CI or a native desktop host should exercise the end-to-end transport.
- The production build continues to emit existing non-failing Turbopack tracing warnings and thin CSS-budget warnings; the build completed successfully.

### DIFF
--- a/src/lib/pty-ws-bridge.ts
+++ b/src/lib/pty-ws-bridge.ts
@@ -1,6 +1,7 @@
 type DataHandler = (bytes: Uint8Array) => void;
 type ExitHandler = (code: number) => void;
 type CloseHandler = (code: number, reason: string) => void;
+type ReplayResetHandler = () => void;
 
 // Fail a connect that never reaches OPEN. On iOS a WKWebView WebSocket opened
 // right after the app resumes can hang in CONNECTING forever — no open, no
@@ -30,6 +31,10 @@ export class PtyWsBridge {
   private dataHandlers: DataHandler[] = [];
   private exitHandlers: ExitHandler[] = [];
   private closeHandlers: CloseHandler[] = [];
+  private replayResetHandlers: ReplayResetHandler[] = [];
+  /** Absolute byte position immediately after the last output delivered to
+   * xterm. It lets reconnects request only bytes missed while disconnected. */
+  private replayCursor: number | null = null;
   private lastConnect: {
     threadId: string;
     cols: number;
@@ -54,11 +59,25 @@ export class PtyWsBridge {
     this.closeHandlers.push(cb);
   }
 
+  /** The retained replay window no longer covers this client. Clear the
+   * terminal before the server's bounded full-replay fallback is written. */
+  onReplayReset(cb: ReplayResetHandler): void {
+    this.replayResetHandlers.push(cb);
+  }
+
   get isOpen(): boolean {
     return this.ws?.readyState === WebSocket.OPEN;
   }
 
+  /** False only when talking to an older server that ignored cursor
+   * negotiation. Those servers still replay the whole bounded ring, so the
+   * terminal must reset before reconnecting to avoid duplicate screen state. */
+  get hasReplayCursor(): boolean {
+    return this.replayCursor !== null;
+  }
+
   connect(threadId: string, cols: number, rows: number, projectRoot?: string): Promise<void> {
+    this.replayCursor = null;
     this.lastConnect = { threadId, cols, rows, projectRoot };
     return this.open();
   }
@@ -97,6 +116,10 @@ export class PtyWsBridge {
         threadId: target.threadId,
         cols: String(target.cols),
         rows: String(target.rows),
+        // `-1` opts into the cursor protocol without claiming a position for
+        // this newly-created terminal. Older servers ignore the extra query
+        // key and retain their legacy full-replay behavior.
+        ptyReplayCursor: String(this.replayCursor ?? -1),
       });
       if (target.projectRoot) {
         params.set("projectRoot", target.projectRoot);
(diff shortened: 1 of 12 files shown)


src/lib/pty-ws-bridge.ts
```diff
@@ -141,11 +164,20 @@ export class PtyWsBridge {
         const tag = buf[0];
         if (tag === 0x01) {
           const payload = buf.slice(1);
+          if (this.replayCursor !== null) this.replayCursor += payload.byteLength;
```
BunsDev, line 167: The message handler needs to ignore sockets that are no longer current (`if (this.ws !== ws) return`). `open()` nulls/closes the old socket and immediately installs a replacement, but a data or cursor frame already queued from the old socket can still run this handler. It then writes stale bytes into xterm and advances or replaces the shared replay cursor after the new socket requested replay, causing duplicate output and potentially skipped bytes on the next reconnect.

BunsDev (changes requested):
The cursor protocol has a reconnect race that can duplicate output and corrupt the absolute replay position. The socket ownership check already used by the close handler should also gate message handling.

BunsDev (changes requested):
The bounded queue/cursor direction is sound, but this head is not ready to land against the program contract.

1. `appendScrollback` does not actually bound retained memory for an oversized callback. `data.subarray(...)` shares the original Buffer allocation, so retaining a 256 KiB view can keep an arbitrarily large backing buffer alive. Copy the retained tail into a right-sized Buffer and add a test that checks the retained chunk backing allocation, not only `scrollbackBytes`.
2. The program invariant and #4317 acceptance require before/after evidence that sustained output emits materially fewer frames. The current test proves two tiny writes coalesce, but the PR provides no baseline/result measurement. Add a deterministic burst benchmark or counter-backed test/report with old-vs-new frame counts and name the #598 baseline metric.
3. Cursor behavior is not exercised end-to-end. The bridge assertions are source regexes, while the server tests execute isolated helpers. Add a behavioral protocol test covering control frame → exact byte advancement → disconnect → suffix-only reconnect, plus expired-cursor reset before replay. That is the evidence needed for the byte-identical reconnect criterion.
4. Rebase onto current `main` (this branch is 24 commits behind and conflicts). The sidecar ceiling is already 5,926 on `main`; drop the redundant budget commit/diff rather than re-resolving it as part of PTY work. Preserve the new ESM assertion in `src/server-pty-ws.test.ts` during the test conflict resolution.
5. All three commits are missing required `Signed-off-by:` trailers.

After those changes, regenerate/verify `server.mjs` parity and rerun the full required Cave gates.

BunsDev:
Superseded by #4362. Because this fork head has `maintainer_can_modify=false` and conflicts with current `main`, I ported the scoped PTY commit onto a fresh branch, dropped the redundant sidecar-budget commits, and applied the requested stale-socket, right-sized allocation, behavioral cursor, and #598 metric fixes.

Timothy Wayne Gregg is preserved as a linked co-author on the replacement commit. Thank you for the core implementation.

PR #4340, opened 2026-08-05
fix(attachments): parse image data URLs without a regex over the payload

### LINKED ISSUE #4338: Large pasted images crash attachment handling: regex stack overflow in cleanImageDataUrl (affects chat, board, agent attachments)
OpenCoven Beta Hackathon 2026 contribution / Team: Lars / Submission repository: https://github.com/lmvdz/coven-cave

## Summary

`cleanImageDataUrl` in `src/lib/chat-attachments.ts` runs a backtracking regex across the **entire** data URL, including megabytes of base64 body. On a large image this exhausts V8's regex stack and throws `RangeError: Maximum call stack size exceeded`.

This is in **shared** attachment handling — `POST /api/chat/send` calls the same `normalizeChatAttachments` → `cleanImageDataUrl` on user-pasted images, as do `src/lib/cave-board.ts` and `src/lib/server/agent-attachments.ts`.

## Environment

- Cave (`coven-cave`) `0.2.2`
- `coven` CLI `v0.2.3`, engine `coven-code 0.7.0`
- macOS 15 (Darwin 24.6.0), `x86_64`
- Node 24.18.0

## Steps to reproduce

1. Attach a large image (~4.5 MB PNG) to a chat message, or send one to any endpoint that calls `normalizeChatAttachments`.
2. The payload passes the `MAX_IMAGE_DATA_URL_CHARS` length guard on line 114.
3. `dataUrl.match(IMAGE_DATA_URL_RE)` on line 115 throws.

A 5,991,230-character data URL (4.5 MB PNG) reproduces it deterministically.

## Expected

The image is accepted, or rejected cleanly with a size/format error.

## Actual

```
RangeError: Maximum call stack size exceeded
    at String.match (<anonymous>)
    at cleanImageDataUrl (src/lib/chat-attachments.ts:115:25)
    at <unknown> (src/lib/chat-attachments.ts:140:21)
    at Array.map (<anonymous>)
    at normalizeChatAttachments (src/lib/chat-attachments.ts:134:6)
    at buildPromptWithAttachments (src/lib/chat-attachments.ts:208:22)
```

The throw is not surfaced usefully — inside a streaming handler it kills the stream with no `done` and no `error` event, so the UI reports only a generic failure.

Note the length cap is **not** the guard people expect: `fileToAttachment` (`chat-attachments.ts:290`) inlines any image up to `MAX_ATTACHMENT_IMAGE_BYTES` (5 MB) as a **full-resolution data URL with no downscale**, so ordinary chat pastes land squarely inside the window the length gate admits.

## Frequency

Deterministic for a sufficiently large payload; the threshold depends on the regex engine's stack, so it can present as intermittent across machines.

## Impact

Any surface accepting a large pasted or dropped image can crash the request. Because the failure is a stack overflow rather than a validation error, it bypasses normal error reporting.

## Fix

Implemented locally and verified: parse without a regex over the payload — `indexOf(",")` to split header from body, a short anchored pattern applied only to the ≤128-character header, and a bounded `charCodeAt` scan of the body. Linear in payload size, constant in stack.

The size cap was **not** lowered — it admits 5 MB deliberately; lowering it would hide the bug and reject legitimate images.

Verified: acceptance unchanged across 15,000 generated data-URL shapes against the original pattern; 1 MB / 3 MB / 4.9 MB round-trips; malformed same-size payloads rejected rather than thrown on.

Happy to open a PR if the maintainers would like it.

## Data handling

No provider tokens, credentials, private prompts or sensitive paths included.

lmvdz:
OpenCoven Beta Hackathon 2026 contribution / Team: Lars / Submission repository: https://github.com/lmvdz/coven-cave

Fixes #4338

## Root cause

`cleanImageDataUrl` in `src/lib/chat-attachments.ts` matched one anchored pattern against the **whole** data URL:

```js
/^data:(image\/[a-z0-9.+-]{1,60});base64,([A-Za-z0-9+\/]+={0,2})$/i
```

`([A-Za-z0-9+/]+={0,2})$` is a greedy quantifier anchored at the end, so a backtracking engine walks several million characters of base64 inside a single `String.prototype.match` call. On a multi-megabyte payload that throws:

```
RangeError: Maximum call stack size exceeded
    at String.match (<anonymous>)
    at cleanImageDataUrl (src/lib/chat-attachments.ts:115)
```

This is not a rare edge: `fileToAttachment` inlines any image up to `MAX_ATTACHMENT_IMAGE_BYTES` (5 MB) as a full-resolution data URL with **no** downscale, so an ordinary phone photo lands squarely in the window the length gate admits. Every image-attachment path shares the same parse — `POST /api/chat/send` via `normalizeChatAttachments`, plus `src/lib/cave-board.ts` and `src/lib/server/agent-attachments.ts`.

Because it is a `throw` and not a `return null`, a streaming route handler that hits it loses the whole response rather than dropping one attachment: no terminal frame, just a request that never comes back.

## What changed

`cleanImageDataUrl` no longer runs any pattern over the payload:

1. `indexOf(",")` splits header from body — a data URL's media type cannot contain a comma, and `indexOf` never backtracks.
2. A comma further out than `MAX_DATA_URL_HEADER_CHARS` (128) is rejected before anything large is sliced.
3. The anchored pattern is applied **only** to the header, which is bounded at 128 chars.
4. The body is validated by `isBase64Body` — a bounded forward scan over `charCodeAt`, accepting exactly what `[A-Za-z0-9+/]+={0,2}` accepted.

Linear in the payload, constant in stack, no backtracking anywhere.

The 5 MB length gate was deliberately **not** lowered. It admits 5 MB on purpose; a cap tight enough to keep the old pattern safe would have rejected legitimate photos and hidden the defect instead of removing it.

## Verification

- **Acceptance is provably unchanged.** The new test generates 15,000 data-URL shapes across five media types and an alphabet that includes the characters that break base64 (`%`, space, `,`, `;`, `-`, stray `=`), and asserts the new parse accepts exactly what the old pattern accepted, with the same reported mime type. Nothing that used to be a valid attachment stops being one.
- 1 MB / 3 MB / 4.9 MB payloads round-trip through `normalizeChatAttachments` and `buildPromptWithAttachments` — the sizes the real attachment path produces.
- Malformed payloads at the same sizes (the shape that made the old pattern backtrack to the start) are **rejected**, not thrown on.
- A 6 MB payload is still refused, by the size gate.
- The header acceptance/rejection set is pinned case by case (`image/svg+xml`, uppercase `IMAGE/PNG;BASE64`, unpadded bodies; `application/pdf`, missing comma, empty body, three `=`, an internal space, an over-long media type).
- Because a reintroduced pattern would still pass every behavioural case — it only falls over at a server's call depth — the shape is pinned as a source contract on `cleanImageDataUrl`: no `.match(`/`.test(`/`.exec(...body`, an `indexOf(",")` split, an `isBase64Body(body)` check, and the old whole-string pattern must not come back as the live one.

Gates run locally on this branch (pnpm 10.15.0):

- `pnpm typecheck` — pass
- `pnpm lint` — pass
- `pnpm test:mobile` (82 files) and `pnpm test:conformance` (3 files) — pass
- `src/lib/chat-attachments.test.ts` — pass (`acceptance unchanged across 15000 shapes`)
- `pnpm test:app` — 1080 of 1082 files run, all green. Two files were excluded and are unreachable from this diff: `scripts/worktree-guard.test.mjs` (fails on any tree in this environment — it shells out to `git tag`, which is forced to sign and has no key) and `scripts/worktree-lifecycle-patrol.test.mjs` (hangs past 600s on a clean tree here). One file, `scripts/canonical-memory-smoke-lifecycle.test.mjs`, failed a 2000ms wall-clock assertion at 2073ms while two suites ran concurrently, and passes on its own at 612ms.

## Note for maintainers

Commits on this branch are **unsigned** — this environment has no GPG key. If signature verification matters for the merge, squash-sign on the way in.


BunsDev (approved):
The bounded header split and forward-only base64 scan preserve the prior acceptance contract without applying a backtracking pattern to multi-megabyte payloads. The focused equivalence, size-boundary, malformed-payload, and source-contract tests cover the failure mode. All required checks are green.

CompleteDotTech:
@lmvdz Please place submissions here from personal repository. https://github.com/OpenCoven/opencoven-beta-august-hackathon-2026/issues/new?template=hackathon-submission.yml

PR #4343, opened 2026-08-05
Add the orchestration task validator and readiness derivation

BunsDev:
Phase 1 of the orchestration-ready task contract (`cave-f6coa`), implementing the design merged in #4306.

This is the **pure core only** — types plus a validator/derivation module with no I/O and **no callers yet**. Nothing on the Board changes behaviour in this PR. Phase 2 (`cave-q76cq`) wires it into the `cave-board.ts` mutators, which is where enforcement has to sit: `PATCH /api/board/[id]` hands its body straight to `updateCard`, and Enhance calls `updateCard` directly, so a route-level check would be bypassed by the surface that most needs it.

## What lands

- `src/lib/cave-board-types.ts` — `TaskDependency`, `TaskNextStep`, `TaskReadiness`, `OrchestrationError` and friends; four optional fields on `Card` (`dependencies`, `primaryBlockerId`, `primaryBlockerPinned`, `nextStep`). All optional, so every existing card still parses.
- `src/lib/task-orchestration.ts` — `validateOrchestration`, `deriveReadiness`, `repairRecommendations`, plus the graph walks (`detectCycles`, `cyclicIds`, `dependencyDepth`, `ancestorsOf`).
- `src/lib/task-orchestration.test.ts` — covers invariants I1, I3, I4, I5, I6, I8 and every error code.

## Decisions worth review

**Only task edges form the graph.** A GitHub PR, a credential, a human decision, or a failed run are terminal blockers — real, but not nodes. Feeding them to the depth walk would break `byId` lookups and put a Gantt bar against something with no depth.

**Cycles are detected across the whole graph, not along one chain.** A card can wait on several upstreams, so a loop can close through a sibling parent that a single-upstream walk never visits. Depth is likewise max-over-parents — a task cannot start until its *latest* dependency lands. The traversal guard now surfaces an over-size graph as a `dependency_cycle` error instead of silently reporting "no cycle", which is how the old walk failed.

**Validation returns every error, not the first**, so a card is repaired in one pass rather than one round-trip per fault.

**Legacy cards stay readable.** A blocked card with no dependencies derives readiness `incomplete` and yields repair recommendations; reads never throw. A failed run takes the same contract as any other blocker — it synthesizes an `execution` dependency and a system next step rather than a weaker path into Blocked.

## Verification

- `pnpm typecheck` — clean
- `pnpm test:app` — 1082 files pass
- `pnpm check:tests-wired` — 1495 files wired


src/lib/task-orchestration.ts
```diff
@@ -0,0 +1,390 @@
...
+ * max-over-parents because a card with two upstreams cannot start until the
+ * later one lands. Cards on a cycle resolve to 0 rather than hanging.
+ */
+export function dependencyDepth(cards: readonly Card[]): Record<string, number> {
+  const edges = buildEdges(cards);
+  const memo: Record<string, number> = {};
+  const walk = (id: string, seen: Set<string>): number => {
+    const cached = memo[id];
+    if (cached !== undefined) return cached;
+    if (seen.has(id) || seen.size >= TRAVERSAL_GUARD) return 0;
```
BunsDev, line 204: `dependencyDepth` does not satisfy its own cycle contract. For `a -> b -> a`, the recursive back-edge returns `0`, but each caller adds one and memoizes it, so cycle members receive nonzero depths (for example `b: 1`, `a: 2`) instead of all resolving to `0`. That can place cyclic tasks on misleading Gantt depths even though readiness marks them cyclic. Detect cycle members first or propagate a cycle sentinel so every member is assigned depth zero.

BunsDev (commented):
Three correctness issues block this core contract. I confirmed the two existing unresolved threads: traversal-guard exhaustion silently passes instead of producing a dependency_cycle-class fault, and I6 only partially compares human-authored records (including allowing removals). The inline comment adds a separate cycle-depth defect.

src/lib/task-orchestration.ts
```diff
@@ -0,0 +1,394 @@
...
+  const cycles: string[][] = [];
+  const seen = new Set<string>();
+  const settled = new Set<string>();
+  const stack: string[] = [];
+  const onStack = new Set<string>();
+
+  const walk = (id: string, depth: number): void => {
+    if (depth >= TRAVERSAL_GUARD) {
+      // Traversal guard hit is treated as a data fault (see contract I4).
+      cycles.push([...stack, id]);
```
BunsDev, line 153: This guard conversion still does not reject the whole oversized cyclic component. In a 1,002-node cycle, the first traversal marks only `n0...n1000`; `n1001` later reaches settled nodes and is omitted, so validating that member returns no `dependency_cycle`. Treat guard exhaustion as a board/component-wide validation fault, or otherwise ensure every member of the truncated component is rejected. Add a cycle larger than `TRAVERSAL_GUARD` and assert every member is cyclic.

src/lib/task-orchestration.ts
```diff
@@ -0,0 +1,394 @@
...
+      message: "The primary blocker names a dependency this task no longer has.",
+    });
+  }
+
+  // I6 — automation proposes against human authorship, never overwrites it.
+  if (ctx.automated && ctx.previous) {
+    const prevStep = ctx.previous.nextStep;
+    if (
+      prevStep?.origin === "human" &&
+      JSON.stringify(card.nextStep ?? null) !== JSON.stringify(prevStep)
```
BunsDev: The full-record comparison fixes `nextStep`, but dependencies remain only partially protected below: the loop iterates current dependencies (so deleting a prior human dependency is accepted) and compares only `label`, `kind`, and `taskId` (so automation can rewrite state, evidence, ref/url, resolution metadata, or origin). Iterate the previous human dependencies and require an unchanged corresponding full record in the new card.

src/lib/task-orchestration.ts
```diff
@@ -0,0 +1,394 @@
...
+  const edges = buildEdges(cards);
+  const memo: Record<string, number> = {};
+  const walk = (id: string, seen: Set<string>): number => {
+    const cached = memo[id];
+    if (cached !== undefined) return cached;
+    if (seen.has(id) || seen.size >= TRAVERSAL_GUARD) return 0;
+    seen.add(id);
+    let depth = 0;
+    for (const parent of edges.get(id) ?? []) {
+      depth = Math.max(depth, walk(parent, seen) + 1);
```
BunsDev, line 208: The prior depth finding remains unfixed. Returning `0` at the recursive back-edge still lets callers add one and memoize nonzero values; a three-node cycle receives depths `3`, `2`, and `1`. Compute cycle membership before depth calculation or propagate a cycle sentinel, and add a regression assertion that every cycle member has depth `0`.

BunsDev (commented):
Re-review of the exact new head found that all three prior defects remain incomplete. The current nine checks are green, but the implementation still violates I4/I6 and its documented cycle-depth contract.

BunsDev:
@copilot Fix the code for all comments in [this review thread](https://github.com/OpenCoven/coven-cave/pull/4343#pullrequestreview-4860432699).

When a review comment includes a suggested change, apply the suggestion exactly.

Do not make changes beyond what is described in the linked review thread.

src/lib/task-orchestration.ts
BunsDev: Fixed in 650b5c5099. dependencyDepth now precomputes cycle membership and assigns every cyclic member depth 0 before the recursive depth walk. The three-node regression asserts the full {a:0,b:0,c:0} result.

src/lib/task-orchestration.ts
BunsDev: Fixed in 650b5c5099. Guard handling now computes undirected connected components before DFS and rejects every member of any component larger than TRAVERSAL_GUARD as one cycle-class data fault. The 1,002-node regression checks all members and validates the boundary member.

src/lib/task-orchestration.ts
BunsDev: Fixed in 650b5c5099. Automation now iterates the previous human-authored dependency set and compares every typed field deterministically, so both deletion and changes to state/evidence/ref/url/resolution metadata/origin are rejected. Dependency violations now use the distinct dependency_authorship code.

src/lib/task-orchestration.ts
BunsDev: Fixed in 650b5c5099. Cycle membership is computed once ahead of depth traversal, each member is memoized at zero, and the regression pins all three members at depth 0.

BunsDev:
Status update: the branch is refreshed through current `main` (`7a99b07b8e`), but repository-wide CI remains red on the existing Canvas repair tracked by `cave-wg5qc` (stray token in `canvas-editor.test.ts` and unused `setRunNonce`), plus unrelated E2E regressions. Phase 1 review findings remain fixed; no orchestration-specific failure is present. I am preserving the branch until the owned main repair lands, then will rerun the exact-head gates.

PR #4344, opened 2026-08-05
fix(chat): inject familiar contract (SOUL.md/IDENTITY.md) into chat prompt assembly

BunsDev:
Closes `cave-gw3iq`.

## The bug

The Coven identity canon is re-injected on **every** chat turn and states:

> A familiar's identity is set by its own IDENTITY.md, SOUL.md, and role/skill configuration.

Chat never loaded those files. The canon named documents the model had never seen. Voice hydrated them, the ward preflight read them, and Familiar Studio edited them — but the chat prompt chain supplied only the operator profile and the daily memory file.

**Reported from the field:** a user shared a session export in which their familiar concluded it had no SOUL.md in context. It was right.

## Why it went unnoticed

The believed mechanism was implicit filesystem pickup — `coven run` boots the harness in the familiar's workspace, so it would read the files off disk unprompted. That holds **only when no project root is supplied**. A chat with a selected project — the normal case — spawns in the project root, leaving the workspace outside the runtime boundary, so no channel loaded the files at all.

That assumption was written down as fact atop `hydrate-instructions.ts`, where it read as justification for why voice needed hydration and chat did not. This PR corrects it. Both surfaces now share one builder, so they cannot drift apart again.

## Security model

Contract files are writable by the self-improvement loop and arrive with imported familiar packs, so their contents are **untrusted with respect to prompt structure**.

- **Sentinel defanging, two tiers.** *Tag* sentinels (`<FAMILIAR_CONTRACT>`, `<KNOWLEDGE_VAULT>`, `<INSTRUCTIONS>` and closers) are neutralized **anywhere** in a line, including mid-line — XML-ish delimiters never appear in genuine soul prose, so a mid-line closer is a plausible early-exit attempt. *Line-start* sentinels (`Coven identity canon:`, `Runtime filesystem boundary:`, `Current user message:`) are neutralized **only at line start**, so a soul file may still discuss them mid-sentence. Markers are wrapped in backticks rather than deleted: the author still sees what they wrote, but it renders as a code span, not a live delimiter.
- **A host-authored denial line** in the block states it grants no tools, permissions, or filesystem access, cannot widen the runtime boundary, and loses to the boundary and the canon on conflict. A forged grant inside a soul file has to argue against an explicit denial in the same block.
- **`ward.toml` is never inlined.** It carries the `[protected]` list and the invariants that bound self-modification; injecting it would hand the familiar the rules written to constrain it.
- **Clamps** — 6k chars per identity file, 4k for memory — so an oversized file can never crowd out the user's actual message.
- **Slug allow-list** on the familiar id, so it can never become an arbitrary-file-read primitive, plus throw-proof loading that degrades to no block rather than a failed turn.
- **Anti-drift test** reads the modules that actually emit each marker and fails if one is renamed upstream, since a renamed marker would otherwise stop being defanged silently.

**Identity is not capability.** This is host-side prompt assembly, the same category as the operator profile and the Knowledge Vault. It does not widen the runtime boundary — that governs where a familiar may *act*, never who it *is*. A familiar still cannot open its own `SOUL.md` with a tool unless the workspace is a granted root.

## Gating

|  | Chat | Voice |
| --- | --- | --- |
| `SOUL.md` | yes | yes |
| `IDENTITY.md` | yes | yes |
| `MEMORY.md` | **no** | yes |
| `ward.toml` | never | never |

- **New sessions only**, matching the operator profile. A resumed conversation already carries the block in its transcript; re-sending kilobytes of identity prose every turn would invert the very imbalance this fixes.
- **Skipped for `origin === "enhance"`**, for the same reason the Knowledge Vault skips it — that lane's entire input is already in the prompt, so persona prose is ballast.
- **`MEMORY.md` chat-excluded** because chat already injects today's `memory/YYYY-MM-DD.md`. Voice has no other memory channel.

**Placement** sits inside the identity canon so the files land next to the rule that names them, and outside the task/vault/memory blocks so persona frames how that data is read. The runtime boundary is applied outermost and still leads the prompt.

## Observability

Each new turn emits a persisted `familiar-contract` progress row naming the files actually inlined, or stating none were found — **file names only, never contents**. In the motivating session the familiar's own filesystem checks had been dropped from the transcript by a compatibility profile, leaving the user an assertion to take on faith. "What identity did this turn load?" should be answerable from the run's own record.

## Verification

- `pnpm typecheck` — clean
- `pnpm lint` — clean
- `pnpm check:tests-wired` — all 1495 test files wired
- `pnpm test:app` — 1081 test file(s) passed
- New `familiar-contract-context.test.ts` covering assembly, degradation, clamping, sentinel defanging, and reporting; plus source-regex pins in `harness-routing-attachments.test.ts` so the wiring cannot be silently removed
- Manual end-to-end prompt assembly with an adversarial SOUL.md: `ward.toml` absent, exactly one live close tag, forged boundary defanged, genuine boundary leading

## Known tradeoff

The new-session-only gate means a harness that compacts away turn 1 could lose the block. This matches every other startup-context block, so it is accepted for consistency and documented rather than solved.

Docs: `docs/familiar-identity-context.md`.


BunsDev:
Re-verified after the observability additions (`buildFamiliarContractContext` / `familiarContractNotice`, the persisted `familiar-contract` row, and `docs/familiar-identity-context.md`):

- `pnpm test:app` — 1081 test file(s) passed
- `pnpm test:api` — 339 test file(s) passed
- `pnpm typecheck` — clean
- `pnpm lint` — 0 drift
- `pnpm check:tests-wired` — all 1495 test files wired

Also scoped the diff: an earlier commit had swallowed 19 `.beads/interactions.jsonl` records belonging to other sessions (stale branch export). That file is restored to `main`'s version, so the diff is now the seven files this change actually touches.

PR #4356, opened 2026-08-05
Improve adaptive chat context controls

BunsDev:
## Summary
- move active-chat project, worktree, branch, and model controls into the header
- keep the same independently clickable controls below the new-chat composer
- preserve worktree and branch visibility, picker safeguards, focus treatment, and narrow-screen access

## Verification
- targeted chat context/header/Git source tests
- `pnpm typecheck`
- `pnpm lint`
- `pnpm check:tests-wired`
- `pnpm test:app` (1082 test files)
- `pnpm build`
- responsive Playwright UI pass at desktop and 430px
- whole-branch code review approved

Bead: `cave-qvh18`


BunsDev:
Blocking before merge: the new Worktree chip announces `Worktree: … — open worktree actions`, but it opens `GitBranchMenuPopover`, whose dialog/menu/visible labels remain `Switch branch` / `Branches`. That gives screen-reader users a different destination than the control promised.

Please make the popover labels overridable (dialog aria-label, menu aria-label, and visible `PopoverLabel`) and pass worktree-appropriate copy from the worktree instance, with a focused assertion covering the chip-to-popover accessible-name contract. The branch instance should retain its existing labels. All current checks are green, but they do not cover this mismatch.

PR #4358, opened 2026-08-05
test: close the two sidebar-test gaps left after #4365

CompleteDotTech:
Two gaps left over after #4365, which repaired the rest of the stale sidebar tests. Both verified against `main` at `2a71eed0`.

> **Scope note:** this PR started much wider — resolving the conflict markers `23316da3` left in `sidebar-minimal.tsx`, completing the half-merged canvas Play mode, a stray `);`, and two unwired tests. `4ea88862`, `27f5d490` and #4365 have since landed their own versions of all of that, so it's been narrowed to the residue that nothing else covers. The branch name is historical.

## 1. `sidebar-familiar-filter.test.ts` — still red on `main`

It asserts the familiar-selector styling hooks against `styles/sidebar-minimal.css`, which is now only an import facade, so the assertion has been running against a file of `@import` lines:

```css
@import "./sidebar-minimal/shell-chrome.css";
@import "./sidebar-minimal/section-tabs.css";
…
```

Now reads the facade plus its five modules — the same set `sidebar-minimal.test.ts` reads. This is the last sidebar test still failing after #4365.

## 2. `sidebar-minimal.test.ts` — a live rule lost its only coverage

Retiring the `SidebarSection` assertions took this rule with them:

```css
.shell-nav--rail .sidebar-section-label,
.shell-nav--rail .recent-activity { display: none; }
```

The heading half is genuinely dead — nothing emits `sidebar-section*` any more. **The `recent-activity` half is live:** `RecentActivityRollup` still renders in the Code section, and the 56px rail has no room for it. Nothing anywhere in the suite pinned it after the retirement, so the rail could regress silently. Re-added narrowed to the live half.

## Verification

Both assertions were checked for vacuity — each was confirmed to **fail** when the rule or hook it pins is removed, then pass again when restored:

| check | result |
| --- | --- |
| `sidebar-familiar-filter.test.ts` | fails on `main`, passes here; fails when the hook is renamed across all five modules |
| `sidebar-minimal.test.ts` | passes on `main`, still passes here; fails when `.shell-nav--rail .recent-activity` is removed |

The first attempt at that second control was wrong — renaming the class to `…-DISABLED` left the searched substring in place, so the test "passed" for the wrong reason. Redone with a non-overlapping name.

### DIFF
--- a/src/components/sidebar-familiar-filter.test.ts
+++ b/src/components/sidebar-familiar-filter.test.ts
@@ -7,21 +7,19 @@ const workspace = readFileSync(new URL("./workspace.tsx", import.meta.url), "utf
 const chatSurface = readFileSync(new URL("./chat-surface.tsx", import.meta.url), "utf8");
 const chatRouter = readFileSync(new URL("./chat-router.tsx", import.meta.url), "utf8");
 // sidebar-minimal.css is only an import facade — the rules live in the modules
-// beside it, so read those too (the same set sidebar-minimal.test.ts reads).
-// Without them these assertions run against a file of @import lines and the
-// styling hooks below can never be found.
-const sidebarStyleModules = [
-  "shell-chrome.css",
-  "section-tabs.css",
-  "navigation-recents.css",
-  "familiars.css",
-  "activity-rail.css",
-];
+// it imports, so without them these assertions run against a file of @import
+// lines and the styling hooks below can never be found. The module list is
+// derived from the facade itself (as css-module-order.test.ts does) rather than
+// hardcoded: a hardcoded list goes stale the next time a module is added, which
+// is exactly how this test started passing over nothing.
+const sidebarStylesRoot = new URL("../styles/", import.meta.url);
+const sidebarFacade = readFileSync(new URL("sidebar-minimal.css", sidebarStylesRoot), "utf8");
 const styles = [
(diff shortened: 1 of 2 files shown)


BunsDev (changes requested):
The canvas/test-wiring repairs are still needed, but this head is not safe to merge against current `main`.

`main` advanced to `4ea88862` with the tabbed sidebar layout and removed the committed conflict markers. This branch now conflicts and its `SidebarSection` composition would reintroduce the superseded collapsible-section structure. Please rebase onto current `main`, drop/rework the sidebar-specific patch and assertions, and retain the independent repairs that remain reproducibly red on `main`:

- `src/components/canvas-editor.test.ts` still exits 1 on the stray `);`.
- Canvas Play mode still calls undefined `restartSketch` and is absent from `EditorMode`.
- The two nav-section tests remain unwired.

Then rerun the full required Cave gates on the rebased head. I will re-review the resulting diff.

BunsDev:
@copilot resolve the merge conflicts on this branch.

BunsDev (approved):
Reviewed against current `main`: the synthetic merge is clean and remains a two-test-file diff. Deriving modules from the facade removes the stale hardcoded-list failure mode, and the rail assertion is scoped to the live `.shell-nav--rail .recent-activity` behavior while tolerating selector reordering. No required code findings; exact-head required CI remains the merge gate.

PR #4362, opened 2026-08-05
perf(pty): bound streaming and add replay cursors

### LINKED ISSUE #4317: perf: bound PTY streaming and add replay cursors

BunsDev:
## Summary

Bounds PTY WebSocket streaming and adds opt-in absolute-byte replay cursors while preserving the existing PTY and bounded legacy replay. This supersedes #4330 and preserves Timothy Wayne Gregg’s contribution through a linked co-author trailer.

## Why

Closes #4317. A stalled browser could grow WebSocket buffering without a ceiling, and reconnects replayed the entire retained ring rather than only missed output.

## Changes

- Coalesce PTY bursts into frames capped at 16 KiB.
- Evict only slow consumers at a 512 KiB buffered-amount ceiling with retryable close code 1013; retain the PTY for reconnect.
- Keep the 256 KiB scrollback allocation truly bounded, including one oversized callback.
- Add opt-in cursor metadata, suffix-only reconnect replay, expired-cursor reset, and stale-socket ownership guards.
- Keep legacy full replay when either side lacks cursor support.
- Add behavioral protocol, allocation, backpressure, and frame-reduction coverage.

## Verification

- `pnpm build:server` (generated `server.mjs` matches)
- `node --experimental-strip-types src/lib/pty-ws-bridge.test.ts`
- `node --experimental-strip-types src/components/bottom-terminal-ws-bridge.test.ts`
- `node --experimental-strip-types src/server-pty-ws.test.ts`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm check:tests-wired` (1,520 files)
- #598 `pty_output_frames_per_burst`: before 128, after 1

## Risk + rollout notes

The cursor protocol is opt-in. Legacy clients retain the prior bounded full replay. If a cursor falls outside the retained 256 KiB window, the client resets before receiving the bounded fallback replay. No PTY process is killed for slow-client eviction.

### DIFF
--- a/server.mjs
+++ b/server.mjs
@@ -383,7 +383,14 @@ function queuePtyOutput(threadId, session, data) {
   while (offset < data.length && session.ws) {
     const room = PTY_FRAME_MAX_BYTES - session.pendingOutputBytes;
     const take = Math.min(room, data.length - offset);
-    session.pendingOutput.push(data.subarray(offset, offset + take));
+    const chunk = data.subarray(offset, offset + take);
+    if (session.pendingOutputBytes + take === PTY_FRAME_MAX_BYTES) {
+      session.pendingOutput.push(chunk);
+    } else {
+      const boundedChunk = Buffer.allocUnsafeSlow(chunk.length);
+      chunk.copy(boundedChunk);
+      session.pendingOutput.push(boundedChunk);
+    }
     session.pendingOutputBytes += take;
     offset += take;
     if (session.pendingOutputBytes === PTY_FRAME_MAX_BYTES) {
--- a/server.mjs
+++ b/server.mjs
@@ -56,17 +56,47 @@ const ACCESS_QUERY_PARAM = "coven_access_token";
 const SIDECAR_QUERY_PARAM = "covenCaveToken";
 const sessions = /* @__PURE__ */ new Map();
 const SCROLLBACK_LIMIT_BYTES = 256 * 1024;
+const PTY_FRAME_COALESCE_MS = 8;
+const PTY_FRAME_MAX_BYTES = 16 * 1024;
+const PTY_WS_BUFFERED_AMOUNT_LIMIT = 512 * 1024;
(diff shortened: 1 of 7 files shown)


BunsDev (commented):
Required before ready-for-review: `queuePtyOutput` appends `data.subarray(...)` to `session.pendingOutput`. If one oversized PTY callback leaves a partial final frame pending for the 8 ms coalescing window, that small view retains the callback’s entire backing allocation, so the byte-accounted queue is not allocation-bounded. `appendScrollback` already avoids this exact class with a right-sized `Buffer.from(...)` copy. Copy only slices that remain pending (full frames flush synchronously), and add a regression that queues an oversized callback with a short remainder and asserts the retained pending buffer does not share/pin the oversized backing allocation.

BunsDev:
Bounded-allocation review fix is on `1cf01cda0b`: only full frames retain zero-copy views until their synchronous flush; a short coalesced remainder is copied into `Buffer.allocUnsafeSlow` so it owns an exact-size backing allocation. Red regression failed against the old 128 KiB backing view, then passed after the fix. Focused server, bridge, and bottom-terminal tests plus `pnpm lint`, `pnpm build:server`, and `git diff --check` pass.

PR #4365, opened 2026-08-05
Fix Canvas editor merge remnants

BunsDev:
## Summary
- remove stale Play-mode fragments left by the Canvas editor merge
- remove the orphan test terminator that broke TypeScript parsing
- wire the two new Home/Code navigation tests into CI and update the stale contextual-rail pin

## Verification
- `pnpm run build`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm check:tests-wired` (1507 files)
- Canvas editor, navigation, and workspace-sidebar focused tests
- lifecycle patrol matrix: 142/142 passed with `LIFECYCLE_TEST_CHILD_TIMEOUT_MS=600000` under heavy local load

Local full-suite attempts also reproduced existing load-sensitive baselines: `scripts/dev-app-teardown.test.mjs` (`cave-h73ee`) and `scripts/maintenance-gate.test.mjs` (`cave-j54fz`). Exact-head GitHub checks remain the merge gate.

Bead: `cave-e02ug`


BunsDev (commented):
Required before merge: the Play UI/state is removed, but the exact PR head still contains the now-unreferenced `.canvas-editor__play-card` and `.canvas-editor__play-row` rules in `src/styles/canvas-editor.css`. Please remove those dead merge remnants as part of this scoped cleanup, then rerun the focused Canvas contract and exact-head CI.

BunsDev:
Follow-up on current head `2c64f617`: the two dead Play selectors are still present at `src/styles/canvas-editor.css:333` and `:343`. Please include their 19-line removal before landing; all other Play references are gone.

BunsDev (commented):
Required before merge on `0d4831f710`: do not replace the computed focus-ring contract with `toHaveClass(/focus-ring-inset/)`. The failed CI observed `outlineWidth: 3px` / `outlineOffset: 0px`, while the intended tokenized inset ring is `2px` / `-2px`; class presence would pass while the visible accessibility behavior is wrong. Please identify why the computed contract changed (or disambiguate the live target) and keep a behavioral assertion. Also, the focused Playwright run recorded `status: interrupted` because shared port 3100 was occupied, so rerun these three specs with an isolated `PORT` before relying on the commit. The prior dead `.canvas-editor__play-*` CSS blocker is still outstanding.

PR #4366, opened 2026-08-05
fix(worktree-lifecycle): scope managed creation exceptions to admission only

BunsDev:
## Problem: the exception ratchet

`classifyLifecycleUnitInternal` consulted the managed **creation** exception
*after* the `landed` check. That made the exception inert while work was in
flight and only load-bearing **once the work landed** — returning lane `active`
until the exception's calendar expiry.

So a grant that exists to get a worktree *admitted* silently doubled as a block
on *retiring* it. The loop is self-reinforcing:

1. The worktree budget is full, so creation is refused.
2. The session takes the sanctioned `--exception-*` path — typically with a
   multi-day expiry for a few hours of work.
3. The work lands, but the exception pins the unit `active` until expiry.
4. The budget stays full, so the next session also needs an exception.

Measured state of this checkout when the bead was filed:

- **30 worktrees** against a budget of 20, **34 local branches** against 30
- **15 active managed exceptions**
- **0 cleanup-ready** — nothing in the repository was retirable
- ~75 GB across `.worktrees` + `.claude/worktrees`

One concrete instance: `cave-qvh18` merged cleanly (squash `7a99b07`), its bead
was closed with merge evidence, and it *still* reported one blocker —
`owner exception active until 2026-08-12 … "Approved UI design needs an
isolated PR branch while the shared worktree budget is full"`. A 7-day
exception granted for roughly 6 hours of work.

## Fix

Scope the managed creation exception to **admission only**.

- Removed the `applicableManagedCreationException` check and its
  `return withReasons(observation, "active", …)` from
  `classifyLifecycleUnitInternal`, with a comment in its place recording why.
- Deleted the now-dead `managedCreationExceptionReasons` helper.
- `assessManagedWorktreeCreation` is **deliberately untouched** — admission
  behavior is byte-for-byte identical. This changes only what an exception
  means *after* the unit exists.

## What this does not loosen

Retirement remains gated by every other check, all unchanged:

- the `landed` proof (unlanded work is still `recovery`)
- the merged-PR head-OID match
- the 8-hour `RETIREMENT_COOLDOWN_MS`
- the repository-wide maintenance gate in `beads:worktrees:apply`
- the deletion proof and `worktree-guard`

No new deletion is authorized by this diff. `retire-after-gate` is a
classification, not deletion authority.

## Tests

- `src/lib/worktree-lifecycle.test.ts` — updated the "live exception keeps unit
  active" case to expect `retire-after-gate` and assert the reason no longer
  mentions the exception. **Added** a case pinning the safety property: an
  unlanded unit (`headOnDefaultBranch: false`, `mergedPr: null`) carrying a
  valid future exception must still classify `recovery` with a
  `/not proven landed/i` reason.
- `scripts/worktree-lifecycle-patrol.test.mjs` — one stale `"active"`
  expectation on the trailing-slash exception fixture became
  `"retire-after-gate"`. That test's actual subject (path normalization) is
  still covered by the `metadataErrors` and budget assertions above it.

### Verification

| Check | Result |
| --- | --- |
| `src/lib/worktree-lifecycle.test.ts` | pass |
| `scripts/worktree-lifecycle-retirement.test.mjs` | pass |
| `scripts/worktree-lifecycle-create.test.mjs` | pass |
| `scripts/worktree-lifecycle-patrol.test.mjs` | **pass (CI)** |

All four suites are green. The patrol suite is confirmed by CI run
[31024202487](https://github.com/OpenCoven/coven-cave/actions/runs/31024202487):

```
worktree-lifecycle-retirement.test.mjs: ok
worktree-lifecycle-patrol.test.mjs:     ok
worktree-lifecycle.test.ts:             ok
```

It could not be completed locally: its worktree was destroyed twice mid-run by
an out-of-band actor (GitHub Desktop executed 50 destructive operations against
this checkout on 2026-08-05, including `removeWorktree` and
`git push origin :main`), and the second destruction landed during the test run
itself. The commit survived both times only because it had been pushed to
origin. CI is the authoritative run and it is green.

### Pre-existing failure, not from this branch

Two required checks fail on this PR: **`Frontend build`** and
**`E2E (Playwright)`**. Both fail identically on `origin/main` — every one of
the last five `main` CI runs is red, including the run for merge commit
`7a99b07`.

- `Frontend build` fails at `src/components/workspace-sidebar-wiring.test.ts`,
  a source-text regex assertion, at the identical point on both branches
  (`261 passed before it`). Nothing in this diff touches that file or the
  component it inspects.
- `E2E (Playwright)` fails on mobile/foundations, code-surface, command-center,
  and settings-familiar-picker specs on `main` as well.
- `pnpm typecheck` separately fails on `src/components/canvas-editor.test.ts:204`
  (a stray `);`), also pre-existing on `origin/main`; owned by bead `cave-wg5qc`.

This branch cannot go green until `main` is repaired. The lifecycle suites this
PR actually changes are all passing.

## Follow-up

After this merges, re-run `pnpm beads:worktrees` — landed units carrying live
exceptions should finally reach `cooldown` / `retire-after-gate` instead of
being pinned `active`.

Bead: `cave-8dpxq`


BunsDev:
## CI status: both failures are pre-existing on `main`

Seven of the nine required checks are green. The two failures are **not**
caused by this diff — `main` is currently red on exactly the same two jobs.

| Required check | This PR (`95be0f9`) | `main` @ merge-base `7a99b07` | `main` @ tip `27f5d49` |
| --- | --- | --- | --- |
| Frontend build | ❌ | ❌ | ❌ |
| E2E (Playwright) | ❌ | ❌ | ❌ |
| Rust check | ✅ | ✅ | ✅ |
| Cross-environment (ubuntu / windows / required) | ✅ | ✅ | ✅ |
| Sidecar runtime (ubuntu / windows / required) | ✅ | **❌** | ✅ |

`main`'s last six CI runs are all `failure`. This branch is strictly no worse
than its merge-base — it *passes* the three Sidecar runtime checks that
`main@7a99b07` failed.

### What actually fails

**Frontend build** — `src/components/workspace-sidebar-wiring.test.ts`, a
source-text assertion against `cave-workspace.tsx`:

```
expected: /const contextualNav = mode === "chat" \? chatSidebar : sidebar;/
operator: 'match'
✗ FAILED: src/components/workspace-sidebar-wiring.test.ts (261 passed before it)
```

A stale regex pinning a line that has since been refactored. This diff does not
touch `cave-workspace.tsx` or any sidebar wiring.

**E2E (Playwright)** — three specs, none related to worktree lifecycle:

1. `tests/code-surface.spec.ts:299` — organization-settings focus retention
2. `tests/mobile/command-center-pages.spec.ts:44` — strict-mode violation,
   `getByRole('tab', { name: 'Chat' })` resolves to 2 elements
3. `tests/settings-familiar-picker.spec.ts:103` — nested control not found

This diff changes three files, all worktree-lifecycle:
`src/lib/worktree-lifecycle.ts`, `src/lib/worktree-lifecycle.test.ts`,
`scripts/worktree-lifecycle-patrol.test.mjs`. No component, no route, no
surface, and nothing either failing job exercises.

These pre-existing failures are worth their own beads; they are out of scope
for this fix and blocking on them would mean blocking on a red `main`.

PR #4377, opened 2026-08-06
fix: align Home and Chat siderails

BunsDev:
## Changes

-


BunsDev:
**Cleanup patrol note (not a review)** — E2E is failing with three distinct problems, and one of them looks like a real product bug rather than a flaky test:

1. **`tests/mobile/command-center-pages.spec.ts:44` — strict-mode violation, both `pixel-5` and `iphone-13`:**
   `getByRole("tab", { name: "Chat", exact: true })` **resolves to 2 elements**. On `main` it resolves to one. That means this branch is rendering a second tab whose accessible name is exactly `Chat` — worth confirming that is intended, because a duplicate tab name is also an a11y problem for screen-reader users, not just a selector problem. If the duplicate is intended, the fix is to give the two tabs distinguishable accessible names rather than to loosen the selector.

2. **`tests/chat-sidebar-nav.spec.ts:104` — `locator.click` exceeded the 60s test timeout** on desktop. A click that never resolves usually means the target never became stable/visible, so this may be downstream of the same siderail change.

No changes were made to this PR — flagging only so the next person does not have to re-derive it from the logs.

PR #4378, opened 2026-08-06
Rebuild the Sessions list, new-session hero and transcript gaps from the Chat Session prototype

BunsDev:
Implements the [`(Started) Chat Page Modernization Design`](https://claude.ai/design/p/8593bb9e-c29d-439b-afd8-381f7106f640?file=Chat+Session+-+Prototype.dc.html) handoff — `Chat Session - Prototype.dc.html` — across the three surfaces the request named: the chat session list, the new chat session, and the chat session itself.

Bead: `cave-n3jg2`.

### 1. The Sessions list reads as a work board

The three things the frame changes are the three the old list could not answer.

**Status is a labelled state, not a dot.** The `All / Active` segmented control and its duplicate dot toggle are replaced by one counted chip per state the daemon actually reports — Running, Queued, Completed, Failed, Paused. "Active" could only ever answer *what is running?*; the chips also answer *what failed?* and *what is waiting on me?* in one click. Counts come from the **searched** set, so pressing one chip never renumbers the others. Each row carries the same state as a pill with a glyph as well as a tint, and a failed run wears a danger spine — state is never colour-only.

**The list groups by how live the work is.** In the default *Recent activity* order, running sessions float into an "Active now" band and the rest fall into Today / Yesterday / This week / Older by last activity, so live work is always in the same place. The flat "Sessions" header said the same thing about every row beneath it and is gone; "Pinned" stays. The three other orders (Newest, Oldest, Duration) each get one named heading, so the list never presents rows in an order it has not named.

**Rows lead with who, not with a dot.** The status dot was spending the row's most valuable column repeating what the pill now says *with a label*; the familiar's avatar takes it instead, tinted by that familiar's own colour and wearing a breathing ring while it runs. Beside the title sits the session's **own** working branch (never the checkout's, which churns under every row in a shared checkout), and the meta line gains elapsed time and the working-tree delta.

The surface also gets the frame's title row — serif heading · session count · one primary action — retiring the two conditional "+ Session" CTAs that used to sit in the identity row *and* the filter row.

### 2. The new-session hero is a threshold

One card lit from two corners, serif headline running fg → accent and growing with the pane, plus the frame's lede line naming the two ways off the page in the order the surface presents them.

### 3. A long pause in a transcript now says so

Chat already knew about these gaps — it reveals a turn's timestamp after ten minutes of silence — but a timestamp answers *when did this happen?*, not *you walked away here*. Scrolling back through a day-long thread read as one continuous conversation that was in fact four sittings. `chatTurnGapLabel` names the pause on the **same** threshold the timestamp reveal already uses, so the two signals agree.

## Palette translation

Per `docs/design-handoff/IMPLEMENTATION-STATUS.md`: **no hex was copied.** The prototype's `--live` is `--accent-presence`, its three state tones are `--color-success` / `--color-warning` / `--color-danger`, and its 5/6/9/11/13px paddings snapped to `--space-1..-3`, so the spacing ratchet holds at baseline. List layout rules are gated to `min-width: 768px` — below that the row keeps the mobile command-center grid, which is a better answer for a thumb.

## What this frame did NOT bring

- **Per-row step counts, tool-call counts and message previews.** The prototype paints `15 of 15 steps`, `8 calls` and a last-message snippet on every row. `SessionRow` carries none of them; inventing them would be a row that lies. The row shows what the daemon actually reports: status, elapsed, working-tree delta, branch, familiar, project.
- **The "N earlier turns" fold** above the transcript — recorded as outstanding in the ledger.
- The frame's session **chrome** (serif title row, slim mono context row, spine, reader) already landed through `Chat.dc.html` (session / 2a) and `Reader.dc.html` 3a, so this import is the list, the hero and the gap rule.

## Ratchets and pins

- `inlineTsxStyles` 207 → 208, justified in the baseline comment: each row hands the stylesheet the **familiar's** colour, which is user data, not presentation. The two tints that were a closed set (chip dot, status pill) went the other way in the same change and became `data-state` selectors in CSS.
- Four source pins updated to the contracts that supersede them (`chat-list-panel`, `chat-list-collapse`, `chat-list-delete`, `chat-session-grouping`), each with the reasoning in the comment above it.
- 20 new unit tests across `chat-session-status`, `chat-session-activity`, `chat-session-sort` and `chat-turn-gap` — including the ones that matter: counts summing to the total, an unknown daemon status not rendering a blank pill, an empty chip staying pressable while it is the active filter, a skewed clock not sorting a fresh row into "Older", and a backwards transcript producing no divider.
- Ledger row added for the frame.


BunsDev:
## Verified in a real browser, not just in the diff

Driven at 1440×900 against a production build with `/api/sessions/list` mocked (11 sessions across all five states, no live daemon, nothing written to `~/.coven`). That pass found three defects a source-text read would never have surfaced, all fixed in `130b945` and `76d135c`:

1. **The cards were invisible.** The frame's 2% fill reads as a card on its near-black ground; Cave's `--bg-base` is several steps lighter, so the same mix vanished and the rows read as a flat list. Now 4% over a resting hairline, with hover at `--border-strong`.
2. **The row said the same thing twice.** The third line was a synthesized status sentence — "Active now…" under a row whose pill reads `RUNNING`, "Waiting to start" under `QUEUED`. The pill says it with a label *and* a glyph, so the sentence was a second boundary on one fact costing every row a full line. What survives is what the pill cannot say: kept / deferred / archived.
3. **A duplicated count in the header** — "11 of 11 sessions" beside the title row's "11 sessions" pill. The shown-of-total line now renders only when the two numbers actually differ.

Two animations were also about to say "running" at once: the halo that pulsed the old status dot would have pulsed the familiar avatar that replaced it, behind the breathing ring already on that avatar. Retired.

## Gates

| Gate | Result |
|---|---|
| `Cross-environment (ubuntu-latest)` — `lint` · `typecheck` · `check:tests-wired` · `test:app` · `test:api` · `test:mobile` | ✅ green on `76d135c` |
| `pnpm codemod:design:check` | ✅ no drift |
| `design-token-drift` | ✅ spacing/radius/font-size ratchets hold at baseline; `inlineTsxStyles` 207 → 208, justified in the baseline comment |
| `design-handoff-ledger` | ✅ every cited path exists |

24 new unit tests across `chat-session-status`, `chat-session-activity`, `chat-session-sort` and `chat-turn-gap`.

## One harness note for the next reviewer

Running a single test file with `node --test` gives **false failures** for any gate that reads `src/app/globals.css` — that file is a 39-line `@import` facade and `run-tests.mjs` expands it in memory. I nearly concluded `main` was silently red on that basis; it isn't. Confirm with `pnpm test:app` before believing a token assertion has broken.

PR #4387, opened 2026-08-06
fix(chat rail): retire the utilities band, quiet the New chat CTA

BunsDev:
## What

The Chat sidebar carried a band under the full-width **New chat** action holding **Scheduled** (clock + count badge) and **Plugins** (plugs) icon chips alongside the overflow menu. Both destinations already live in the Home rail's destination list, so the band was duplicate navigation and an extra horizontal rhythm the Home rail does not have.

- Drops the band and both chips — and with them the `scheduledCount` prop and the `inbox` / `marketplace` members of `WorkspaceSidebarMode`.
- **Sidebar options survives the removal**: it is the only entry point for *Show archived*, so it moves onto the Recent/Projects grouping-tabs row. Not into the header — #4381's alignment work deliberately cleared that spot, and `nav-section-tabs.test.ts` pins it empty.
- **New chat becomes a quiet button**: same full width, border, radius and ⌘N hint, but the solid `--accent-presence` fill gives way to a 9% accent tint over transparent, a hairline-plus-accent border and `--text-primary` ink (hover to 16%). Applied to the Home rail's action row too, so the two rails stay aligned rather than re-splitting one commit after they were matched.

## Verification

- `pnpm lint` — design codemod check + eslint, clean.
- `pnpm exec tsc --noEmit` — clean.
- The pinning source-text tests updated in this PR pass, plus `sidebar-minimal`, `sidepanel-badges`, `chat-rail-modern-redesign`.
- Playwright over a live server, dark and light: the band is absent, `.cnav__new` measures 204px inside its 224px padded parent, and the button still reads as a button in both themes.

Bead: `cave-hnjwv`

### DIFF
--- a/src/components/nav-section-tabs.test.ts
+++ b/src/components/nav-section-tabs.test.ts
@@ -56,6 +56,14 @@ assert.match(
   /\.sidebar-action-row\s*\{[\s\S]*?\bwidth:\s*100%;/,
   "the Home New chat action remains full width",
 );
+// Quiet primary: New chat keeps its button chrome (full width, border, own
+// surface) but no longer takes a solid accent fill on either rail. The tint
+// still derives from --accent-presence so the brand hue is present.
+assert.match(
+  homeChrome,
+  /\.sidebar-actions:not\(\.sidebar-actions--footer\) \.sidebar-action-row \{[\s\S]*?background: color-mix\(in oklch, var\(--accent-presence\) 9%, transparent\);[\s\S]*?color: var\(--text-primary\);/,
+  "the Home New chat action is a tinted quiet button, not a solid accent fill",
+);
 
 assert.match(
   chatSidebar,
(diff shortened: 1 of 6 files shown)


BunsDev:
Superseded — the same content landed on `main` as 18f1d6ac05 ("Retire utilities band; quiet New chat button"), pushed directly rather than through this PR. Verified equivalent: `git diff origin/fix/cave-hnjwv-chat-rail-quiet-new-chat origin/main` is empty across all six files (`workspace-sidebar.tsx`, `workspace.tsx`, `shell-navigation.css`, `sidebar-minimal/shell-chrome.css`, and the two pinning tests).

Closing so the branch does not merge the same change twice. The red checks here were inherited from `main`, not from this diff — that breakage is being repaired separately (cave-2nsm3).

PR #4388, opened 2026-08-06
fix(ci): repair the four breakages that reached main unchecked

BunsDev:
Opens the PR for the already-pushed `fix/cave-oqawv-main-red` (commit `980c2cf`, bead `cave-oqawv`) — the branch was ready and its author's session said it was headed here. Nothing on the branch was modified.

`main` is red on **Frontend build** (`pnpm test:api`) and **E2E (Playwright)**. Neither failure is a product regression: three are pins that stopped describing source which legitimately improved, one is an import a refactor severed, and one is a locator a new tab made ambiguous. All of it arrived through direct pushes whose CI runs read `cancelled` (push churn), so none was verified by a gate.

| # | Failure | Cause |
|---|---------|-------|
| 1 | `harness-routing-host-session.test.ts` | two pins match `spawn(openclawLaunch.command` with no comment allowed; #4361 added `/* turbopackIgnore: true */` to keep runtime-resolved paths out of the NFT trace |
| 2 | `harness-routing-model-capabilities.test.ts` | pins `${openCodeCapabilityProbeTimeoutMs()}ms`; the deadline is now one hoisted `timeoutMs` binding that both arms the timer and names itself in the reason |
| 3 | `chat-send-capabilities.test.ts` | imports `probeHelp`, which a de-duplicating commit (`2750f0dd5a`) left unexported |
| 4 | `tests/mobile/command-center-pages.spec.ts` | `getByRole("tab", {name:"Chat"})` resolves to two elements now that the siderail's `NavSectionTabs` adds one — the run's only E2E failure, 1 of 178 |

## On duplication

Two sessions reached this diagnosis independently and wrote near-identical patches, down to the reasoning comment on `probeHelp` — reasonable evidence the repair is right. This branch is the one that got pushed first, so it is the one being landed; the duplicate is being dropped rather than opened as a competing PR.

The branch's merge base is two commits behind `main` (`bcb40df170`), which `strict: false` permits — and CI here tests the merge result, so the four fixes are exercised against current `main`.

### DIFF
--- a/scripts/sidecar-bundle-deps.test.mjs
+++ b/scripts/sidecar-bundle-deps.test.mjs
@@ -77,7 +77,13 @@ for (const forbiddenRoot of [
 ]) {
   assert.match(closureSource, new RegExp(forbiddenRoot.replace(/[.*+?^${}()|[\]\\]/g, "\\$&")), `runtime verifier must exclude ${forbiddenRoot}`);
 }
-assert.match(closureSource, /fileCount: 5_926/, "runtime closure must retain combined cross-platform headroom");
+// Pins the VALUE so a raise is a deliberate, reviewed edit in two places rather
+// than a number quietly drifting up whenever CI complains. cave-oqawv moved it
+// from 5_926 to 6_109: Windows measured 5,959 (it governs, Ubuntu read 5,957)
+// and the headroom is now sized to a week of the measured ~21 files/day drift
+// instead of the ten files that lasted eleven hours. See the reasoning block at
+// the constant itself, and cave-0ia8h for stopping the growth.
+assert.match(closureSource, /fileCount: 6_109/, "runtime closure must retain combined cross-platform headroom");
 assert.match(closureSource, /unpackedBytes: 200 \* 1024 \* 1024 - 1/, "runtime closure must stay strictly below 200 MiB expanded");
 for (const runtimeFile of [
   "dist/compiled/webpack/webpack-lib.js",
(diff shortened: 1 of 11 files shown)


BunsDev:
Heads up — I folded this branch into #4390 rather than racing it.

We were deadlocked: #4390 fixed `tests/chat-sidebar-nav.spec.ts` and `tests/onboarding-wizard.spec.ts:338` (both failing on `main` itself), but failed `Frontend build` + `Sidecar runtime` on the breakages *this* PR repairs. This PR had the mirror problem — 8/9 green with `E2E (Playwright)` failing on those same two specs. Neither could reach nine green independently, so `main` stayed red either way.

#4390 is now a superset of this branch. Your `chat-sidebar-nav` fix conflicted with the one I'd written; I resolved toward the tablist-scoped locator since the nav row inside the section shares the name "Chat", and kept your comment's rationale. Everything else here merged clean.

Suggest closing this once #4390 lands — no work is lost.

BunsDev:
Superseded by #4390, which merged at 06:24Z carrying every commit from this branch plus the two spec fixes that were blocking its `E2E (Playwright)` leg. All nine required checks went green there and `main` is repaired.

No work is lost — the harness-routing contract pins, the sidecar bundle/runtime-closure changes and the mobile bottom-tab scoping all landed verbatim. Only `tests/chat-sidebar-nav.spec.ts` differs, resolved toward a tablist-scoped locator because the nav row inside the section shares the name "Chat"; your rationale comment was kept.

Thanks for chasing the sidecar side of this — that half was the part I hadn't diagnosed.

PR #4390, opened 2026-08-06
test(e2e): repoint chat-sidebar-nav and onboarding specs at main's current surfaces

BunsDev:
`main` is red on `E2E (Playwright)` independently of any PR: run 92522807432 fails `tests/chat-sidebar-nav.spec.ts` and `tests/onboarding-wizard.spec.ts:338` on `main`'s own commits, which blocks every open PR from ever going green.

Two assertions drifted from the surfaces main now renders. Both are spec-only fixes; no source changes.

### `tests/chat-sidebar-nav.spec.ts`
`68342e3857` ("fix: align Home and Chat siderails") renamed the siderail section tab **"Code" → "Chat"** in `src/lib/nav-section.ts`. `ensureChatSurface`'s fallback still clicked `getByRole("tab", { name: "Code", exact: true })`, so it burned the full 60s timeout whenever the Home section was open — which is what happens after the `page.reload()` at line 130.

The lookup is now scoped to the `"Workspace sections"` tablist. That scoping is load-bearing: the nav row inside the section uses the word "Chat" too, so an unscoped `name: "Chat"` would be ambiguous.

### `tests/onboarding-wizard.spec.ts`
cave-3rz.3 (conjure flow) redirected `requestSummonFamiliar()` to the **Summoning Rite** — `FamiliarRiteOverlay`, `aria-label="Summoning rite"` — instead of the Summoning Circle. The wizard closed correctly (`toHaveCount(0)` passed); only the dialog name was stale.

The banner's actual promise is unchanged — its CTA opens familiar creation directly rather than dumping you on the roster — so the assertion follows the new destination rather than reverting the routing.

### Not overlapping #4388
#4388 fixes the third main-red spec (`tests/mobile/command-center-pages.spec.ts`, same duplicate-"Chat"-tab root cause) plus the harness-routing contract pins. No file overlap; the two PRs are complementary and together clear every known main-red failure.

### Verification
- Reproduced `onboarding-wizard.spec.ts:338` failing on clean `origin/main` locally, then traced it to `aria-label="Summoning rite"` in `src/components/familiar-rite.tsx:1083`.
- Reproduced `chat-sidebar-nav.spec.ts:104` timing out at helper line 47, then confirmed the rename in `git log -p src/lib/nav-section.ts`.
- `tests/chat-sidebar-nav.spec.ts` passes locally against the fix. The onboarding spec's local runs are unreliable right now (a concurrent session is running its own Playwright suite and the shared `cave home reconciliation lock` times out under it), so CI is the authority here.


BunsDev:
Folded `fix/cave-oqawv-main-red` (#4388) into this branch. The two PRs were deadlocked: this one failed `Frontend build` and `Sidecar runtime` on the breakages #4388 repairs, while #4388 failed `E2E (Playwright)` on the two specs this one repairs. Neither could ever go green alone, so neither could un-red `main`.

Both sessions independently found the `"Code"` → `"Chat"` rail-tab rename, so `tests/chat-sidebar-nav.spec.ts` conflicted. Resolved toward the tablist-scoped locator (`getByRole("tablist", { name: "Workspace sections" })`) rather than #4388's sidebar-scoped one — the nav row *inside* the section is also named "Chat", so scoping to the sidebar alone leaves the match ambiguous. #4388's fuller rationale is kept in the comment.

#4390 now carries every known main-red fix: the harness-routing contract pins, the sidecar bundle/runtime-closure changes, the mobile bottom-tab scoping, and the two specs above. Once it merges, #4388 is redundant.

PR #4391, opened 2026-08-06
Docs/cave zs85n chat sidebar attention

BunsDev:
## Summary
Reconciles the chat sidebar/attention branch with current `main` and closes replay, stream, and Markdown-fence regressions discovered in review.

## Why
Replayed Cave conversations use a stable local ID while the daemon owns a different harness ID. That split must not break event traces or title compare-and-set writes. The chat marker pipeline must also keep protocol-looking Markdown examples inert without suppressing live controls after a valid fence closes.

## Changes
- Merged current `origin/main` and restored the selected-row siderail tick contract.
- Route owned replayed conversation event requests through their persisted `harnessSessionId`; use the displayed Cave ID for title revision lookup.
- Preserve non-user Gateway abort outcomes and terminally strip incomplete attention directives from failed streams.
- Reconciled renderer-aware Markdown fence parsing across GitHub, image, attention, next-path, quick-chat, and rendered-text consumers, including empty-fence coverage.

## Verification
- `pnpm test:app`
- `pnpm test:api`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm build`

## Risk
Event resolution remains authorization-first and validates the stored daemon ID before proxying. Fence handling is parser-sensitive, so the change adds coverage for list, blockquote, overlapping, and empty fence boundaries. Required GitHub checks are queued for this exact head.

PR #4401, opened 2026-08-06
fix(dev-app): resolve the dev port without ANSI colour contamination

BunsDev:
## The bug

`scripts/dev-app.sh` captured its dev port with:

```bash
dev_port="$(node -e "import('./scripts/ports.mjs').then((m) => console.log(m.resolvePort('dev', process.env)))")"
```

`console.log` renders a **Number** through `util.inspect`, which colourises it whenever `FORCE_COLOR` is set — agent harnesses and many CI runners set it routinely. So the capture returned `$'\e[33m3000\e[39m'`, not `3000`, and the escapes survived command substitution into every later use of `$dev_port`.

```
$ FORCE_COLOR=3 node -e 'console.log(3000)' | cat -v
^[[33m3000^[[39m
$ FORCE_COLOR=3 node -e 'console.log(String(3000))' | cat -v
3000
```

## Why it was misdiagnosed twice

`dev-app-origin-health.mjs` validates `--port` with `/^\d+$/`. A decorated port makes `cliArgs` return `null`, so the probe exits 1 **immediately**. The launcher then prints:

```
[dev:app] loopback origin on 127.0.0.1:<port> did not return the root document within 2s; desktop shell was not opened
```

That reads as a readiness timeout, but a `bash -x` trace shows it raised at **+0.1s**, having never waited at all — `origin_is_ready` swallows stdout and stderr, so a usage error is indistinguishable from an unready origin.

This was filed twice as a timing flake (cave-h73ee, cave-voglb). The suggested fix in both — raise the `waitFor` deadline — could not have worked, because no deadline was ever reached. It also explains the reported environment-dependence exactly: fails locally, green in CI, because CI does not set `FORCE_COLOR`.

**`pnpm dev:app` is broken the same way** in any `FORCE_COLOR` environment; the desktop shell never opens. That is the larger half of this bug — the failing test was the symptom that surfaced it.

## The fix

- Emit the port with `process.stdout.write(String(...))`. Strings are not colourised; Numbers are.
- Reject a non-numeric `$dev_port` at the point of capture, so a future decorator fails loudly instead of as a readiness timeout twenty seconds downstream.

## Verification

| | before | after |
|---|---|---|
| `node scripts/dev-app-teardown.test.mjs` | fails 3/3 | passes 3/3 |
| `pnpm typecheck` | clean | clean |
| `pnpm lint` | clean | clean |

The regression test is behavioural, not shape-based: it extracts the launcher's own one-liner from `dev-app.sh` and runs it under `FORCE_COLOR=3`, so it cannot drift from the script. Confirmed non-vacuous — restoring the old capture fails it with `actual: '\x1B[33m3000\x1B[39m\n'`.

## Scope

Every other `node` capture in `scripts/` emits a String (`process.platform`, `process.arch`, `encodeURIComponent`, `url.toString()`, `sidecar-target.mjs --sh`) and is unaffected. This was the only numeric one.

Fixes cave-h73ee. Closes cave-voglb as a duplicate.

### DIFF
--- a/scripts/dev-app.sh
+++ b/scripts/dev-app.sh
@@ -37,7 +37,22 @@ origin_is_ready() {
 }
 
 # The dedicated dev port, honouring COVEN_CAVE_PORT then PORT.
-dev_port="$(node -e "import('./scripts/ports.mjs').then((m) => console.log(m.resolvePort('dev', process.env)))")"
+#
+# Write the digits directly instead of console.log'ing the Number: console.log
+# renders a Number through util.inspect, which wraps it in ANSI colour codes
+# whenever FORCE_COLOR is set. Agent harnesses and many CI runners set it, and
+# those escapes survive command substitution — the contaminated value then
+# reaches --port, fails dev-app-origin-health's /^\d+$/ check, and the launcher
+# dies instantly while reporting a readiness timeout it never actually waited
+# for. Strings are not colourised; Numbers are.
+dev_port="$(node -e "import('./scripts/ports.mjs').then((m) => process.stdout.write(String(m.resolvePort('dev', process.env))))")"
+case "$dev_port" in
+  ''|*[!0-9]*)
+    echo "[dev:app] ERROR: resolved dev port is not numeric: $(printf '%q' "$dev_port")" >&2
+    echo "[dev:app]        Something is decorating this command's stdout (FORCE_COLOR, a shell wrapper, a Node loader)." >&2
+    exit 1
+    ;;
+esac
 if [ -n "${COVEN_CAVE_PORT:-}" ] || [ -n "${PORT:-}" ]; then
   echo "[dev:app] using overridden port ${dev_port}"
 else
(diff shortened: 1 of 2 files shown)


BunsDev:
End-to-end confirmation on the real launcher (not the unit test), same held port, `FORCE_COLOR=3`:

**Before — pristine main:**
```
[dev:app] using overridden port ^[[33m50891^[[39m
[dev:app] starting dev server on ^[[33m50891^[[39m
[dev:app] loopback origin on 127.0.0.1:^[[33m50891^[[39m did not return the root document within 180s; desktop shell was not opened
```

**After — this branch:**
```
[dev:app] using overridden port 50891
[dev:app] starting dev server on 50891
[dev:app] warning: 127.0.0.1:50891 is still listening after teardown
```

Note the third line before the fix. It claims a **180s** wait, but the whole invocation ran under a 60-second `timeout` and completed inside it — so the launcher never waited at all. That is the misleading message this PR removes, reproduced live rather than inferred.

The "still listening after teardown" warning after the fix is correct: the port is held by a foreign listener the launcher does not own.

---

CI note: this branch got **zero** `pull_request` check runs when opened (only the Copilot reviewer). Same for #4400 and #4391. `gh workflow run CI --ref <branch>` delivers them correctly and they attach to the head SHA. Tracked as cave-0udfz.

PR #4408, opened 2026-08-07
docs(nav): implementation prompt for tracking every navigation level

CompleteDotTech:
## Problem

The shell's Back/Forward controls (`DesktopHistoryNav`) traverse **two** levels:

1. Workspace mode — `workspace.tsx:358`
2. Chat session id, inside Chat only — `workspace.tsx:363` + `chat-router.tsx:334`, the app's single `history.pushState`

Everything else is component-local `useState` or a `replaceState`-only hash — addressable, but creating no history entry. Chat's `Sessions / Projects / Canvas / Familiar` strip (`chat-surface.tsx:138`) is the reported case: Back from Canvas exits Chat entirely instead of stepping to Projects.

## What this PR adds

`docs/nav-history-full-tracking.md` — documentation only, no code touched. An implementation brief for tracking **every** level.

- **Model** — a `WorkspaceLocation` entry (`mode` + ordered `path` + `axes` + `overlay`), and the one change needed in `workspace-navigation-history.ts`: its `===` identity comparisons take an injected `keyOf` so the stack can hold objects. Its push/move/restore/replace algebra is already generic over `T` and is kept as-is.
- **`useSurfaceHistory` hook** — surfaces register a level; user-initiated changes push, traversal / deep-link hydration / `localStorage` restore replace, unmounting a level truncates the path below it.
- **Conversion table** — every level anchored to file:line, across Chat, Settings (section → sub-tabs → studio tabs → mobile drill-down), Board, Inbox/Rituals, Marketplace, Grimoire, GitHub, Browser, Researcher desk, Familiars view, Submissions, and the Inspector pane.
- **Overlays** — 17 modals/drawers/sheets/palette become `overlay` entries where Back closes rather than navigates, with Escape routed through `history.back()` so the paths can't desync. Onboarding, the voice-call overlay, and the confirm dialog are excluded by design.
- **Phasing, tests, acceptance criteria** — nine phases starting with a no-behaviour-change refactor; E2E plan honours the daemon-less `COVEN_CAVE_E2E=1` constraint.

## Relationship to #4407

#4407 is the read-only inventory. This supersedes it on one judgement call: that document marked filter chips and A/B toggles as deliberately untracked. Per the request, they're tracked here too.

One thing worth a look before implementing: tracking filter axes means four filter taps cost four Back presses. The brief specifies **same-axis coalescing** (consecutive same-axis changes within ~700 ms collapse to one entry, per-axis configurable, `0` to disable) so bursts don't turn Back into a per-tap undo. Deliberate, spaced-out changes still each get an entry.

## Invariants preserved

- Traversal never escapes into unrelated webview history (`workspace.tsx:360`).
- Deep links and restored state replace rather than synthesizing a phantom previous entry.
- `MODE_ALIASES` produce a `{ mode, tab }` entry directly instead of a mode entry plus a silent tab set.

### DIFF
--- a/docs/nav-history-full-tracking.md
+++ b/docs/nav-history-full-tracking.md
@@ -1,0 +1,272 @@
+# Full navigation history tracking — implementation prompt
+
+**Goal:** the shell's Back/Forward controls traverse **every** navigable level of
+the app, not just two. Back steps up exactly one level from wherever the user
+is; Forward retraces. No level is exempt.
+
+Companion to `docs/nav-history-tracking.md` (PR #4407), which is the read-only
+inventory. That document marked filter/toggle axes as deliberately untracked.
+**This document supersedes that judgement:** every axis listed below is tracked,
+including filters and A/B toggles.
+
+---
+
+## 1. Current state
+
+Two levels are tracked, and only two:
+
+| Level | Where | Mechanism |
+| --- | --- | --- |
+| Workspace mode (`CaveMode`) | `src/components/workspace.tsx` — `navigationHistoryRef` (:358) | in-memory stack, `src/lib/workspace-navigation-history.ts` |
+| Chat session id (Chat only) | `workspace.tsx` — `chatNavigationHistoryRef` (:363) + `src/components/chat-router.tsx:334` | the app's **only** `history.pushState`; `#chat-<id>` hash |
+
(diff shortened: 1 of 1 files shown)


CompleteDotTech:
Consolidating into #4407, which now carries the implementation as well as the inventory. The full-tracking conversion table from this PR moves there so the remaining phases stay documented.

PR #4371, opened 2026-08-05
feat(summoning): summon a familiar from a likeness — image-driven rite with a holographic card

lmvdz:
OpenCoven Beta Hackathon 2026 contribution / Team: Lars / Submission repository: https://github.com/lmvdz/coven-cave

Implements #4370 — read that issue for the design rationale; this describes the change itself.

Paste or drop an image, a local vision-capable harness reads it, and the familiar assembles as a holographic trading card while you answer three questions. Every scried value is an editable suggestion, never a silent commit.

## What lands

| | |
|---|---|
| `POST /api/scry` | Takes an image, picks the first **local** vision-capable harness from `/api/harnesses`, spawns it directly, streams stage + suggestions back over SSE. **No familiar required** — scrying needs a harness, a model and an image path, not a `familiarId`. |
| `src/lib/foil` | Deterministic halftone/SDF plate generator. Isomorphic: identical maths for the in-browser live preview and server-side print resolution. Seeded from identity only, so one seed composes identically at any resolution. |
| Summoning rite | Card left, one question right. Picking advances — no Next button. Vessel tiles, model stones sized by real context window, multi-select office sigils. Everything after the likeness is skippable. |
| Familiar card | Full-art, material-aware foil, flip to a scannable QR seal. |
| Fullscreen card | Opened from a familiar's avatar. Drag to spin, hover for foil. |
| Scaffolder | `purpose` is now its own scried field; the portrait description no longer lands in the "My purpose is to …" slot. |

The existing Summoning Circle is **untouched and still reachable** — it binds SSH, OpenClaw and Hermes-profile vessels that the rite does not, so the rite offers an explicit "summon by hand" escape rather than stranding those users.

## Safety properties, each with a test

- **Generated content cannot forge a contract file.** Scried values are sanitised before reaching `SOUL.md`/`IDENTITY.md`; values shaped like headings, ward sections, TOML assignments or fenced blocks are rejected whole. `evaluateFamiliarContract` still scores zero violations *and* zero warnings, proven by the existing test running the real validator against generated and hostile input.
  - This also closed a **pre-existing** hole: `role` and `description` were already interpolated raw, so a description containing `\n## I am Root` forged a second declared name. Found because a hostile fixture went through the real create route rather than the scaffolder alone.
[...]

PR #4359, opened 2026-08-05
fix(onboarding): allow cold managed npm startup on Windows

BunsDev:
## Summary
- keep the lightweight managed `node --version` probe capped at 1.5 seconds
- allow `npm-cli.js --version` 10 seconds for Windows cold startup
- add injected-runner regression coverage for the distinct timeout budgets
- document the scoped design and implementation plan

## Validation
- `node --require ./scripts/css-source-contract-hook.cjs --experimental-strip-types src/app/api/onboarding/install/route.test.ts`
- `node --require ./scripts/css-source-contract-hook.cjs --experimental-strip-types src/lib/server/managed-node-toolchain.test.ts` (8/8)
- code review: no significant issues

## Existing main blockers
- `pnpm test:api` reaches the pre-existing `scripts/worktree-guard.test.mjs` failure, reproduced unchanged on clean `main`
- `pnpm typecheck` reaches pre-existing conflict markers in `src/components/sidebar-minimal.tsx` (PR #4358) and the existing `src/components/canvas-editor.test.ts` syntax error; neither file differs in this branch

Closes #4355

PR #4413, opened 2026-08-08
fix(windows): stop console windows popping up over the app (cave-7jb)

CompleteDotTech:
## The bug

On Windows, running research in the Research Desk pops black terminal windows
over the app — one per familiar step, so seven per mission iteration.

## Cause

The Tauri shell starts the Node sidecar with `CREATE_NO_WINDOW`
(`src-tauri/src/sidecar_startup.rs:415`), so the server process running our
route handlers has **no console attached**. Under Win32, `CreateProcess` on a
console-subsystem child from a console-less parent *allocates a new console* — a
real, visible conhost window — unless the child is given `CREATE_NO_WINDOW` too.
Node spells that flag `windowsHide: true`.

The harness spawn every research familiar step goes through
(`src/app/api/chat/send/route.ts`) built its options with `cwd`, `env` and
`shell: false`, but no `windowsHide`. Most other spawn sites were the same.

It reads as an omission rather than a decision, because the sites right next to
it already get it right — `chat-send-capabilities.ts:153` (taskkill),
`grok-compatibility.ts`, `claude-models.ts`, `daemon-start.ts`, and the
`coven-bin.ts` toolchain probes all pass it.

## The fix

`windowsHide: true` on every `child_process` call under `src/` — 70 sites across
45 files. It is a no-op off Windows, so there is no platform branch to get
wrong.

Three call sites forward an options object they do not own
(`NpmPrefixExecFile`, `execLatestVersion`, `SpawnProcess`). Their signatures now
*require* `windowsHide: true`, so the obligation moves to the caller and the
compiler checks it.

Two spawns had no options argument at all — including
`execFileAsync("powershell.exe", …)` in `process-intent-lock.ts`, which opened a
PowerShell window.

## The ratchet

`src/lib/child-spawn-window.test.ts` fails the build on any new call site that
omits the flag. It parses every `node:child_process` importer under `src/`,
blanks comments and string bodies so prose can't match, and resolves one level
of indirection for options passed by name or spread. Waivers are an explicit
three-entry list, each with a reason, and the test fails if a waiver stops being
needed.

This mirrors `src-tauri/release-runtime.test.mjs`, which already pins
`creation_flags(0x08000000)` on the Rust sidecar launcher.

Verified it catches a regression: deleting the flag from `chat-work-branch.ts`
fails the test with that exact file and line.

**What it does not cover**, called out in the test header: a child launched via
`shell: true` or a `.cmd`/`.bat` shim run by `cmd.exe`. That window belongs to
the shell, not the child. `covenLaunchCommandForBinary()` already resolves npm
shims to a direct executable, so the harness path is covered.

## Test note

One existing source-shape assertion in `api/harnesses/route.test.ts` pinned
`{ env,` as the *literal start* of the options object. It now matches `env`
wherever it sits, which is what it meant to assert.

## Verification

- `pnpm typecheck` — clean
- `pnpm lint` — clean
- New guard test passes; confirmed it fails when the flag is removed
- Sibling tests for all 45 changed modules, plus every source-shape test that
  reads a changed module, run individually

Six of those failed **identically on `origin/main`** with the change stashed —
all pre-existing Windows-environment failures (symlink creation, `chmod`,
looking for a PATH entry literally named `git` rather than `git.exe`, a fixture
directory with a trailing space). Only `api/harnesses/route.test.ts` was a real
regression from this change, and it is fixed here. CI on Linux is the real
signal for those six.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ breaking config changes; patch releases stay additive.
 
 ## [Unreleased]
 
+### Fixed
+
+- Windows: console windows no longer pop up over the app when Cave runs a CLI.
+  Every child process now launches with `windowsHide: true`. Most visible in the
+  Research Desk, where a single mission iteration opened one window per familiar
+  step (cave-7jb).
+
 ## [0.2.4] - 2026-08-06
 
 > A tabbed sidebar, the Coding Room rebuilt as a three-zone workbench, image
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -124,6 +124,7 @@ export const SUITES = {
     "src/lib/model-control-capabilities.test.ts",
     "src/lib/chat-transcript-groups.test.ts",
     "src/components/chat-view-chunk-coalescing.test.ts",
+    "src/lib/child-spawn-window.test.ts",
     "src/lib/chunk-coalescer.test.ts",
     "src/components/chat-view-scroll-pin.test.ts",
     "src/components/home/use-home-model-state.test.ts",
--- a/src/app/api/beads/prs/route.ts
+++ b/src/app/api/beads/prs/route.ts
@@ -35,6 +35,7 @@ const MERGED_PR_LIMIT = "30";
 
 async function ghPrList(repoRoot: string, args: string[]): Promise<unknown> {
   const { stdout } = await execFileAsync("gh", args, {
+    windowsHide: true,
     cwd: repoRoot, // cwd's repo → gh targets the right OWNER/REPO without hardcoding
     env: { ...caveToolSpawnEnv(), GH_PROMPT_DISABLED: "1" },
     timeout: GH_TIMEOUT_MS,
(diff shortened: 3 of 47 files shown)


BunsDev:
Heads up on the bead reference: `cave-7jb` did not exist in the Beads graph — the branch name and title cited an id that nothing resolved to. Same for three other recent PRs (`cave-0rw` on #4411, `cave-3rz` on #4371, `cave-xys` on #4359).

The cause looks mechanical rather than anyone's mistake: nothing had been pushed to `refs/dolt/data` since 2026-08-06 05:27, so the shared graph was ~80 Dolt commits behind and anyone reading it was working from a stale copy. That is now pushed.

`cave-7jb` has been created with that exact id, so this PR's branch and title resolve as intended. It carries the diagnosis from this description — the sidecar starting under `CREATE_NO_WINDOW` with no console attached, so `CreateProcess` on a console-subsystem child allocates a visible conhost unless the child also gets the flag — and records the fix as `windowsHide: true` across the 70 `child_process` sites, plus the three signatures tightened so a forwarded options object cannot drop it.

It is `in_progress` and should be closed when this merges.

PR #4417, opened 2026-08-08
feat(research): rebuild the Research Desk from the Research Desk App handoff

BunsDev:
Implements the Claude Design handoff **`Research Desk App.dc.html`** (project *Research Desk interface redesign*, `0b43b32c`) across the whole `researcher-desk` surface — the frame's second, much larger pass over a room whose chrome landed in `cave-mxqz`.

Tracked by `cave-na7oc`. Ledger row added to `docs/design-handoff/IMPLEMENTATION-STATUS.md`.

## What changed

**Prompt** — the composer stops being a textarea and becomes a brief you can see: a five-element strength meter, a prompt-builder dialog, an assembled-brief strip that reads those elements back out of whatever is typed, ⚡ recommendations, and Quick saves as a grouped drawer docked to the bottom edge.

**Desk** — both rails drag-resize and collapse to a labelled spine. The six-phase stepper now reports what each phase *found* (source counts, conflicts, artifacts, "your turn", published counts) over a wash that widens as phases settle; bound readings gain fill bars; pass dots join the pass count.

**Library** — the static "Sorted by newest" caption becomes a real sort control, joined by search and a pager.

**Studio** — a provider readiness strip beside the heading, so the reason a card is disabled is visible before you click it, plus the media mosaic.

**Resources** — kind-banded cards.

**Shell** — the frame's engine / live / review status cluster on the tab strip.

## The honesty line

The frame invents data this app does not have. Where that happened, the shape was re-derived from something real rather than shipped as decoration:

- its four recommendation cards → derived from the live mission list (a stopped run to retry, a checkpoint to split out, a finished run to extend, a run short of its source target). Every card names the mission it came from; no missions, no cards.
- its five media providers → the four the readiness endpoint actually reports, each blocked chip carrying that endpoint's own hint verbatim so the chip and the card it blocks say the same string.
- its hard-coded quick-save clusters → the saved-link categories the store already carries.

**Not adopted**, because each would need a backend that does not exist and would otherwise be a surface that lies: inline run rename (no rename action), the report's per-section approve/comment flow and inline citation chips (artifacts are files, not sectioned documents), per-question evidence coverage, and the version-compare dialog.

One deliberate departure: the frame's 2×2 podcast hero tile. `RESEARCH_GENERATION_CREATABLE_KINDS` is a live list, so a hand-tuned span plan leaves a half-empty row the moment a kind is added — and equal tiles stop the layout claiming one kind matters more than another when nothing in the model says so.

## Verification

- `pnpm test:app` — 1127 files pass; `pnpm test:api` — 345 files pass
- `tests/research-desk-tabs.spec.ts` — 10 pass (updated for the Quick saves drawer; now also pins the suggested-match group and the attached-count readout)
- `pnpm lint` (design ESLint gate + TSX codemod check) clean; `tokenize-css.mjs` a no-op
- Token drift: spacing and radius ratchets hold — the handoff's 5/6/7/10/11/26/28px paddings and gaps are snapped to `--space-1..-6`, keeping only the banked 1/2/3px micro-marks. `inlineTsxStyles` is raised by 3 with justification in the baseline: `--research-progress`, `--research-bound-fill` and `--research-queue-width` are per-mission or per-user measurements that exist only at render time.
- **Driven in a real browser** through the desk's own mocks at 1440px and 720px, open and collapsed. That is where the three defects in `b0c7df7` came from — all three passed every test first.

New pure modules ship with behaviour tests: `research-prompt-brief` (14), `research-quick-saves` (9), `research-library-view` (12), `research-studio-providers` (6), plus 10 new cases on `research-missions`.

## A11y

Rail separators are focusable `role="separator"` with arrow/Home/End — a drag handle no keyboard can reach is a rail no keyboard user can size. Drag uses pointer capture, so releasing outside the window ends it. Stored widths clamp on read as well as write, so a stale value cannot wedge a rail off screen. Colour is never the only channel: the strength meter carries a tone word and an "add: …" tail, provider chips carry an sr-only detail, scope dots sit beside counts, and stepper nodes keep their sr-only status. Every new animation has a `prefers-reduced-motion` story.


BunsDev:
Both review findings were real and are fixed in `7c3ed31da1`.

**`N/0 src` in the phase meta** — a zero `sourceTarget` is a real (if legacy) bound that the composer's clamp cannot reach on already-stored missions, and `3/0` reads as a broken divide. `researchPhaseMeta` now reports the count alone when there is nothing to divide by, pinned by a new case in `research-missions.test.ts`.

**`use-research-pane`'s "both Desk rails" claim** — correct that only the run queue was wired. Rather than weaken the comment to match the code, the evidence rail is now genuinely resizable: its own persisted pane with an **inverted** drag direction (it grows leftward, so sharing the queue's `+1` would move it the wrong way under the cursor), the same keyboard parity on its separator, and a grid track driven from the dragged width. The separator withdraws in focus mode and on narrow desks, where the layout stacks and a horizontal drag means nothing.

Re-verified after both fixes: `pnpm lint`, the token-drift ratchets, `research-desk-tabs.spec.ts` (10 e2e), and `research-tab-desk.test.ts` (27, now pinning the rail wiring).

PR #4419, opened 2026-08-08
fix(research): Resources grid strands single-item groups

BunsDev:
Fixes `cave-93jz1`, found while driving the merged Research Desk (#4417) in a production build.

## The bug

`surface-research-resources.css` used `repeat(auto-fill, minmax(250px, 1fr))`. **`auto-fill` creates and keeps empty tracks**, so a group holding fewer links than the row fits leaves its cards at the 250px minimum beside a wide dead area. Most visible at one link per group — which is the common case for "In this run".

Groups with 5+ links filled the row and looked correct, which is why it survived this long.

## The fix, and why it needed two parts

`auto-fit` collapses the empty tracks. On its own that overcorrects: one saved link would stretch into a ~1300px card. So the card also carries a cap.

Measured in a production build at 1440px:

| | lone card | packed row |
|---|---|---|
| before | 255px, beside ~1000px dead space | 255px each |
| after | **360px** | 255px each (unchanged) |

360px sits just above the widest track a packed row produces, so a lone card reads as *the same component* as one in a full group rather than as a banner.

## The cap is scoped twice, deliberately

Both escapes would re-create the very gap this removes, so each is pinned by the test:

- **`[data-view="grid"]`, never the bare card** — rows view is a flex column whose cards span the surface (measured 1314px). A cap there would shrink every row to a quarter width.
- **Lifted at the 560px breakpoint**, where the grid is already one column (measured 496px filling a 520px viewport).

## Not a regression from #4417

The `auto-fill` rule predates the Research Desk rebuild. The new kind-banded card header just makes each card read as a more definite object, so the short row became conspicuous. Calling that out so nobody bisects into the redesign looking for it.

## Verification

Driven in a real browser against a production build, with fixtures giving one group a single link and another seven — the exact comparison the bead asks for. All three cases measured, not eyeballed: grid 1440 (360 vs 255), rows 1440 (1314, uncapped), grid 520 (496, filling).

`pnpm lint`, the token-drift ratchets and `tokenize-css.mjs` are all clean — `max-width` is not a ratcheted property, and no baseline moved. The new test in `research-tab-resources.test.ts` pins `auto-fit`, the exact cap, and both scope boundaries, so a future `auto-fill` or an unscoped cap fails CI rather than quietly regressing.

### DIFF
--- a/src/components/role-surfaces/research-tab-resources.test.ts
+++ b/src/components/role-surfaces/research-tab-resources.test.ts
@@ -238,3 +238,27 @@ test("manual zero-result announcements suppress duplicate EmptyState live output
   assert.match(emptyState, /live = true/);
   assert.match(emptyState, /role=\{live \? "status" : undefined\}/);
 });
+
+test("the grid collapses empty tracks, and the card cap is grid-only (cave-93jz1)", () => {
+  // auto-FILL keeps the empty tracks it created, so a group holding fewer links
+  // than the row fits left its cards stranded at the 250px minimum beside dead
+  // space. auto-fit collapses them.
+  assert.match(styles, /\.research-res__items\[data-view="grid"\] \{[\s\S]*?repeat\(auto-fit, minmax\(250px, 1fr\)\)/);
+  assert.doesNotMatch(
+    styles,
+    /\.research-res__items\[data-view="grid"\] \{[\s\S]*?repeat\(auto-fill/,
+    "auto-fill strands single-item groups — see cave-93jz1",
+  );
+  // …but collapsing alone would stretch ONE saved link across the whole row,
+  // so the card carries a cap. Measured: a lone card renders 360px against
+  // 255px in a packed row — recognisably the same component.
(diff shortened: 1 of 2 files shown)


BunsDev:
Good catch in `c222f4d3c5` — and the problem was worse than reported.

The unbounded `[\s\S]*?` fails in **both** directions: it can false-FAIL (an unrelated `auto-fill` added later in the sheet reads as this rule regressing, blaming the wrong line) and it can false-PASS (a later `auto-fit` satisfies the positive check even if this rule had been reverted to `auto-fill` — precisely the regression the test exists to catch).

Each assertion now slices the one rule body it is about. The base grid rule is identified by the `display: grid` it declares, so the slice cannot silently pick up the 560px override that shares its selector; the override is read by searching from the start of that container block.

Verified by breaking it on purpose rather than trusting the green: restoring `auto-fill` fails the test, unscoping the cap onto the bare card fails it, and both pass again once reverted.

PR #4427, opened 2026-08-08
fix(test): resolve file:// URLs with fileURLToPath so two guards run on Windows

CompleteDotTech:
Closes cave-01n.

## What was wrong

A `file://` URL's `.pathname` is `/C:/Users/…` on Windows. Handed to a filesystem API, that leading slash is an absolute path from the drive root, so two test files threw on their first statement rather than running:

| File | Symptom on Windows |
|---|---|
| `src/server-pty-ws.test.ts` | `Could not resolve "/C:/Users/…/server.ts"` from esbuild |
| `src/components/dnd-context-stable-ids.test.ts` | `readdirSync` on `C:\C:\Users\…\src\` |

Reproduced on `main` at 87c714ded, so it predates the branch it was found on.

## Scope of the impact — CI was never affected

Worth stating plainly, because the obvious guess is wrong: both files live in the `app` suite, and `Frontend build` (which runs `pnpm test:app`) is `runs-on: ubuntu-latest`. The `Cross-environment (windows-latest)` matrix runs only the three files in the `conformance` list. So these guards have always run, and always passed, on the only platform CI runs them on.

The cost is local: a contributor on Windows cannot run `pnpm test:app`, because it aborts on `server-pty-ws.test.ts` before reaching anything after it. That is a real tax on anyone developing the desktop shell from the platform the desktop shell targets, but it is not a hole in the required checks, and this PR should not be read as closing one.

## The fix

`fileURLToPath`, which is the whole reason it exists. The dnd guard also normalizes separators in its offender list so a failure reads as `src/components/foo.tsx` on every platform rather than mixing them.

## Deliberately not touched

`scripts/beads-pr-bridge.test.mjs` and `scripts/beads-pr-patrol.test.mjs` use the same `.pathname` idiom, but both `process.exit(0)` on `win32` by design — their fixtures stub POSIX executables — so the line is never reached. Changing them would be churn.

## Verification

Run natively on Windows 11 (`win32/x64`, node 24.18.0), where both previously threw:

```
$ node --experimental-strip-types src/server-pty-ws.test.ts
server-pty-ws.test.ts OK

$ node --experimental-strip-types src/components/dnd-context-stable-ids.test.ts
dnd-context-stable-ids: OK

$ pnpm typecheck
(clean)
```

The dnd guard's failure path was also exercised directly by dropping a temporary `<DndContext someProp>` into `src/components/`, confirming it both catches the offender and reports it as `src/components/__tmp-dnd-probe.tsx` rather than a mangled or backslashed path.

Both files are assertion-only guards over source text — no runtime behavior changes, and Linux behavior is untouched since `fileURLToPath` and `.pathname` agree on POSIX. The Linux CI run on this PR is therefore the regression check that the fix changes nothing there.


CompleteDotTech:
@codex review this.

PR #4442, opened 2026-08-08
test(e2e): wait for citation anchors and chips separately, with real budgets

BunsDev:
**`main` is red on `E2E shard 3/3`** from `79a061fb6c` (#4436) onward, and every
PR inherits it because CI builds the merge commit. This unblocks it. `cave-nrptt`.

## What #4436 got right and what it missed

#4436 replaced a prose wait with a 30s wait for `a.cave-citation-chip`, on solid
reasoning: prose is a proxy for a milestone it does not reach. The milestone was
right. The budget was not.

Playwright's `error-context` DOM at the 30s mark holds the reader's own citation
anchors — `#cite-1`, `#cite-2`, inside the reader region — carrying neither the
`aria-label` nor the `aria-haspopup` that `citation.tsx` adds beside the class.
So the footnote pipeline **had** run and produced anchors; only the client-side
decoration had not.

Reproduced locally, not inferred:

```
rm -rf .next && CI=true COVEN_CAVE_E2E=1 pnpm exec playwright test --shard=3/3
```

One reader test fails per cold run. The prose wait settles, and the anchors then
need **more than another 30s** — so prose and anchors are not one render, and
what is being measured is `next dev` compiling the citation chunk on demand.

## The change

- Wait for the **anchors** (90s), then the **chips** (45s), instead of the end
  state alone.
- File-wide timeout `240s → 360s`, to stay above the inner sum of 315s.

Splitting the wait is worth as much as the budget. A single chip wait cannot
distinguish *"the pipeline never emitted anchors"* from *"the anchors are there,
undecorated"* — and those need opposite fixes. The first run of this fix failed
on the chips and the second on the anchors; under the old wait both would have
reported identically, which is precisely what made this look like an
unpredictable flake.

## The timeout invariant

The outer timeout must exceed the sum of `openReader`'s own waits, or it fires
mid-wait and kills the allowance an inner wait exists to provide — surfacing as
`element(s) not found`, a product failure rather than the compile it is. That
already cost `cave-n7wm5` its diagnosis at 180s. The header now records the sum,
says to recompute it when any wait changes, and notes that the `configure` is
the single place to set it: these tests are top level, so a per-test
`test.setTimeout()` intended as headroom silently **lowers** the file-wide value.

## Deliberately not claimed

I do not know why the *same single test* reds each run. Serial mode runs the
file in order on one worker and the failing test is mid-file, so "the first test
pays the compile" does not explain it. I wrote that explanation first, could not
support it, and removed it. The budget is sized to the measurement, not to the
theory.

### DIFF
--- a/tests/reader.spec.ts
+++ b/tests/reader.spec.ts
@@ -22,15 +22,23 @@ import { expect, test, type Page } from "@playwright/test";
 // navigation — which reads as "the reader is broken" when it is only cold.
 // Serial lets the first test pay the compile and the rest run warm. Same
 // posture as canonical-memory / code-surface / research-desk-tabs.
-// 240s, not 180s: openReader's own waits already budget 30s (chat surface) +
-// 30s (the answer) + 90s (the reader chunk) = 150s, and the FIRST test also
-// pays cold navigation and compile on top. At 180s the outer timeout fired
-// mid-wait and killed the very allowance the inner 90s exists to provide —
-// observed as `.cave-reader` "element(s) not found" with "Test timeout of
-// 180000ms exceeded" in the call log, on the first test only, passing on retry
-// (cave-n7wm5). Serial mode means only that first test is cold, so this buys
-// headroom for one test rather than lengthening the suite.
-test.describe.configure({ mode: "serial", timeout: 240_000 });
+// The budget must exceed the SUM of openReader's own waits, or the outer
+// timeout fires mid-wait and kills the very allowance an inner wait exists to
+// provide — reported as a product failure ("element(s) not found") rather than
+// as the compile it actually is. That happened at 180s and cost cave-n7wm5 its
+// diagnosis, on the first test only, passing on retry.
+//
+// Current inner budget: 30s (chat surface) + 30s (the answer) + 90s (the
+// reader chunk) + 30s (the prose) + 90s (the citation anchors) + 45s (the
+// decoration) = 315s, and the FIRST test also pays cold navigation and compile
+// on top. 360s covers that; 240s no longer did once the citation waits were
+// added (cave-nrptt). Recompute this sum whenever a wait above changes — the
+// failure it prevents looks like a product bug, not a budget one.
+//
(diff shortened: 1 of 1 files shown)


BunsDev:
Closing as superseded by #4441 (`cave-t69ok`), which is the better fix.

This PR treated the symptom. #4441 treats the cause: it moves citation wiring to the layout phase and reconciles late or replaced markdown anchors with a subtree observer, so the decoration no longer misses anchors that arrive after the effect first runs. It also removes the chip wait from the shared `openReader` helper and keeps citation readiness in the dedicated citation test, which is where that assertion belongs.

Merging this on top would work against that: it re-adds a wait to the shared helper that #4441 deliberately removed, and raises the file-wide timeout to 360s to accommodate a delay that should no longer exist. That is exactly the kind of budget inflation that hides the next real regression.

The diagnosis here still corroborates #4441 rather than competing with it. Playwright's error-context DOM at the 30s mark held the reader's own anchors — `#cite-1`, `#cite-2` — carrying neither the `aria-label` nor the `aria-haspopup` that `citation.tsx` adds beside `.cave-citation-chip`. Anchors present, decoration absent, which is precisely the late-anchor reconciliation #4441 implements.

What I got wrong, recorded so the next pass does not repeat it: I read that evidence correctly and then stopped at the test budget instead of following it into `citation.tsx`. The DOM was already saying the effect was not observing late anchors.

Two observations from this branch worth keeping regardless of which fix lands:

1. Waiting on the anchors and the chips as separate steps names the culprit. A single chip wait cannot distinguish "the pipeline never emitted anchors" from "the anchors are there, undecorated" — my first run failed on the chips and the second on the anchors, and under one wait those report identically. If a reader wait is ever reinstated in the shared helper, split it.
2. The file-wide `test.describe.configure()` timeout must exceed the SUM of `openReader`'s inner waits, and a per-test `test.setTimeout()` on this top-level file silently LOWERS it rather than raising it. I hit that myself.

Branch and commit `c827e7c5ad` stay on the remote; nothing here is lost if any of it is wanted later.

PR #4499, opened 2026-08-10
feat(worktree): name worktrees wedged in an abandoned merge

BunsDev:
Closes `cave-97svy`.

## The signal

A worktree left mid-merge blocked its branch for three days and was reported repeatedly as a persistent blocker. Every session that found it did the right thing — saw 137 uncommitted paths, read them as another session's in-flight work, and backed off rather than clobber a colleague.

Backing off was correct each time. The missing information was whether anyone was still there.

## Root cause

Nothing in `src/lib` or `scripts` references `MERGE_HEAD`, `rebase-merge`, or `CHERRY_PICK_HEAD` — verified by grep before this change. So a paused git operation has no name, no age, and no remedy anywhere in the tooling:

- the **lifecycle patrol** classifies on dirty/retention/PR state,
- the **auto-lock hook** locks anything dirty,
- the **retention-push hook** pushes unpushed commits,
- the **status dashboard** printed `137 dirty`.

All four are satisfied by a wedged worktree. It looks exactly like live work, forever.

`scripts/worktree-status.mjs` was also documented in neither `AGENTS.md` nor `CLAUDE.md`, so the one tool positioned to answer this was not something a session would think to run.

## The fix

A `WEDGED` verdict on the local dashboard that detects merge / rebase / cherry-pick / revert / bisect markers, dates the stall from the marker mtime, lists the unresolved paths, and — the part that unblocks a decision — compares every dirty path's mtime against the moment the operation stalled:

- **nothing tracked touched since** → no hand resolution exists; the tree is raw merge output both parents reproduce, and abort is lossless
- **something tracked touched since** → a human is mid-resolution; the report says `do NOT abort` and marks the abort command *"only with the owner's say-so"*
- **unreadable tree** → fails closed to the same warning, since "nobody touched it" is the fact the abort decision leans on

Retention of both merge parents is reported alongside (network-free, via remote-tracking refs and tags), so an abort is only called lossless when nothing would be stranded; otherwise the archive-tag command prints instead.

`WEDGED` outranks every verdict including `PRIMARY`, and is never `SAFE-RETIRE` — so neither `--prune` nor `wt:retire-on-exit` can remove a tree holding unfinished work.

## Verification

Against the real wedged worktree, the tool independently reproduces the manual forensics that took ~20 minutes:

```
🟠 WEDGED — docs/cave-zs85n-chat-sidebar-attention
   merge paused 2d ago (2026-08-07T23:50:31.967Z), marker MERGE_HEAD
   unresolved paths (5):
     src/app/api/chat/send/chat-send-capabilities.ts
     ...
   ✓ no tracked file touched since the merge stalled (1 untracked file(s) added since) —
      the tree is raw merge output with no hand resolution to lose.
      Both sides are reachable from a remote-tracking ref or tag, so an abort strands nothing.
   finish it:  cd '…' && git commit
   or drop it: cd '…' && git merge --abort
```

`node --test scripts/worktree-status.test.mjs` — 11/11 pass (6 pre-existing, 5 new covering: mid-merge is WEDGED not DIRTY; abandoned reports no hand resolution; hand-resolved warns against aborting; a paused rebase reports rebase-shaped remedies and survives the detached HEAD it creates; `--prune` never offers a WEDGED tree however merged it looks).

Docs added to `CLAUDE.md` (full section) and `AGENTS.md` (pointer at the "is this someone's live work?" decision point).


BunsDev:
Triage: **this PR is not broken — it's blocked by a break on `main`.**

The only failing check is `iOS build`:

```
apps/ios/CovenCave/CovenCave/Views/ConnectionView.swift:390:9: error: switch must be exhaustive
```

This branch does not touch `apps/ios` at all (`git diff --stat origin/main...HEAD -- apps/ios/` is empty). `AppModel.DiscoveryOutcome` gained a `credentialFailure(String)` case on `main` without `ConnectionView` being updated.

Fix is up as #4510, verified with a real local `xcodebuild`: `main` fails, the fix builds. Once that lands, rebase or re-run and this should go green.

Context for why it appeared here at all: the iOS job no longer runs on pull requests (it moved to `release.yml`), so it only surfaced because this PR's CI came from a `workflow_dispatch` recovery run, which builds unconditionally.

PR #4501, opened 2026-08-10
ci: take Windows off the pull_request fan-out; require seven contexts

BunsDev:
## Why

Every pull request scheduled **19 jobs, three of them Windows**. On 2026-08-10 this repository held **39 queued Actions runs with ZERO in progress**, the oldest stuck since **2026-07-18**, while GitHub Actions was globally operational and a hosted runner completed an unrelated job mid-stall. The queue — not the suite — was the blocker, so every required check on every open PR sat pending indefinitely.

macOS was already pulled from these matrices for exactly this reason (the comment is still in `ci.yml`). Windows now follows it.

## What changed

| | pull_request | push to `main` | release |
|---|---|---|---|
| `conformance` | ubuntu only | ubuntu + windows | — |
| `sidecar-runtime` | ubuntu only | ubuntu + windows | — |
| `windows-native` | skipped | runs | — |
| Windows MSI | — | — | built by `release.yml` |

**`main` keeps identical platform coverage.** Only the per-PR fan-out shrinks, by three jobs billed at 2x.

## The two ways this wedges every PR, both handled here

1. **`sidecar-runtime-required` failed unless `windows-native` reported `success`.** It now accepts `skipped` **on `pull_request` only**, and still hard-fails a skip on any other event — that is the event whose coverage the rollup exists to guarantee. Both rollups are already `if: always()`, so they keep reporting.

2. **Branch protection required the two `(windows-latest)` contexts**, which no longer report on a PR. A required context that never reports is exactly the `BLOCKED`-with-nothing-failing wedge documented in `CLAUDE.md`. Protection must drop to the seven contexts that always report — see the checklist below. The two `*-required` rollups stay required, so Windows still gates `main`.

Required contexts **9 → 7**: drop `Cross-environment (windows-latest)` and `Sidecar runtime (windows-latest)`.

## Verification

- `scripts/ci-recovery-workflow.test.mjs` extended to pin the new shape. Its exact-string SHA-guard assertion becomes **structural**: a job may narrow itself further, but only by ANDing onto the guard, and a top-level `||` in the suffix is rejected so the head-moved protection cannot be silently weakened.
- **Mutation-tested** — each of these fails a distinct assertion: reverting the `windows-native` PR skip; letting the rollup accept a skip on any event; swapping the `&&` for `||`.
- Passing: `ci-recovery-workflow`, `branch-to-merge-contract` (20), `ci-recovery` (17), `ios-build-ci`, `branch-cap-workflow`, `beads-familiar-workflow`, `dependency-policy`, `ui-consistency`, `check:tests-wired` (1603 files wired).
- `ci.yml` and `release.yml` both parse; release confirmed to build `windows-latest` (MSI) and both macOS arches.

Docs updated in the same change: `CLAUDE.md`, the `branch-to-merge` skill, and the `dependency-policy` comment noting the sharp-skew guard is now the *only* pre-merge signal for that class of defect.

## Merge checklist

- [ ] Merge this PR through the current nine contexts.
- [ ] Immediately update protection to the seven contexts above, or subsequent PRs will sit `BLOCKED` with nothing failing.

```bash
gh api -X PATCH repos/OpenCoven/coven-cave/branches/main/protection/required_status_checks   -f 'contexts[]=Frontend build' -f 'contexts[]=Rust check'   -f 'contexts[]=E2E (Playwright)'   -f 'contexts[]=Cross-environment (ubuntu-latest)'   -f 'contexts[]=Cross-environment required'   -f 'contexts[]=Sidecar runtime (ubuntu-latest)'   -f 'contexts[]=Sidecar runtime required'
```

## Follow-on

This is a safe subset of the approved [release-candidate CI design](docs/superpowers/specs/2026-08-05-release-candidate-ci-design.md) (`cave-7kix8`, "approved design, not yet implemented"), whose Phase 1 reduces enforcement further to a **single `PR checks` context** once the release-candidate promotion gate exists. That larger change is not attempted here.

Bead: `cave-jtuu7`


BunsDev:
Triage (queue sweep, not a review): the failure here is a **deliberate guard test firing against this PR's own intent**, so it needs an author decision rather than a rerun.

`Frontend tests (app)`:

```
✖ Windows conformance runs the native harness parser and DryRun fixture
  AssertionError [ERR_ASSERTION]: the conformance matrix must retain a real Windows runner
```

Something in the repo asserts the conformance matrix keeps a real Windows runner, and this change moves Windows coverage off it. That's exactly the kind of regression the test was written to block — so either the guard's premise is now obsolete and should be updated in this PR (with the reasoning recorded), or the change needs to keep a Windows runner in the matrix.

`Frontend build` is failing too and is the sole required check, so it must go green regardless.

BunsDev:
Closing as superseded by `main`, not as stale.

This PR's premise no longer holds:

1. **The Windows fan-out is already gone.** `#4503` (`ci: reduce routine validation fanout`) and `#4516` landed on `main`, and `.github/workflows/ci.yml` there now contains no `windows-latest` matrix entry and no `windows-native` job at all. Cross-environment, sidecar-runtime and Windows-native validation moved to `release.yml`.
2. **The required-contexts checklist is already satisfied, and then some.** Branch protection now requires exactly one context:

   ```
   $ gh api repos/OpenCoven/coven-cave/branches/main/protection --jq .required_status_checks.contexts
   ["Frontend build"]
   ```

   Nine did not drop to seven — it dropped to one. The two `(windows-latest)` contexts this PR asked to remove are long gone, so the `BLOCKED`-with-nothing-failing wedge it guards against cannot occur through them.

The branch is 65 commits behind and conflicts in five files including `ci.yml` itself, `CLAUDE.md`, and two test files — resolving those would mean re-deriving a workflow that has since been rewritten twice.

The analysis was right and it did land, by another route. If any specific assertion from `scripts/ci-recovery-workflow.test.mjs` here is not present on `main` — particularly the structural SHA-guard check that rejects a top-level `||` in the suffix — that is worth re-landing on its own as a small PR against current `main`.

PR #4506, opened 2026-08-10
test(daemon): repeated endpoint churn against real sockets

BunsDev:
Third increment of `cave-58eoq.4`. Covers the bead's *"repeated lifecycle stress detects orphaned processes, stale endpoint state, unsafe retries, and false healthy/fixed diagnostics"* criterion.

## Scoped narrow on purpose

Two neighbouring criteria are already met, and padding them would be noise rather than coverage:

- **sleep/wake** is pinned by `daemon-connection-supervisor.test.ts` — 501 lines, 14 tests, including *hiding clears timers and aborts active work without new publication or backoff* and *foregrounding a started supervisor triggers exactly one immediate fresh request*, plus the full backoff ladder and jitter bounds.
- **single-transition endpoint faults** are pinned by `daemon-endpoint-faults.test.ts` from increment 2 (#4500).

## What was actually missing

The **repeated** case. A single bind/close cycle can pass while the tenth fails, and that is where the real failures live:

- an address that reports `occupied` after its owner left
- a listener that cannot rebind because something held the path
- a probe that leaks one connection per cycle and slowly starves the daemon of accept slots

Each of those reaches a user as *"the daemon won't start"* or as a duplicate daemon — and none reproduces on the first try, which is exactly why single-shot tests miss them.

## The four tests

| test | what it catches |
|---|---|
| 12 bind/release cycles | a transition that drifts — stale `occupied` after release, or a cycle that stops reading `free` |
| rebind after churn | `EADDRINUSE` on the next launch because a released path was left behind |
| concurrent probes during churn | a launcher, status poll and recovery banner disagreeing — one stale `occupied` wrongly refuses a launch |
| per-cycle orphan check | a probe that holds its connection open, accumulating one per cycle |

Addresses are platform-appropriate — a named pipe on Windows, a temp socket elsewhere — so the same cases run on both legs.

## Verification

4 tests in 140ms; `check:tests-wired` (1607 files), `pnpm typecheck`, and `pnpm lint` all exit 0.

## Remaining on the parent bead

Real-daemon (process-level) startup/shutdown stress, which needs spawning the actual binary rather than a socket stand-in. Per-PR macOS coverage is `cave-5tpxy`.

Built in the existing `cave-58eoq.4.1` worktree rather than creating another — the checkout is at 31 against a 28 budget with writer-intent contention, and adding to that to avoid a `git checkout -b` would have been the wrong trade.

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -1281,6 +1281,7 @@ export const SUITES = {
     "src/lib/daemon-start.test.ts",
     "src/lib/daemon-socket-occupancy.test.ts",
     "src/lib/daemon-endpoint-faults.test.ts",
+    "src/lib/daemon-endpoint-churn.test.ts",
     "src/lib/daemon-startup-contract.test.ts",
     "src/lib/runtime-startup-throttle.test.ts",
     "src/lib/daemon-update-lifecycle.test.ts",
--- a/src/lib/daemon-endpoint-churn.test.ts
+++ b/src/lib/daemon-endpoint-churn.test.ts
@@ -1,0 +1,177 @@
+import assert from "node:assert/strict";
+import { mkdtempSync, rmSync } from "node:fs";
+import { createServer, type Server } from "node:net";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import test from "node:test";
+
+import { inspectDaemonAddress } from "./daemon-socket-occupancy.ts";
+
+/**
+ * Repeated startup/shutdown churn against a real endpoint (cave-58eoq.4).
+ *
+ * The bead asks for repeated lifecycle stress that "detects orphaned processes,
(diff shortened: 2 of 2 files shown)


BunsDev:
Fixed in `01b883f461`, and the review was right.

`bind()` created a real server without registering it in the file-level `cleanups` list. These tests assert **mid-cycle**, between `bind()` and `unbind()`, so an assertion that throws skips its unbind and leaves a listener holding the address — one genuine failure would cascade into a hung suite. That is exactly the behaviour the fault harness in #4489 exists to prevent, and increment 2's `listenOn()` already registers correctly; this file did not.

One addition beyond the reported line: the orphan-check test creates its own server inline (it counts connections, so it cannot use the helper) and had the identical flaw. Both are fixed.

PR #4512, opened 2026-08-10
Accept current Coven daemon health shape

pasmud:
## Summary
- Accept the current named daemon API version, `coven.daemon.v1`, during Cave startup compatibility checks.
- Treat daemon health `covenVersion: "0.0.0"` as a placeholder only after Cave has verified the installed Coven CLI version.
- Add regression coverage for both compatibility cases.

## Root Cause
Cave rejected a healthy local daemon because the daemon now reports API version `coven.daemon.v1`, while Cave accepted only `1` and `v1`. The same Windows v0.2.5 health shape can report `covenVersion: "0.0.0"`, which the status display already treats as a placeholder.

## Validation
- `node --experimental-strip-types src\lib\daemon-startup-contract.test.ts`
- `node --experimental-strip-types src\lib\daemon-start.test.ts`
- `pnpm test:daemon-faults`
- `pnpm check:tests-wired`
- `pnpm typecheck`

### DIFF
--- a/src/lib/daemon-startup-contract.test.ts
+++ b/src/lib/daemon-startup-contract.test.ts
@@ -15,6 +15,18 @@ assert.deepEqual(assessDaemonStartupCompatibility(healthy, "1.2.3"), {
   apiVersion: "v1",
 });
 
+assert.deepEqual(assessDaemonStartupCompatibility({ ...healthy, apiVersion: "coven.daemon.v1" }, "1.2.3"), {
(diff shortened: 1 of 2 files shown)


pasmud:
Closing this draft because current main now contains the daemon startup contract fix in a cleaner form. It accepts coven.daemon.v1 directly and no longer rejects the daemon runtime just because it does not match the CLI package version. Thanks.

PR #4519, opened 2026-08-10
feat(worktrees): let --apply run on the local plane behind an explicit flag

BunsDev:
`cave-s03wp` — option 2 from `cave-3aqvr`. Adds `--allow-unenforced-planes`.

## Why now

`--apply` refuses unless every maintenance plane reports `enforced`. Beads and GitHub are hard-coded off pending `cave-wqa0b.3`/`.4`, both **BLOCKED with no movement**, so automated retirement has been unreachable for as long as those have been open.

The cost compounds, because **creation is automated and retirement is not**:

- *2026-08-08* — four units hand-retired, and the checkout still went 19 → 22 in the same session.
- *2026-08-10* — **42 worktrees against a budget of 28**, 20 active exceptions, and every sweep that day reported **zero retirable units**. The two that had landed were held by cooldown and by this gate.

## The exception is narrow, and that is the design

| rule | why |
|---|---|
| **`local` is never waivable** | it performs the exclusion that stops two actors retiring the same unit — waiving it is an *unguarded* run, not a degraded one |
| **only known-pending planes** | held as an explicit set, so a plane added later blocks by default until someone decides otherwise |
| **never implicit** | without the flag the answer is exactly what it was; the first test asserts this, since it is the regression that would matter most |

`local` is checked **before** the known-set test, so an operator whose local plane is missing gets that specific refusal rather than a vaguer one.

## Structure

The decision is pure — `src/lib/maintenance-plane-admission.ts` — so the safety-critical part is exhaustively testable without a repository. The patrol keeps the boundary: it renders the refusal, or **announces the degraded run on stderr before touching any unit**, naming each waived plane and its blocking Bead. Auditing before acting matters because a run that dies mid-retirement must still have left the record of what was waived.

## Verified end to end against the live checkout

**Without the flag** — unchanged:
```
--apply unavailable; missing maintenance planes: beads, github
```

**With the flag** — and this is the more useful result:
```
DEGRADED APPLY — proceeding with unenforced maintenance planes
  waived: beads, github
    beads   cave-wqa0b.3
    github  cave-wqa0b.4
  local plane is enforced; exclusion still applies.
  Authorized by --allow-unenforced-planes (cave-s03wp).

failed to acquire maintenance gate: local-acquire-failed: gate-held
```

It passed the plane check for the first time, printed the audit, **and was then refused by the local plane** because another actor held the gate. The waivable planes were waived, the non-waivable one still stopped the run, and nothing was retired — the hard requirement demonstrated in practice rather than only in unit tests. With this many concurrent sessions, that gate is actively preventing double-retirement right now.

## Also

Corrects stale help text claiming `--apply` was unusable and that the coven plane was pending — coven is enforced now; only beads and github are off.

## Verification

9 unit tests; `check:tests-wired` (1612 files), `pnpm typecheck`, `pnpm lint` all exit 0.

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -1286,6 +1286,7 @@ export const SUITES = {
     "src/lib/daemon-socket-occupancy.test.ts",
     "src/lib/daemon-endpoint-faults.test.ts",
     "src/lib/daemon-endpoint-churn.test.ts",
+    "src/lib/maintenance-plane-admission.test.ts",
     "src/lib/daemon-startup-contract.test.ts",
     "src/lib/runtime-startup-throttle.test.ts",
     "src/lib/daemon-update-lifecycle.test.ts",
--- a/scripts/worktree-lifecycle-patrol.ts
+++ b/scripts/worktree-lifecycle-patrol.ts
@@ -12,6 +12,10 @@ import {
   repositoryMaintenanceCapabilities,
   verifyMaintenanceGateOwnership,
 } from "./maintenance-gate.mjs";
+import {
+  assessMaintenancePlaneAdmission,
+  type MaintenancePlaneCapabilities,
+} from "../src/lib/maintenance-plane-admission.ts";
 import { collectWorktreeLifecycleInventory } from "./worktree-lifecycle-inventory.ts";
 import {
   createGitRetirementOperations,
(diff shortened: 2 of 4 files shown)


BunsDev:
Fixed in `051cc15cf3`. All four points were correct, and the first was a real safety defect.

**Fail-open on missing data.** The filter used `capabilities[plane]?.enforced === false`, which is `false` when the entry is `undefined` — so an absent plane read as **enforced**. In the code that decides whether to waive safety planes, missing data must never mean satisfied. It is now `!== true`. An absent `local` plane would otherwise have waived the exclusion silently, which is the one direction this must never fail.

Two tests pin it: an absent `local` entry is refused **both with and without** the opt-in, and a malformed entry (`null`, `{}`, `enforced: "yes"`, `enforced: 1`) counts as unenforced rather than being coerced truthy.

**The coven plane is not hard-coded off.** It is opportunistic — enforced when `@opencoven/cli` maintenance is available, unenforced when it is not (`cave-wqa0b.2`) — so it can be either on a given run. I had written the wrong thing in three places: the module header, the CLI help text, and a test comment. All corrected. That is also precisely why coven belongs in the waivable set rather than being assumed present; only `beads` and `github` are genuinely hard-coded off.

11 tests, typecheck and lint clean.

PR #4529, opened 2026-08-10
chore(worktrees): version the scheduled sweep script and its prompt

BunsDev:
`cave-6qzo0`.

## Why

The scheduled worktree sweep ran from a script and prompt that existed on exactly one machine and in no history. That became more consequential on 2026-08-10, when the sweep switched from hand-retirement to `beads:worktrees:apply --allow-unenforced-planes` (`cave-s03wp`): **nothing in the repository recorded that the scheduled sweeps had changed behaviour**, and the prompt carries the discipline that makes that change safe —

- the `--max-retire 3` bound on an unattended run,
- the rule that `local-acquire-failed: gate-held` is *correct behaviour*, not a reason to fall back to hand-retirement,
- and why the apply path is used at all: hand-retirement is a multi-step ritual done from memory, and on 2026-08-10 two of five hand-retired units were removed with **no archive tag**, while every one left behind the bead metadata record that then blocks that bead from holding another worktree.

Losing that file loses the reasoning, not just the shell.

## Portable, not a copy of one machine's files

- repository root derived from the script's own location — no absolute user path in either file
- `SWEEP_LOG` and `SWEEP_REPO` overridable
- lock keyed by checkout, so two clones cannot block each other
- notification is macOS-only and best-effort (`command -v osascript`), so the script runs anywhere

The scheduler entry and, on macOS, the app bundle holding the Full Disk Access grant stay **out** of the repository — they are machine-specific — but the script header and prompt now state what that wiring must provide, so it is reconstructable.

## Behaviours preserved deliberately

Both were bugs found the hard way and are worth not re-introducing:

- **log rotation happens before the redirect.** Rotating afterwards is a no-op that looks like it worked — the shell's open fd follows the moved inode, so output lands in the rotated file and the "current" log stays empty.
- **the lock records its owner pid and stale locks are reclaimed loudly.** A bare `mkdir` lock is a silent-failure trap: a SIGKILLed run leaves it behind and every later run exits 0 saying "another sweep holds the lock" — no error, no sweep, forever.
- **the filesystem preflight aborts loudly.** Without it, a permissions gap lets `cd` succeed while every git call fails, and the agent runs against an empty view of the repository — which looks exactly like "nothing to retire".

## Verification

Ran the committed copy: it derives the repo root, passes the preflight, redirects to the overridden log, and reaches the agent. `check:tests-wired` (1614 files) and `pnpm lint` both exit 0.

One property is documented rather than fixed: deriving the root from the script's location means invoking a worktree's copy sweeps the **whole repository**, since `git worktree list` there enumerates every unit. The scheduled job is pinned to the primary checkout, which is what is wanted, but running it by hand from a worktree is not scoped — noted in the script.

### DIFF
--- a/scripts/worktree-sweep-prompt.md
+++ b/scripts/worktree-sweep-prompt.md
@@ -1,0 +1,92 @@
+Scheduled worktree sweep for the coven-cave repository. You are running
+unattended from a scheduler, with the primary checkout as your working
+directory. Other interactive sessions may be live on this same checkout right
+now. Assume nothing; verify everything.
+
+The wiring that launches you is machine-specific and lives outside the repo: a
+scheduler entry (launchd on macOS) invoking scripts/worktree-sweep.sh, and on
+macOS an app bundle holding the Full Disk Access grant that a scheduled job
+needs to read a protected folder. If that grant is missing the sweep aborts
+loudly rather than reporting an empty repository as "nothing to retire".
+
+Your job: run the lifecycle patrol and retire ONLY what has genuinely become
+cleanup-ready. Preserving a unit that should have been retired costs nothing.
+Destroying a unit that was still live is unrecoverable. When those two trade off,
+always preserve.
+
+## 1. Assess
+
+- Check GraphQL quota first: `gh api rate_limit --jq .resources.graphql`. If it is
+  nearly exhausted, STOP and report — the patrol will exit 1 with an incomplete
+  inventory, and an incomplete inventory must never be treated as "nothing to do".
+- Run `pnpm beads:worktrees:json` (takes several minutes). Skip the pnpm banner
+  before the first `{` when parsing. Each item's `lane` field is the
+  classification; `counts` is the tally.
+
+## 2. Decide
+
+- Retire ONLY units in lane `cleanup-ready`.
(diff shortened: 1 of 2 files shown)


BunsDev:
Checked this rather than applying it, and it does not hold on the target platform.

```
$ /bin/date -Iseconds
2026-08-10T14:54:15-05:00      # exit 0

$ PATH="$HOME/.local/bin:/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin" command -v date
/bin/date                      # BSD date, not a Homebrew coreutils shim
```

Apple's BSD `date` gained `-I` some releases ago, so this is not GNU-only in practice. The evidence is also empirical: this exact script has been running under launchd all day and its log is full of well-formed timestamps —

```
[sweep] start 2026-08-09T19:45:28-05:00
[sweep] start 2026-08-10T09:00:05-05:00
```

Declining the change: swapping working code on a false premise is churn, and `date -Iseconds` is the more readable form.

The one case where the concern would be real is macOS 12 or older, whose `date` predates `-I`. If this script ever needs to run there, `date +%Y-%m-%dT%H:%M:%S%z` is the portable equivalent — noting it here so the next reader does not have to re-derive it.

PR #4532, opened 2026-08-10
fix(worktrees): make a deferred sweep distinguishable from a clean one

BunsDev:
Found on the **first run of the newly versioned sweep** (#4529), 2026-08-10T15:10.

## What happened

The agent checked GraphQL quota, found it nearly exhausted, and correctly declined to run the patrol on a thin pool — exactly what the prompt asks for. The run then logged:

```
[sweep] worktrees remaining: 38
[sweep] no change (38 -> 38); no notification sent
[sweep] end 2026-08-10T15:13:57-05:00
```

Exit 0. No notification. **Identical to a healthy sweep that assessed everything and found nothing retirable.**

The patrol never ran — the run's output contains no lane tally, no `cleanup-ready` count, and no `DEGRADED APPLY` banner. A reader had no way to tell the difference, which is the same silent-success shape as the TCC gap and the stale lock this script already guards against.

## Two changes

Prompt discipline alone would not survive a future agent phrasing its refusal differently, so both sides are covered:

**Prompt** — stopping at the quota check now requires the literal first line `SWEEP DEFERRED: quota`, and explains why: the wrapper measures worktrees before and after, and a deferral is invisible in that measure. It also tells the agent **not** to wait for the reset and **not** to promise a later report — the process ends when it stops writing, and the scheduler is the retry. The observed run said *"I'll report once the patrol returns"*, which it could not.

**Script** — reads the agent's own output back from a recorded byte offset and, on finding the marker, logs `DEFERRED` with the reason instead of `no change`.

## Verification

Exercised the branch both ways: a deferred report is flagged, and an ordinary "nothing was retired" report is **not** false-positived. `pnpm lint` exits 0.

The wrapper still counts worktrees itself for the retirement case — a retirement is a filesystem fact, not a claim — so this adds a signal rather than starting to trust the agent's summary.

### DIFF
--- a/scripts/worktree-sweep-prompt.md
+++ b/scripts/worktree-sweep-prompt.md
@@ -19,6 +19,17 @@ always preserve.
 - Check GraphQL quota first: `gh api rate_limit --jq .resources.graphql`. If it is
   nearly exhausted, STOP and report — the patrol will exit 1 with an incomplete
   inventory, and an incomplete inventory must never be treated as "nothing to do".
+
+  **If you stop here, say so unmistakably.** Begin your report with the literal
+  line `SWEEP DEFERRED: quota` and do not describe what a later run might find as
+  though you will still be here to run it — you will not; this process ends when
+  you finish writing. A deferred run is otherwise invisible: the wrapper counts
+  worktrees before and after, sees no change, logs `no change (N -> N)` and exits
+  0, which reads exactly like a healthy sweep that found nothing retirable. That
+  happened on 2026-08-10T15:10 and the run looked clean.
+
+  Do not wait for the reset either. Waiting burns the run, and the scheduler will
+  come back on its own; the next scheduled sweep is the retry.
 - Run `pnpm beads:worktrees:json` (takes several minutes). Skip the pnpm banner
   before the first `{` when parsing. Each item's `lane` field is the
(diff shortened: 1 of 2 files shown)


BunsDev:
Both fixed in `1708a62395`, and the second was the same class of bug this PR exists to close.

**Marker matched anywhere.** The check accepted any later line beginning with `SWEEP DEFERRED`, so a normal report that merely quoted the phrase would be misread as a deferral — and now that the marker is documented, a report explaining why it did *not* defer is a realistic way to trip it. Only the first non-empty line is considered now, matching what the prompt requires.

**A deferral could mask a real failure.** If the agent printed the marker and then exited non-zero, the deferral branch ran and the error notification was skipped — hiding a genuine failure behind the marker, which is precisely the silent-success shape this guard was added to prevent. Non-zero exit is now checked **first** and always notifies; a deferral alongside it is noted in the log rather than replacing the alert.

Verified across five cases:

| case | result |
|---|---|
| marker on first line | DEFERRED |
| marker after leading blank lines | DEFERRED |
| report quoting the phrase mid-body | no change (no false positive) |
| deferral **and** non-zero exit | ERROR notified, deferral noted |
| ordinary clean report | no change |

`pnpm lint` exits 0.

PR #4533, opened 2026-08-10
feat(security): require user-bound auth for access-token-only loopback

BunsDev:
Access-token-only deployments admitted verified direct-loopback REST and PTY
requests **with no credential**. TCP loopback does not distinguish OS users, so
any local account could reach privileged workspace APIs — or spawn and adopt a
PTY as the Cave process owner, which is a local privilege-escalation primitive
rather than merely an over-broad read.

Implements the design merged in #4483, with the owner's scoping applied:
**desktop shell only, browser bootstrap dropped.**

## What changes

- When access auth is armed, a direct loopback browser uses the same access
  gate as a remote one. The per-boot local-peer stamp still separates direct
  from forwarded traffic, but is no longer treated as *identity*.
- The Tauri shell bypasses only with its per-launch sidecar credential.
- A local valid access token passes final auth without being marked remote.
- PTY requires cryptographic auth even on loopback.
- Forged forwarding or tailnet identity no longer authenticates REST or PTY: a
  direct loopback client can set `x-forwarded-*` itself, so tailnet context is
  consumed only *after* bearer authentication, for passkey-presence binding.
- Explicit tokenless development is unchanged.

## Provenance

The implementation was written in an earlier session and left **uncommitted for
18.5 hours**. I verified rather than authored it, then carried it forward so a
P1 security fix stopped sitting in a working tree. Everything was independently
retained first (branch head archived, plus a retention tag on the WIP), so no
route through this risked the original.

## The merge is the interesting part

`main` had independently converged on the same fix, adding
`shouldRejectUnauthenticatedPtyUpgrade({...})` where this branch had the
equivalent inline `isPtyAuthRequired() && !tokenAuthenticated`.

I took main's named helper. Partly because it is better — but mainly because
`src/lib/pty-upgrade-auth.test.ts` asserts the helper's **source text**, so
restoring the inline form would have broken a test in a way that is easy to
miss.

Two things I kept that main's side would have dropped:

**The `#714` history.** This branch recorded that removing the credential-less
path once 401'd every local terminal, reproducing the v0.0.72 *"Terminal
connection failed"* regression `server-pty-ws.test.ts` still guards. Main's
rewrite is more concise but lost it, and that history is exactly why the path is
load-bearing rather than an oversight. Merged into main's wording.

**A negative assertion**, restored and adapted to the new helper. Main asserts
the helper is *called correctly* — which would still pass if someone re-added a
direct-loopback escape hatch beside it, the very bypass this removes. Verified
non-vacuous: it matches a re-introduced `isDirectLoopbackRequest` term and does
not match the current code.

## Verification

`tsc --noEmit` clean; `pnpm test:api` 368 files; and all five affected suites
pass — `proxy-behavior`, `middleware`, `tailnet-identity`, `server-pty-ws`, and
main's own `pty-upgrade-auth`.

Not covered here: both flips want a check in the native Tauri shell before
release, since headless checks did not catch #714.

Closes cave-ruw4z.


BunsDev:
**Not merging — CI found a real consequence of the design, not a flake.**

`Frontend build` fails on `tests/mobile/foundations.spec.ts:177`:

```
preference PATCH failed: {"ok":false,"error":"unauthorized"}
```

Failed identically on retry. This is the #714 shape — tightening loopback auth 401s a legitimate local client.

**Why it happens, and why the code is arguably correct.** `playwright.config.ts:185` arms `COVEN_CAVE_ACCESS_TOKEN` for the `preferences-*` projects. Under this change, an armed access token means a plain local browser must pass the access gate rather than ride the direct-loopback stamp. The spec's `request.patch("/api/preferences")` carries only `x-coven-cave-local-peer` (the config's shared `extraHTTPHeaders`), which is exactly what this PR demotes from identity to routing information. So the gate is doing what the merged design says.

**The bind is the scoping decision.** The design was scoped to *desktop shell only, browser bootstrap dropped* — deliberately, by the owner. The Tauri shell authenticates with its per-launch sidecar token. A plain local browser with an access token armed therefore has no bootstrap path left, and the e2e harness is precisely that: a plain local browser in a token-armed environment.

So one of these has to give, and it is a product decision rather than a test fix:

1. **The e2e harness authenticates like any access-gated client** — supply the fixture token through the documented mobile-access mechanism in the `preferences-*` projects. Narrowest change; asserts the new contract honestly.
2. **Restore a minimal local-browser path** — the bootstrap that was explicitly dropped. Reopens the scoping question rather than settling it.
3. **Narrow the rule** so an armed *access* token alone does not gate direct loopback, leaving the sidecar requirement for PTY. Weakens exactly the REST half this bead exists to close.

I did not pick one. Choosing (1) unilaterally means editing a test so a security change passes, and that is the move that hides regressions; choosing (3) silently discards half the bead.

**Also worth recording:** this work was verified with build, lint, typecheck, `test:api` and focused proxy/PTY suites — but never against Playwright. e2e is the only layer that exercises a real local browser against the gate, which is the same reason #714 escaped headless checks. The PR body already flags that both flips want a native Tauri shell check before release; this failure is the cheaper half of that warning arriving early.

Branch is pushed and retained; nothing is lost while this is decided.

PR #4534, opened 2026-08-10
fix(claims): canonicalise worktree edits instead of skipping them

BunsDev:
`cave-ahc91`.

## The hook was inert for exactly the sessions it exists to coordinate

`surface-claim-guard.mjs` returned early on any path containing `.worktrees`, reasoning that those sessions are already isolated and diverge visibly at PR/merge time. But `CLAUDE.md` mandates **a worktree per concurrent session** — so that skip made every session the hook exists to coordinate invisible to it.

| measured | sessions | worktrees | claims.json |
|---|---|---|---|
| 2026-08-09 | 5 live | 14 | 1 session, 3 days stale |
| 2026-08-10 | — | **38** | still 1 session |

`CLAUDE.md` advertises this hook as operationalising §1 of the coordination doc so "you no longer have to grep claims.json by hand". It was doing nothing.

**PR/merge is not the fallback it was assumed to be.** It catches *conflicting* edits. It never surfaces *duplicated* work — two sessions independently solving the same symptom conflict with nothing, and that is the costlier failure the coordination doc is actually about.

## The fix

A file is **one surface** regardless of which worktree it is edited from. The `.worktrees/<unit>/` prefix is stripped, and the claim records which worktree it came from so the warning can name it.

**A second defect, found while reading:** the ledger path was derived from the project root, so a session whose own root is a worktree would write `.claude/claims.json` *inside that worktree*. Every session would have had a private ledger and seen no one else's — the same inertness by a different route. It now always resolves to the primary checkout.

The collision message no longer advises "isolate in a worktree", which is incoherent when both sessions already are.

## Tests

Test 5 asserted the old skip; it now asserts canonicalisation. Three cases added:

- the same file in two different worktrees collides, and the warning names the other worktree
- a worktree edit collides with a primary-checkout edit of the same file
- a worktree-rooted session writes to the **primary** ledger, and no ledger appears inside the worktree

The never-block invariant is re-verified — garbage input and a missing file path both still exit 0. `check:tests-wired` (1614) and `pnpm lint` clean.

## Scope

This is option (a) from the bead. Option (b) — a pre-flight `bd`/`ps` check before claiming a bead — targets duplicated *effort* rather than concurrent edits and remains unaddressed; the bead records it.

### DIFF
--- a/scripts/surface-claim-guard.mjs
+++ b/scripts/surface-claim-guard.mjs
@@ -17,9 +17,15 @@
  *     clobbering the first session's work.
  *
  * It is ADVISORY ONLY. It never blocks or fails a tool: on any error, or for any
- * path it doesn't care about, it exits 0 silently. Edits inside `.worktrees/`
- * are skipped — those sessions are already isolated, and their divergence
- * surfaces at PR/merge time (per the doc), which is the whole point of worktrees.
+ * path it doesn't care about, it exits 0 silently.
+ *
+ * Edits inside `.worktrees/` are CANONICALISED, not skipped (cave-ahc91). They
+ * used to be skipped on the reasoning that worktrees are isolated and diverge
+ * visibly at PR/merge time — but the convention mandates a worktree per
+ * concurrent session, so that made every session this hook exists to coordinate
+ * invisible to it, and PR/merge only catches conflicting edits, never duplicated
+ * work. A file is one surface no matter which worktree it is edited from, and
+ * the ledger lives in the primary checkout so every session shares one view.
  *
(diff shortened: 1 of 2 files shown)


BunsDev:
Fixed in `5d03e4332d`, and the point was right.

Both collision tests I added matched regexes against `res.stdout`, which is the JSON envelope — so they were asserting against escaping as much as against the message, and were inconsistent with test #2, which already parses. Both now `JSON.parse` and assert on `systemMessage`.

One thing tightened beyond the report: each now also asserts the **canonical surface** appears (`src/shared.ts`) rather than a worktree-qualified path. That is the actual claim under test — that two worktrees collapse to one surface — and it was only implied before.

PR #4541, opened 2026-08-11
fix(security): scope research grants to individual mission workspaces

BunsDev:
### Motivation

- A recent change made the shared `researchMissionsRoot()` a single registered Cave project and auto-granted familiars write access, which broadened per-mission grants to the entire research tree and broke mission isolation.  
- The intent of this change is to restore isolation by ensuring grants are created only for the active mission's workspace so familiars cannot access sibling missions' files.

### Description

- Restrict landing grants to a mission-scoped project by changing `ensureResearchLandingAccess` to register/grant the specific mission workspace via `researchMissionWorkspacePath(missionId)` instead of the shared root.  
- Remove the unsafe behavior that folded/widened legacy ad-hoc mission grants into a global landing project and instead remove any insecure shared-root project and its grants if present.  
- Wire mission identity through run-start preflight by passing `mission.id` into `ensureResearchAccess` and `missionStartTarget` so the runner ensures a mission-scoped grant only for the active workspace.  
- Add and update regression tests: `src/lib/server/research-landing.test.ts` to verify mission-scoped grants, legacy shared-root cleanup, and non-inheritance between familiars; and `src/lib/server/research-mission-runner-lifecycle-actions.test.ts` to assert that the mission id is passed to the access preflight.

### Testing

- Ran `node --experimental-strip-types --test src/lib/server/research-landing.test.ts` and the three isolation tests passed.  
- Ran `node --experimental-strip-types --test src/lib/server/research-mission-runner-lifecycle-actions.test.ts` and the updated runner preflight test passed.  
- Ran `pnpm typecheck` (TypeScript) and `pnpm check:tests-wired`; both completed (the typecheck run printed a Node engine warning for the environment but completed).  
- Ran the full API test suite via `pnpm test:api`, which exited due to an unrelated `scripts/dev-app-teardown.test.mjs` assertion failure; the targeted research tests pass independently and the change is limited to research grant scoping.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a784ed3a5488327a5d2ed19bf79abd2)

### DIFF
--- a/src/lib/server/research-landing.test.ts
+++ b/src/lib/server/research-landing.test.ts
@@ -1,5 +1,5 @@
 import assert from "node:assert/strict";
-import { mkdtemp, rm, stat } from "node:fs/promises";
+import { mkdtemp, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import path from "node:path";
 import test from "node:test";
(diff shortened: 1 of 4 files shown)


BunsDev:
Closing: `main` has since gone the other way on purpose, and this branch conflicts.

`git merge-tree origin/main HEAD` conflicts in `src/lib/server/research-mission-runner.ts` (83 commits behind).

More decisively, the design this PR builds on is gone. `main` now has a single **research landing project** plus `migrateAdHocMissionGrants`, which actively *folds* per-mission grants back into it and deletes the per-mission project records. Its own comment names the pattern this PR would reintroduce:

> Before the landing project existed, reaching a mission's output from chat required registering the individual mission directory as its own project and granting THAT — leaving one narrow, easily-forgotten grant per mission. Every registered project strictly inside the landing root is such a relic […] the per-mission project and all of its grants are removed.

So scoping grants to individual mission workspaces is no longer a hardening step — the migration would immediately undo it, and the two together would churn grants every call.

@copilot-pull-request-reviewer's point (security-critical cleanup should run before the grant work, so a throw mid-way doesn't strand the overbroad project) is worth keeping in mind for `ensureResearchLandingAccess`, but it does not apply as written: on `main` the migration already runs *first*, at line 65, before permissions are loaded or any grant is issued.

Reopen with a rebase if the per-mission scoping is still wanted as a direction — that's a design call, not a merge conflict.

PR #4544, opened 2026-08-11
fix(chat): prevent client origin from bypassing project authorization

BunsDev:
### Motivation

- Close an authorization bypass where an untrusted `body.origin` allowed fresh requests to enter the legacy projectless hidden-generation lane and skip project registration and familiar-access checks. 
- Ensure project gating remains server-authoritative while preserving existing persisted hidden-generation resume behavior.

### Description

- Change the projectless-generation decision in `src/app/api/chat/send/route.ts` to require persisted conversation provenance by checking `Boolean(existingConversation)` and using `existingConversation.origin` instead of trusting `body.origin` for authorization. 
- Keep `body.origin` available as conversation metadata but remove its use as an authorization input so fresh requests cannot spoof hidden-generation origins. 
- Add regression assertions in `src/app/api/chat/send/harness-routing-host-session.test.ts` to ensure the route source no longer matches patterns that would let `body.origin` influence authorization and to assert that only persisted provenance may bypass project authorization.

### Testing

- Ran `git diff --check` to validate there are no whitespace or diff errors and it passed. 
- Ran unit tests with `node --experimental-strip-types --test src/app/api/chat/send/harness-routing-host-session.test.ts src/lib/server/chat-project-launch.test.ts` and all tests passed. 
- Ran type checking with `pnpm typecheck` (`tsc --noEmit`) and it succeeded.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a6648e8b22883278d79b05d68ef7bfc)

### DIFF
--- a/src/app/api/chat/send/harness-routing-host-session.test.ts
+++ b/src/app/api/chat/send/harness-routing-host-session.test.ts
@@ -485,6 +485,11 @@ assert.doesNotMatch(
   /isProjectlessGenerationOrigin\([^)]*body\.origin/,
   "client-supplied origin must never decide whether project authorization is required",
 );
+assert.match(
+  chatRoute,
+  /const persistedOrigin = persistedConversationOrigin\(body\.origin\)/,
+  "projectless origins from request JSON should be stripped before transcript persistence",
+);
 
 const authorizeLaunchIndex = chatRoute.indexOf("await authorizeChatProjectLaunch");
 const offlineLaunchIndex = chatRoute.indexOf("const offlineChatResponse = await maybeQueueOfflineChat");
--- a/src/app/api/chat/send/harness-routing-host-session.test.ts
+++ b/src/app/api/chat/send/harness-routing-host-session.test.ts
@@ -477,8 +477,13 @@ assert.doesNotMatch(
 );
 assert.match(
   chatRoute,
-  /const generationOrigin = existingConversation\?\.origin \?\? body\.origin/,
-  "a request must not relabel an existing user chat as a projectless hidden generation",
+  /Boolean\(existingConversation\)[\s\S]*isProjectlessGenerationOrigin\(existingConversation\?\.origin\)/,
+  "only persisted hidden-generation provenance may bypass project authorization",
+);
+assert.doesNotMatch(
+  chatRoute,
+  /isProjectlessGenerationOrigin\([^)]*body\.origin/,
+  "client-supplied origin must never decide whether project authorization is required",
 );
 
 const authorizeLaunchIndex = chatRoute.indexOf("await authorizeChatProjectLaunch");
--- a/src/app/api/chat/send/route.ts
+++ b/src/app/api/chat/send/route.ts
@@ -1103,11 +1103,13 @@ export async function POST(req: Request) {
   const resolvedFamiliarWorkspace = !sshRuntime
     ? await resolveFamiliarWorkspace(body.familiarId)
     : undefined;
-  // A persisted conversation owns its provenance. Never let a request relabel
-  // an existing user chat as a hidden generator to bypass project checks.
-  const generationOrigin = existingConversation?.origin ?? body.origin;
+  // Only persisted provenance may select the legacy projectless generation
+  // lane. `body.origin` is untrusted request metadata: it may label a newly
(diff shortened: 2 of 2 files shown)


BunsDev:
Queue triage — I traced the open review thread against the head, and **the finding holds**, though the blast radius is narrower than the comment implies. Both parts matter for deciding the fix.

**The sequence is reachable.** `origin` is persisted from unvalidated request JSON (`route.ts:1911`, `:3427` — `...(body.origin ? { origin: body.origin } : {})`), and `projectlessGeneration` (`:1109`) reads it back off `existingConversation`. So:

1. Turn 1 — send `projectRoot: <a project you legitimately have>` plus `origin: "canvas"`. `existingConversation` is undefined, so `projectlessGeneration` is `false` and `authorizeChatProjectLaunch` runs normally and passes. The stub persists `origin: "canvas"`.
2. Turn 2 — same `sessionId`, omit `projectRoot`. Now `existingConversation.origin === "canvas"` makes `projectlessGeneration` true, and the `authorizeChatProjectLaunch` call is skipped entirely.

The in-code comment says "`body.origin` is untrusted request metadata: it may label a newly created conversation, but it must never bypass project authorization" — which is exactly the intent, but labelling a new conversation is precisely what makes the later bypass reachable, because that label is the persisted provenance the next turn trusts.

**The bound.** On the bypass path `authorizedProjectRoot` is `resumeCwd ?? resolvedFamiliarWorkspace` — the daemon's own session cwd or the familiar workspace, neither of which the client picks. So this is **not** arbitrary-path escalation. What it does drop is *re-authorization*: a caller who passed the gate once keeps launching against that session after access should have stopped applying (project deregistered, access revoked, root moved). Real, but a persistence bug rather than a path-traversal one — worth pricing the fix accordingly.

**Suggested direction:** make the projectless lane depend on provenance the server owns, not on a field the client supplied. Either stamp `origin` server-side when the conversation is created (deriving it from the authenticated surface rather than `body.origin`), or record the authorized root on the conversation at turn 1 and re-validate that stored root on later turns instead of skipping the check.

Not merging this one on the sweep — the thread is load-bearing and the fix is a design call rather than a typo. Everything else is green.

BunsDev:
Closing — the branch cannot merge (**1546 commits behind `main`**, `git merge-tree` conflicts), and as written the fix would take the hidden-generation lane down with it. The underlying weakness is real and still present on `main`; it needs a different fix.

## The weakness is confirmed

`main:2082` still reads `const generationOrigin = existingConversation?.origin ?? body.origin`, and `existingConversation.origin` is persisted straight from `body.origin` (five sites). So turn 1 sends a legitimate `projectRoot` plus `origin: "canvas"`, auth passes, the stub persists that origin; turn 2 omits `projectRoot`, `projectlessGeneration` is true, and `authorizeChatProjectLaunch` is skipped.

**Bounding it honestly:** on the bypass path `authorizedProjectRoot` is `resumeCwd ?? resolvedFamiliarWorkspace` — daemon-owned, not client-chosen. This is a *loss of re-authorization* (revoked access keeps working on an existing conversation), not arbitrary-path escalation.

## Why this fix can't land as-is

The two halves close the loop on each other:

- `persistedConversationOrigin` strips every projectless origin before persistence, so no conversation can ever hold one.
- `projectlessGeneration` then requires `Boolean(existingConversation) && isProjectlessGenerationOrigin(existingConversation?.origin)`.

Nothing else writes those origins server-side, so the predicate is unsatisfiable and the lane is dead. Worse, turn 1 of a hidden generation has no `existingConversation` at all, so a projectless `canvas` / `enhance` / `journal` call now falls into `authorizeChatProjectLaunch` with no `projectRoot` and fails `project_root_required` (400). Canvas, Enhance and Journal all break on their first call.

## What would work

Either direction, both server-side:

1. **Stamp `origin` server-side** from the authenticated surface that issued the request, so `body.origin` is a hint rather than provenance — then trusting the persisted value is sound.
2. **Persist the authorized project root** on the conversation and re-validate the familiar's grant on it every turn, so the projectless lane never means "skip the check", only "the root is server-owned".

That's a design call rather than a merge fix, so I'm leaving it to @BunsDev rather than picking one. Happy to implement whichever.

PR #4545, opened 2026-08-11
fix(daemon): avoid leaking hub token during probes

BunsDev:
### Motivation
- The `/api/daemon/probe` helper accepted an attacker-controlled URL and could reuse the globally stored Server Hub access token when the candidate URL did not include its own token, enabling SSRF + credential exfiltration.
- Probing arbitrary candidate URLs must not send the stored hub credential unless the candidate explicitly supplies a token.

### Description
- Change `probeDaemonUrl` to avoid sending the stored hub `accessToken` for candidate probes by default, only preserving an access token when the candidate URL explicitly includes `coven_access_token` (via a new `explicitHubAccessToken` helper and use of `normalizeHubUrl`).
- Call the probe with a sanitized `probeTarget` that omits `accessToken` unless the candidate carried an explicit token. (`src/lib/server/daemon-probe.ts`)
- Add regression tests that assert stored credentials are omitted for candidate probes and that an explicitly supplied candidate token is preserved. (`src/app/api/daemon/probe/route.test.ts`)

### Testing
- Ran `pnpm typecheck` and it completed successfully. 
- Ran `node --test src/app/api/daemon/probe/route.test.ts` and the new and existing probe tests passed. 
- Ran `pnpm test:api`: many API tests ran and the new probe tests passed, but the full `test:api` run aborted after 209 tests due to an external environment dependency (`zstd`) required by `scripts/sidecar-archive-manifest.test.mjs`, unrelated to the probe change.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a6032543f8483278ae136357fcf2d37)

### DIFF
--- a/src/app/api/daemon/probe/route.test.ts
+++ b/src/app/api/daemon/probe/route.test.ts
@@ -15,6 +15,41 @@ test("probe helper accepts a stubbed daemon caller for route-level outcomes", as
   assert.equal(result.reason, "hub unhealthy: maintenance");
 });
 
+test("probe helper omits stored hub credentials for candidate URLs", async () => {
+  const previous = process.env.COVEN_CAVE_HUB_ACCESS_TOKEN;
+  process.env.COVEN_CAVE_HUB_ACCESS_TOKEN = "stored-secret";
+  try {
+    let authorization: string | undefined;
+    await probeDaemonUrl("http://candidate.tailnet:8787", async (target) => {
+      authorization = target.mode === "hub" ? target.accessToken : undefined;
+      return { ok: true, status: 200, data: { ok: true } };
+    });
+    assert.equal(authorization, undefined);
+  } finally {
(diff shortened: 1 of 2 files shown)


BunsDev:
Already landed on `main` — closing as obsolete.

This branch is **1885 commits behind `main`** and `git merge-tree origin/main HEAD` reports a conflict. In the meantime `main` grew a strictly stronger version of exactly this fix in `src/lib/server/daemon-probe.ts`:

```ts
const { accessToken: _storedCredential, ...uncredentialedTarget } = resolvedTarget;
const target: DaemonTarget = embeddedToken
  ? { ...uncredentialedTarget, accessToken: embeddedToken }
  : uncredentialedTarget;
```

@copilot-pull-request-reviewer's point — that the key should be *removed* rather than set to `undefined`, so a future `'accessToken' in target` can't read as present — is satisfied by that destructuring. `main` additionally refuses to send any token over a non-loopback plaintext transport.

No action needed; nothing from this branch is missing from `main`.

PR #4548, opened 2026-08-11
fix(security): validate resolved worktree cwd

BunsDev:
### Motivation
- Close a security gap where a requested `.worktrees/<branch>` path was authorized lexically but the actual runtime cwd could resolve (via symlink) outside the parent project, allowing a worktree grant to be used for an unrelated directory.
- Ensure worktree-based authorization only maps to the parent project when both the requested path and the runtime-resolved cwd are actually inside the parent project's `.worktrees/` containment.

### Description
- Change `worktreeParentProject` to accept the runtime `resolvedCwd` and require that both the requested path and the resolved cwd are strictly below the parent project's `.worktrees/` directory before returning the parent project.  
- Update `chatProjectAccessId` to pass `args.resolvedCwd` into `worktreeParentProject`.  
- Tighten the function comment to document the runtime-cwd requirement to prevent symlink escapes.  
- Add a regression test in `src/lib/chat-project-access.test.ts` that asserts a requested `.worktrees/evil` path whose `resolvedCwd` points to an unregistered victim directory fails closed as `unregistered:<requested root>`.

### Testing
- Ran the unit regression with `node --experimental-strip-types src/lib/chat-project-access.test.ts`, which passed.  
- Ran the TypeScript checker with `pnpm typecheck` (`tsc --noEmit`), which completed with no errors.  
- Ran `git diff --check` to validate whitespace/patch issues, which reported no problems.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a62a118a58c83278f659c71f96d1432)

### DIFF
--- a/src/lib/chat-project-access.test.ts
+++ b/src/lib/chat-project-access.test.ts
@@ -165,6 +165,17 @@ assert.equal(
   "a traversal escape below .worktrees/ resolves outside and fails closed",
 );
 
+
+assert.equal(
+  chatProjectAccessId({
+    projects,
+    requestedProjectRoot: "/Users/me/dev/cave/.worktrees/evil",
+    resolvedCwd: "/Users/me/victim-ungranted-project",
+  }),
+  "unregistered:/Users/me/dev/cave/.worktrees/evil",
+  "a symlinked worktree request whose real cwd escapes the parent project fails closed",
+);
+
 assert.equal(
   chatProjectAccessId({
     projects,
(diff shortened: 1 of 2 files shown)


BunsDev:
Superseded by #4550.

This branch is **1768 commits behind `main`** and `git merge-tree origin/main HEAD` reports a conflict in `src/lib/chat-project-access.ts`: the file was substantially reworked since the branch point (`taskWorktreeProjectAccessId` was added, and the familiar-workspace exemption this branch still carries was removed as legacy). It cannot land as-is.

The security fix itself is correct and still needed — `main` today authorizes the `.worktrees/` carve-out on the client-supplied root alone — so I re-landed it on current `main` in #4550.

@copilot-pull-request-reviewer's comment on `chat-project-access.ts:44` was **right, and I confirmed it reachable**: I wrote a symlink test against this branch's head and it failed, returning `unregistered:…` for a legitimate worktree under a symlink-registered project. `resolvedCwd` is realpathed by `resolveLocalRuntimeCwd` while the prefix was only `path.resolve`d, so `/var -> /private/var` on macOS is enough to break it. #4550 canonicalizes both sides through `realpathOrResolve` and keeps a test proving containment is not widened.

PR #4549, opened 2026-08-11
test(e2e): pin Chat-to-Coding workflow baseline

BunsDev:
## Summary
- add a deterministic daemon-less Chat → Changes → Coding Desk → Chat baseline
- prove exact session, project, worktree, file, review, terminal, and draft continuity
- capture asserted dark/light/alternate-theme, constrained, tablet, narrow, rounded-shell, and reduced-motion contracts
- fail closed on unexpected API traffic and freeze visual time deterministically

## Verification
- `COVEN_CAVE_PORT=3100 pnpm exec playwright test tests/chat-code-workflow.spec.ts`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm check:tests-wired`

Bead: `cave-ltl38.1`

### DIFF
--- a/tests/chat-code-workflow.spec.ts
+++ b/tests/chat-code-workflow.spec.ts
@@ -1,0 +1,897 @@
+import { expect, test, type Page, type TestInfo } from "@playwright/test";
+import { writeFile } from "node:fs/promises";
+import { buildChatUsagePlanSnapshot } from "../src/lib/chat-usage-plan";
+import { createDefaultPreferences } from "../src/lib/preferences-schema";
+import { catalogForRuntime, runtimeModelInventoryScope } from "../src/lib/runtime-models";
+
+const ISO = "2026-08-11T05:00:00.000Z";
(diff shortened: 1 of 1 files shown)


BunsDev:
Fixed in df43312 — good catch, and exactly the right objection for a PR whose whole job is to pin a contract.

`project-tree` now asserts:
- `root` present and inside `WORKTREE_ROOT` — **not** equal to it, because `fetchChildren` passes the directory being expanded, so it is the tree root on first load and a child on every expansion
- `depth` parses to a positive number
- `familiarId` **present**, not truthy — `fetchChildren(dirPath, familiarId = "")` sends the key with an empty value when no familiar is scoped, so a truthiness assert would fail on a legitimate request

`project-file` gets the same treatment for `path` (it does not read `root`, per `src/app/api/project-file/route.ts:163`).

The fulfilled entry now echoes the requested `root` rather than hardcoding `WORKTREE_ROOT`, so an expansion returns children of the directory actually asked for.

PR #4560, opened 2026-08-12
chore(release): stamp v0.3.1

BunsDev:
## Summary

- stamp Coven Cave v0.3.1 across desktop, Cargo, and iOS version sources
- advance the iOS build number for App Store Connect
- add finalized release notes for the access-token gate removal and its security boundary

## Verification

- `pnpm release:verify --version 0.3.1`
- `node scripts/stamp-release.test.mjs`
- `node scripts/release-notes.test.mjs`
- `git diff --check`

Bead: `cave-36q9s`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,27 @@ breaking config changes; patch releases stay additive.
 
 ## [Unreleased]
 
+## [0.3.1] - 2026-08-12
+
+> Tokenless access across trusted local and tailnet surfaces, with optional
+> passkey presence and local-only boundaries preserved.
+
+Patch release on top of v0.3.0.
+
+### Changed
+
+- Removed the `COVEN_CAVE_ACCESS_TOKEN` request gate, access page, and
+  query-to-cookie exchange. Remote ingress remains classified separately so
+  host, passkey-presence, automation, and desktop-only route policies can keep
+  applying at their existing boundaries (#4559).
+
+### Security
+
+- Remote deployments no longer authenticate callers with an access token:
+  anything that can reach the Cave port can reach the app unless
+  `COVEN_CAVE_PASSKEY_REQUIRED=1` is enabled. The PTY sidecar-token gate remains
+  in place, which also means paired phones cannot open the packaged terminal
+  without a sidecar credential (#4559).
 ## [0.3.0] - 2026-08-11
 
 > Search that understands Cave's workspace, more legible multi-agent Chat,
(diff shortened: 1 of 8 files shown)


BunsDev:
Reviewed from the #4559 side (I authored the gate removal this stamps). The version bump looks right — all 6 files, same shape as the v0.3.0 stamp `b220817861`. One substantive concern on the changelog, plus three small ones.

### The Security paragraph is good; the headline undoes it

The Security bullet is exactly right and I wouldn't touch it:

> anything that can reach the Cave port can reach the app unless `COVEN_CAVE_PASSKEY_REQUIRED=1` is enabled

It also correctly catches the packaged-terminal consequence. But the headline directly above it is the first (often only) line a reader sees:

> Tokenless access across trusted local and tailnet surfaces, with optional passkey presence and local-only boundaries preserved.

That frames removing an authentication control as a feature — "tokenless access", "preserved". Someone who reads the headline and stops has learned the opposite of what shipped. Every other 0.x headline in this file describes a capability; this one is the only case where the headline and the Security section disagree about whether the release is good news.

Suggested replacement:

> ⚠️ Removes the access-token requirement. A Cave published over Tailscale Serve is reachable by anything on the tailnet with no credential. This partly reverts the loopback authorization added in 0.3.0 (#4533).

### Smaller

1. **Wrong category.** The removal is filed under `### Changed`. Keep-a-Changelog has `Removed`, which is what this is — the entry's own first word is "Removed".

2. **Missing blank line** before `## [0.3.0] - 2026-08-11`. Confirmed with `cat -A`: the last Security bullet runs straight into the previous release heading with no separating newline. Every other section in the file has one.

3. **No link to what this reverts.** v0.3.0 shipped one day earlier and its Security section advertises #4533 — *"Loopback access-token deployments now require user-bound authorization"*. This release partly undoes that. A reader comparing the two entries back to back should be told they are connected rather than left to infer it.

### One thing worth deciding explicitly

The file's preamble says *"patch releases stay additive."* This patch is not additive — it removes a security control. Either this entry acknowledges that explicitly, or the preamble becomes false as of 0.3.1. Flagging rather than prescribing, since the alternative is cutting this as 0.4.0 instead, which was considered and deliberately declined.

Not pushing to this branch — it is yours and holds the maintenance fence. Happy to send a commit if you would rather I did.

BunsDev:
Pushed `0e0e54f1a3` to this branch with the owner's approval, since your session had exited and the fence was released. Prose only — the version bump is untouched.

- Replaced the headline with the warning, folding in the #4533 reference so the connection to 0.3.0 is explicit.
- Moved the entry from `Changed` to `Removed`.
- Added a line noting this patch is not additive the way the file's preamble promises.

The blank-line fix in `2735b13642` is preserved. Reverting or rewording is entirely yours — I would rather you overwrite this than have it merge as something you disagree with.

PR #4567, opened 2026-08-12
fix: Chunk B ops hygiene (flake, DELETE cascade, exception report)

BunsDev:
## Summary

Knockout **Chunk B** — bounded ops/hygiene ready beads.

| Bead | Change |
|------|--------|
| **cave-pw3l0** | Raise `chat-boot-landing` "Start a new chat" heading timeout from default 5s → **15s** (matches sibling flaky specs) |
| **cave-eonxy** | `DELETE /api/projects/[id]` cascades grants **before** 404; 404 only when row and residue are both gone |
| **cave-4oor6** | Patrol reports **expired, orphaned** managed-creation exceptions (read-only; apply-side drop still gated) |
| **cave-ad63l** | Detach standalone-budget sample name from real fixture residue; residue dirs deleted locally |
| **cave-okvpy** | Day-pinned LaunchAgent removed locally (not in repo) |
| **cave-qdjl** | Closed — wrong repo (`OpenCoven/familiar-contract`) |

## Local ops (not in diff)

- `launchctl bootout` + removed `~/Library/LaunchAgents/com.opencoven.coven-cave-worktree-retire-2026-08-11.plist`
- Deleted `~1.9MB` `.worktree-lifecycle-fixture-{bWPzfW,eTlVOX}` from primary checkout
- Kept `.gitignore` / next tracing / standalone-budget prefix accommodations (safety net for any checkout-local fixture form)

## Test plan

- [x] `node --test src/app/api/projects/route.test.ts`
- [x] `node --test scripts/standalone-budget.test.mjs`
- [x] pure `listExpiredOrphanedExceptions` smoke
- [ ] CI Frontend build
- [ ] After merge: remeasure `chat-boot-landing` flake rate on main if still red


BunsDev:
Pushed `5ba161d8a3` to fix the `Frontend build` failure.

The merge that brought `main` in resolved `tests/chat-boot-landing.spec.ts` the wrong way — it kept this branch's 15s timeout bump and dropped main's #4564 (`cave-pw3l0`) fix, which left `projectScopeResolved` assigned but never awaited:

```
tests/chat-boot-landing.spec.ts(261,9): error TS6133: 'projectScopeResolved' is declared but its value is never read.
```

That was more than a lint nit — it reverted the actual repair. #4564 read the failure snapshot and found the first-project gate had *replaced* the launch surface: the two-familiar fixture invents familiars the seeded project does not grant, so `/api/projects` resolves empty and the gate opens every time. The heading never comes back, so a longer timeout fails slower rather than less often. #4564 measured 0 failures in 63 runs against 4 in 15 before.

I took main's version of that one file whole. The other five files on this branch are untouched.

PR #4575, opened 2026-08-12
fix(chat): expose Cave image attachments to familiar image tools (cave-cxwgy)

BunsDev:
## The bug

An uploaded Cave image was written to the OS temp directory, but the familiar
runtime boundary denies tmp. The bytes existed at a path the harness was never
granted, so a familiar's image tools could not read the user's own attachment.

A second cause compounded it: Charm's workspace-local imagegen skill already
existed, but familiar-local skill entrypoints were never surfaced in the chat
prompt, so the generic out-of-boundary copy won every time.

## Changes

**Staging moves inside the boundary.** `writeImageAttachmentsToWorkspace`
replaces `writeImageAttachmentsToTemp` and stages owner-only copies under a root
the runtime already grants (`resolvedFamiliarWorkspace ?? cwd`), removing the
directory once the harness exits. `mkdtemp` under a realpathed root, `0o700` on
the directory, `0o600` plus the `wx` exclusive flag per file so a pre-planted
symlink cannot be written through.

**Familiar-local skills become visible.** The familiar contract now indexes
contained `SKILL.md` entrypoints with local-over-generic precedence and
symlink-escape rejection.

**A leak path found in review, fixed here.** See below — it is the reason for
the second commit.

## The review finding

Moving staging out of tmp fixed delivery but changed what a leak costs, and the
leak path was still open.

`cleanupImageAttachmentDelivery` runs at exactly ONE site near the end of the
chat stream callback, and nothing wraps that callback in `finally`, so a throw
above that line skips it. That used to be harmless: abandoned copies sat in OS
temp storage, which the system reaps and no user sees. After this change they
sit in the familiar's granted workspace as `.coven-cave-attachments-*`
directories — user-visible, readable by the familiar, holding the user's images,
and reaped by nobody. `sweepChatImageAttachments` does not cover them: it matches
FILES named like an attachment id in the persistent store, so both of its filters
miss a directory in a workspace.

Fixed by sweeping stale staging directories at the start of the next delivery,
rather than restructuring a 2,700-line callback around a try/finally. Self-healing
and local to the module that creates them. Conservative by construction: same
realpathed root, staging prefix only, directories only, never a symlink wearing
the prefix, only past a one-hour age so a live turn's directory survives, capped
scan, all failures swallowed so a sweep can never break a turn.

## Provenance

The first commit preserves a prior session's work, which was complete and
verified but left **uncommitted on a local-only branch** — one `rm -rf` from
gone, in a checkout where a sweep destroyed a live worktree earlier the same day.
Committing and pushing it came before reviewing it, deliberately. The review then
produced the second commit.

## Testing

Re-run rather than inherited from the prior session's report:

- `chat-send-image-persistence` — 7/7, including the new sweep case (abandoned
  directory swept, fresh one kept, unrelated dot-directory untouched, delivery
  still succeeds). Mutation-checked: removing the sweep call fails it.
- `familiar-contract-context` — 26/26
- `harness-routing-attachments` — exit 0
- `tsc --noEmit` — clean
- `pnpm lint` — clean
- `pnpm check:tests-wired` — all 1625 wired
- `git diff --check` — clean

Known unrelated baseline failure, reported by the prior session and not
re-investigated here: `src/lib/use-surface-history.test.tsx` fails 12/12 with
`TypeError: act is not a function`, with zero diff from `origin/main` in that
test, its hook, `package.json`, or the lockfile.

Closes cave-cxwgy.

### DIFF
--- a/docs/familiar-identity-context.md
+++ b/docs/familiar-identity-context.md
@@ -1,8 +1,8 @@
 # Familiar identity context
 
-How a familiar's declared identity — `SOUL.md`, `IDENTITY.md`, and (for voice)
-`MEMORY.md` — reaches the model, and what that injection is explicitly *not*
-allowed to do.
+How a familiar's declared identity — `SOUL.md`, `IDENTITY.md`, familiar-local
+skill entrypoints, and (for voice) `MEMORY.md` — reaches the model, and what
+that injection is explicitly *not* allowed to do.
 
 Implementation: [`src/lib/server/familiar-contract-context.ts`](../src/lib/server/familiar-contract-context.ts).
 Consumers: the chat route (`src/app/api/chat/send/route.ts`) and voice
@@ -38,6 +38,7 @@ Both surfaces now share one builder, so the two cannot drift apart again.
 | --- | --- | --- |
 | `SOUL.md` | yes | yes |
 | `IDENTITY.md` | yes | yes |
+| `skills/*/SKILL.md` entrypoint index | yes | yes |
 | `MEMORY.md` | **no** | yes (`includeMemory: true`) |
 | `ward.toml` | never | never |
 
@@ -52,6 +53,12 @@ Both surfaces now share one builder, so the two cannot drift apart again.
 - **`ward.toml` is never inlined.** It is policy configuration carrying the
   `[protected]` file list and the invariants that bound self-modification.
   Injecting it hands the familiar the rules written to constrain it.
+- **Familiar-local skills are indexed, not inlined.** The block lists only
+  real `skills/<id>/SKILL.md` files contained by the familiar workspace and
+  states that a familiar-local same-name skill outranks a generic copy. This
+  lets the harness load the declared procedure from an already-granted root
+  without flooding every turn with every skill body. Symlink escapes are
+  omitted.
 
 ### Placement in the prompt
 
--- a/src/app/api/chat/send/chat-send-attachments.ts
+++ b/src/app/api/chat/send/chat-send-attachments.ts
@@ -1,5 +1,4 @@
-import { mkdir, realpath, rm, stat, writeFile } from "node:fs/promises";
-import { tmpdir } from "node:os";
+import { chmod, mkdtemp, realpath, rm, stat, writeFile } from "node:fs/promises";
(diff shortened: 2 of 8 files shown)


BunsDev:
Superseded by cave-xocwh (`fix/cave-xocwh-runtime-resources`), at the owner's
direction. Closing rather than merging, so the same fix does not land twice.

**Why.** cave-xocwh turned out to be a superset of this PR, not a sibling — its
bead notes already said so: *"cave-cxwgy is an overlapping active unit and must be
reconciled before PR."* Both branches move image staging out of `os.tmpdir` into a
granted root and index familiar-local `SKILL.md` entrypoints. That branch adds
four things this one does not:

- read-only grants for existing skill roots through the prompt, the boundary
  sentinel, and launch args, with remote runtimes kept closed;
- the workspace-local `coven:attachment` marker contract, forbidding delivery
  claims without a renderable marker;
- familiar workspace outputs;
- host-derived visible attachment metadata in reflection — which previously
  serialized only turn text and therefore erased the evidence of the very
  attachments this PR set out to deliver.

That last one is decisive: without it, delivery works but proof of delivery is
silently destroyed downstream.

Its staging design is also better on concurrency. It uses one fixed
`.coven-cave-attachments` directory with UUID-named files and a NON-recursive
directory removal that succeeds only once the last concurrent turn has removed
its file, where this PR created a fresh `mkdtemp` directory per turn.

**Carried across, not discarded** — the two findings from this PR's review:

1. The stale-staging sweep. `cleanupStagedImageFiles` is still called at one site
   with no `finally` around the stream callback, so a throw above it strands the
   user's images inside their granted workspace, where nothing reaps them.
   `sweepChatImageAttachments` does not cover them. Adapted to sweep stale FILES
   in the fixed staging directory rather than whole directories.
2. The offline-queue ordering probe. That branch already repointed the renamed
   symbol, but not the `indexOf` -1 case, where a stale probe fails as an
   ordering violation and sends the reader hunting a reordered handler that is
   fine. The explicit not-found assertion comes across too.

Nothing is lost by closing this: both commits remain on the branch and the work
is fully represented in the superseding PR.

PR #4578, opened 2026-08-12
fix(research): fall back to a direct Copilot spawn when the native supervisor is absent

BunsDev:
Research has been broken for **every** user since #4524 merged (2026-08-10). This restores the launch path.

## The bug

#4524 replaced direct-run tracking with an owned native Coven supervisor lease. Cave now probes `coven --print-native-binary-path` and requires a `process-supervisor` subcommand — **which no published `@opencoven/cli` provides**:

```
$ coven --version
coven v0.3.1 (engine coven-code 0.7.0, pinned 0.7.0)
$ coven --print-native-binary-path
Error: unrecognized flag `--print-native-binary-path`
$ npm view @opencoven/cli version
0.3.1
```

The installed CLI *is* the newest release. There is no fallback in Cave, so every mission is created and then fails at iteration 1:

> Coven's native process supervisor is unavailable. Update or repair the Coven CLI, then retry Research.

The error's advice is unfollowable — there is nothing newer to install.

## The fix

Fall back to the pre-#4524 transport when the supervisor is genuinely missing. The supervisor proxies Copilot's JSONL stream rather than transforming it, so a direct spawn reuses every frame handler unchanged; the admission handshake is supervisor-only and is skipped rather than failed.

Deliberately narrow:

- **Only** `CovenProcessSupervisorUnavailableError` triggers it — a supervisor that exists but misbehaves still fails loudly (pinned by a test).
- `requireProcessSupervisor` opts back into the hard failure for callers that must prove descendants died.
- `assertCopilotCommandLineFitsWindows` still runs. It matters *more* here, because the fallback puts the prompt back in argv — the exact Windows 32,767-unit hazard #4524 moved into the request frame.
- The degradation is surfaced on the run via `compatibilityDiagnostics`, so an unsupervised mission does not read as a normal one.

## What this gives up

An unsupervised run owns only the immediate child. Cancel and timeout stop Copilot itself, but **processes it spawns are not owned and may survive**. Strictly better than a feature that cannot run; strictly worse than the supervisor. The durable fix is a Coven CLI that ships `process-supervisor` — this is a workaround, and `cave-hhwc5` should stay open until that exists.

## Verification

Against a **real mission** on this branch: the launch now reaches Copilot, which runs ~30s of real work. On `main` the identical request fails instantly and never starts a process.

⚠️ **The mission still ends `failed — "Owned session ended"` at transcript reconciliation.** That is a **separate defect downstream of the spawn**, not this change: the Copilot CLI emits a clean terminal `result` frame (`exitCode: 0`) when run standalone, so neither the environment nor the protocol is at fault. This PR unblocks the launch; it does not by itself make research complete end to end.

`flow-copilot-session` suite: **31 pass, 0 fail**, including three new tests (fallback taken, `requireProcessSupervisor` refuses it, non-availability errors are not downgraded).


BunsDev:
Pushed `5fda10be6a` to fix the `Frontend build` failure.

`src/lib/server/flow-executor.test.ts` reads its subject as source text, and **four** assertions pinned `binding.harness` as an exhaustive literal. This branch renames that to a local `harness` and adds `harness`/`model` pass-through to the Research adapter — no behavior change, but all four regexes stopped matching. CI reported the first one:

```
AssertionError: flow session spawn must not pass familiarId natively to the daemon
```

which is misleading: `familiarId` was never in the spawn body, the regex just also pinned the two lines around it.

That guard is the load-bearing one, and the literal made it **weak in both directions** — it broke on any unrelated edit nearby, while still passing if `familiarId` were added under different formatting. It now locates the spawn body and asserts the property itself: `projectRoot` present, `launchMode: "nonInteractive"`, and `familiarId` **absent**.

Mutation-tested — inserting `familiarId,` into the spawn body fails with exactly its own message. The other three now accept either spelling, or tolerate added keys between the two anchors that carry the guarantee, without loosening what they assert.

`flow-executor.test.ts: ok` locally.

PR #4589, opened 2026-08-13
Fix stale archive state in chat sidebar

CompleteDotTech:
## Summary
- remove the obsolete archived-session fetch and visibility projection from `WorkspaceSidebar`
- keep the Chat rail archive-free while retaining row-level archive/unarchive actions
- update focused source-contract assertions for the normalized session input and label markup

## Verification
- `node --experimental-strip-types src/components/chat-siderail-hide-archived.test.ts`
- `node --experimental-strip-types src/components/workspace-sidebar-wiring.test.ts`
- `pnpm typecheck`
- `pnpm lint`
- `pnpm check:tests-wired`
- `pnpm test:app`
- `git diff --check`

Bead: `cave-2ok`

### DIFF
--- a/src/components/chat-siderail-hide-archived.test.ts
+++ b/src/components/chat-siderail-hide-archived.test.ts
@@ -81,7 +81,7 @@ assert.doesNotMatch(
 );
 assert.match(
   workspaceSidebar,
-  /const visibleSessions = useMemo\(\s*\(\) =>\s*filterVisibleChatSessions\(sessions, activeFamiliarId \?\? null\),\s*\[sessions, activeFamiliarId\],\s*\);/,
+  /const visibleSessions = useMemo\(\s*\(\) => filterVisibleChatSessions\(normalizedSessions, activeFamiliarId \?\? null\),\s*\[normalizedSessions, activeFamiliarId\],\s*\);/,
   "the Chat side rail must derive visible sessions with the archive-free default",
 );
 assert.doesNotMatch(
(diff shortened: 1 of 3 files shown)


BunsDev:
Superseded by #4590, which reverted the merge this was repairing.

`main` is green again at `2665d1ed64`, and `workspace-sidebar.tsx` is back to its self-consistent pre-merge form — so this diff no longer applies: it removes state that the revert has legitimately restored.

Worth carrying into the re-land, since this PR could not have passed as-is. The half-landed merge broke **six** things, and `test:app` reveals them one at a time because the runner stops at the first failing file — this PR's own `Frontend build` failed on the second one:

1. `workspace-sidebar.tsx` — 13 TS errors, plus a runtime `ReferenceError: showArchived is not defined` that crashed the chat surface to the error boundary (what this PR fixed)
2. `chat-session-grouping.test.ts` — still pins the retired group-by `<select>`; `chat-list.tsx` now renders the segmented control
3. `chat-sidebar-wiring.test.ts` — asserts `<PopoverLabel>Chat visibility</PopoverLabel>`
4. `chat-siderail-hide-archived.test.ts` — `visibleSessions` regex predates `normalizeSessionAttention`
5. `workspace-sidebar-attention.test.ts` — two behavioral tests reach an archived pinned row only through the deleted "Show archived" item, so the scenario is unreachable by design; repairing them deletes a cave-zs85n Task 6 regression guard
6. `nav-section-tabs.test.ts` — four assertions positively pin the `Popover` imports, `showArchived` state, `includeArchived` fetch and `PopoverItem` wiring, and the file already contradicts itself about whether "Sidebar options" exists

The branch itself was never at fault: `3db955124d` removed the apparatus cleanly and consistently. `main` had diverged on the same file via #4405 and #4391, and the merge resolution reinstated the consumers while keeping the branch's deletion of the declarations.

Full write-up on `cave-9e83f`, cross-referenced from `cave-zdbij`. A re-land needs the component and its test contract in one change.

PR #4596, opened 2026-08-13
fix(copilot): reject clients without session ID support

snowopsdev:
Require Copilot CLI 1.0.70 or newer for the direct JSONL transport in Cave.

Cave preassigns session IDs so chat and Research runs have durable ownership. Copilot CLI 1.0.64-1 advertises `--session-id` in help but exits with `unknown option` when Cave launches it, causing Research to fail before its first phase. The prior protocol range admitted every 1.x client.

This raises the built-in protocol floor to the first version whose session-ID and JSONL contract was verified end to end, and adds regression coverage at the protocol-selection and chat-routing boundaries. Older clients now fail the compatibility gate with an actionable update path instead of starting a doomed run.

Focused E2E validation reproduced the old-client exit and verified the same mission reached a Research checkpoint on Copilot 1.0.79 with all six phases completed. Lint, test wiring, typecheck, targeted tests, and all 1,180 app test files pass. The full API suite reaches an unrelated environment-dependent fixture failure: `route-claude-availability.integration.test.ts` expects `runtime_coven_missing` but receives `runtime_process_failed` when a real Homebrew Coven CLI is present; the same failure reproduces on clean base `5b61ea412`.

### DIFF
--- a/src/app/api/chat/send/harness-routing-copilot-jsonl.test.ts
+++ b/src/app/api/chat/send/harness-routing-copilot-jsonl.test.ts
@@ -65,7 +65,7 @@ assert.deepEqual(
   "the direct full-mode routing decision keeps its approval flag before the reviewed JSONL launch contract",
 );
 
-for (const capabilityVersion of ["1.0.70.1", "0.9.9", "2.0.0", "2.0.0-rc.1"]) {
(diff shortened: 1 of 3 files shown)


snowopsdev:
Superseded by #4597. GitHub automatically closed this fork PR when its head branch was renamed from `t3code/fix-daemon-research-connection` to `snowopsdev/fix-daemon-research-connection`; the replacement draft preserves the same verified commit and uses the requested structured description.

Issue #3283, opened 2026-07-16
Simplify CovenCave navigation and reduce UI complexity


CompleteDotTech:
## Summary

The current CovenCave UI has become difficult to understand because several navigation systems, compatibility modes, nested tabs, and overlapping names now coexist.

A source-level navigation audit found:

- 17 workspace modes, including aliases and retired compatibility modes
- 20 standalone Next.js page routes
- separate desktop, chat-specific, mobile-top, mobile-bottom, command-palette, and analytics-route navigation
- major destinations implemented as nested tabs behind legacy modes
- overlapping product names such as Tasks / Board, Rituals / Rites / Schedules / Inbox, and Familiars / Agents
- many drawers, inspectors, overlays, context menus, split panes, and role-dependent Rooms

The attached Mermaid maps capture the current UI before simplification.

## Goal

Create a coherent information architecture with a small, stable set of top-level destinations and predictable nested navigation. Preserve existing functionality and deep-link compatibility while reducing the number of concepts a user must understand.

A possible organizing model is:

1. Home
2. Chat
3. Work
4. Knowledge
5. Settings

This is a starting point for design exploration, not a mandated final taxonomy.

## Problems to solve

- Canonicalize names so one surface has one user-facing name.
- Stop exposing compatibility modes as separate navigation concepts.
- Align desktop and mobile around the same hierarchy.
- Clarify which UI is a destination, tab, inspector, drawer, or temporary overlay.
- Reduce duplicate entry points that do not add meaningful convenience.
- Keep dynamic Role rooms discoverable without mixing them into universal navigation.
- Preserve direct links, keyboard shortcuts, session restoration, and automation entry points.

## Proposed work

- [x] Audit every current navigation entry against the attached source maps.
- [x] Define and approve the target information architecture and canonical vocabulary.
- [x] Produce a current-to-target mapping for every route, workspace mode, tab, and Role room.
- [x] Consolidate top-level navigation across desktop and mobile.
- [x] Move secondary destinations into clearly labeled tabs or contextual actions.
- [x] Keep legacy modes as internal redirects only where backward compatibility requires them.
- [x] Update command-palette results and keyboard shortcuts to use canonical names.
- [x] Update navigation, deep-link, accessibility, and mobile tests.
- [x] Update these Mermaid maps to document the resulting architecture.

## Acceptance criteria

- Every currently reachable production surface remains reachable.
- Desktop and mobile present the same conceptual hierarchy.
[...]

PR #4618, opened 2026-08-14
fix(tests): assert the initialTab contract, not the literal expression

BunsDev:
`main` is red at `2f30bf6ae`. `Frontend build` fails on
`src/lib/beads-work-queue.test.ts:229` — *"the legacy mode must deep-link onto
the queue tab"* (232 tests passed before it).

## This is not a regression

`cave-x6rw` (`19fff57bf`, *"feat(workspace): embed registered split pages"*)
extended two expressions in `workspace.tsx` so an embedded split page can request
a tab through its `variant`:

```diff
-  initialTab={mode === "familiar-work-queue" ? "queue" : "tasks"}
+  initialTab={mode === "familiar-work-queue" || variant === "queue" ? "queue" : "tasks"}
-  initialTab={mode === "calendar" ? "calendar" : "overview"}
+  initialTab={mode === "calendar" || variant === "calendar" ? "calendar" : "overview"}
```

`variant` is a typed parameter of `renderSurface` (`WorkspacePageVariant`,
defaulting to `"default"`), used the same way for chat scope, grimoire view and
roles/capabilities. **The legacy contract still holds** — `mode ===
"familiar-work-queue"` short-circuits to `"queue"`. The behaviour is intact; four
source-text assertions pinned the exact literal, so an additive change read as a
break.

That branch reached `main` by direct push rather than a PR, so no PR run ever
exercised it.

## Four assertions were stale, not one

The runner stops at the first failure, so fixing only the reported assertion
produces a build that fails again immediately. All three files fail locally on
`2f30bf6ae`:

| file | line | tab |
|---|---|---|
| `src/lib/beads-work-queue.test.ts` | 228 | queue |
| `src/components/rituals-tabs.test.ts` | 130 | calendar |
| `src/components/workspace-alias-modes.test.ts` | 42 | queue |
| `src/components/workspace-alias-modes.test.ts` | 54 | calendar |

## The fix

Each assertion now admits further disjuncts with `[^}]*` while still requiring
the **mode** to select its tab over the default:

```diff
-  /initialTab=\{mode === "familiar-work-queue" \? "queue" : "tasks"\}/
+  /initialTab=\{mode === "familiar-work-queue"[^}]*\? "queue" : "tasks"\}/
```

Relaxed, not defeated. Verified against `origin/main`'s `workspace.tsx`: all four
patterns match, **and** a source with the mode clause removed
(`variant === "queue"` alone) is still rejected — so the assertions keep catching
the regression they exist for.

The tests are asserting a *contract* — the legacy mode deep-links onto its tab —
and pinning an exact expression was over-specification that turned any additive
edit into a red `main`.

## Verification

- All four patterns matched against `origin/main`'s `workspace.tsx`; the
  negative case (mode clause deleted) still fails, confirming the guard survives.
- Full unit suite and build run in CI on this PR.

Built with git plumbing rather than a working tree: this session's worktree was
destroyed twice mid-edit by the local sweeper, so the commit was assembled from a
temporary index over `origin/main`, leaving the shared checkout untouched.


BunsDev:
**Coordination request: please hold pushes to this branch until one CI run completes.**

Two of us are working this PR concurrently (I've been pushing the workspace/registry
contract fixes; the `.worktrees/cave-iylqb-red-main-queue-pin` worktree has been
pushing the Copilot autofixes and the `main` merges). That's been complementary on
content — every one of my commits survived every one of those pushes, and the
alias-loader fix landed from that side before I could push the same diagnosis.

The problem is timing, not content. `Frontend build` takes ~35 minutes on this
branch now (~13 min unit suite + ~20 min Playwright). Three separate runs have
reached deep into `Validate end-to-end behavior` and been superseded by a new push
before they could report:

| head | outcome |
|---|---|
| `de209cf62` | cleared the whole unit suite, ~20 min into e2e, superseded |
| `ac1363d96` | cleared the whole unit suite, ~20 min into e2e, superseded |
| `11aceeb89` | current |

So the check may never report green while pushes keep arriving faster than a run
takes — regardless of whether the fixes are correct.

**The good news:** the last two runs both cleared `Validate frontend` end to end.
The unit suite and build pass. The only unproven step is Playwright.

**Ask:** please don't merge `main` into this branch or push further commits until
`Frontend build` reports on the current head. If something on `main` is genuinely
required here, say so and I'll hold instead — I just don't want us both pushing
into a 35-minute validation window.

For reference, the substantive fixes on my side, all verified to still reject the
regression each pin exists to catch:

- the `variant`-disjunct pins (`initialTab` / `initialSection` / `view=`)
- the `WORKSPACE_MODE_TITLES` → page-registry migration
- **a real regression**: `calendar` and `familiar-work-queue` titles broke the
  issue #3283 alias-vocabulary contract (`"Calendar"`/`"Work queue"` instead of
  `"Rituals"`/`"Tasks"`), fixed in the registry rather than by relaxing the test
- the `SalemChatPanel` → `AskSalemView` re-homing, swept across all four pins
- the segment-splitter arity pins

One correction on the record: I initially "restored" the drag-to-split exclusions
as a regression. That was wrong — `177d96fd2` deliberately retired them and shipped
`page-drag.test.ts` asserting the new contract. I reverted that and fixed the
genuinely stale assertion in `journal-redirect.test.ts` instead.

BunsDev:
**Correction to my earlier comment — the cancellations were not a coordination problem.**

I asked for a push freeze because three runs died deep in `Validate end-to-end behavior`. That diagnosis was wrong, and the freeze was not the fix. The last three runs were all on the **same head** (`11aceeb89`) with no intervening push, and each ran ~60–64 minutes before being cancelled:

| run | created | ended | |
|---|---|---|---|
| `31811258570` | 14:47:27 | 15:48:01 | cancelled |
| `31816366858` | 15:49:21 | 16:53:16 | cancelled |
| `31824385068` | 17:31:30 | 18:34:57 | cancelled |

`Frontend build` carries `timeout-minutes: 60`. `tests/workspace-half-split.spec.ts` was failing ~35 times, each burning a 60s `toBeVisible` timeout, so the job blew past the cap and GitHub reported **cancelled** instead of **failed**. The timeout masked a real, systematic test failure as infra flakiness.

**The underlying bug is in the split deep-link path, and it is a product bug.**

`workspace.tsx` strips any split target whose key matches the primary:

```ts
useEffect(() => {
  const primary = primaryPaneRequest ?? normalizeWorkspacePaneRequest("primary", mode);
  ...filter(target => key(target) !== key(primary))
}, [mode, primaryPaneRequest]);
```

The `?mode=X&split=Y` deep link resolves in a mount effect that queues `setMode(X)` and the split in the same commit — so this effect still sees the **default** mode (`home`) as the primary and deletes a freshly deep-linked `split=home` before the mode update lands. `referencePage()` pairs every page against `home`, so this poisoned half of every case in the spec.

Fixed in `d742712be`, two parts:
1. skip the dedupe's mount pass (splits only exist there because the deep link put them there, and that path already refuses a secondary matching the primary); 
2. compare against the **effective** primary, not just an explicit `?mode=` — fixing (1) alone regressed `?split=home` into splitting home against itself.

Verified in a real browser against this branch:

```
?mode=board&split=home    1 pane -> 2 panes   (the bug)
?split=home  (no mode)    1 pane              (correctly refuses self-split)
?mode=home&split=board    2 panes
?mode=board&split=github  2 panes
```

and the spec itself: **28 passed (2.0m), exit 0** — from ~35 failures.

Worth noting this spec shipped with `cave-x6rw` through a direct push, so it had never run in PR CI. Its first honest run is the one that kept getting cancelled.

The push freeze is no longer needed — please carry on. Sorry for the misdirection.

BunsDev:
Superseded by #4648.

#4637 landed an equivalent of this branch's pin repairs and its own (better) fix for the split deep-link race — a synchronous `primaryPaneRequestRef` that also covers `commitMode`. #4626 landed the alias-loader registration. That left this branch conflicting with `main` across 28 files where `main`'s version should win.

#4648 is rebuilt on current `main` and carries only the three fixes `main` still lacks: the mobile inactive-pane CSS collapse, the half-split spec's viewport, and the stale `Chat options` assertion in `chat-boot-landing.spec.ts`.

One correction worth recording: this branch also changed `workspace-page-registry.ts` to retitle the `calendar` and `familiar-work-queue` aliases. That was wrong — those aliases deliberately keep titles distinct from their canonical entries (`inbox` → "Rituals", `board` → "Tasks"), with the relationship carried by `landmark`. `main` is correct and that change is dropped.

PR #4621, opened 2026-08-14
feat(research): 25k intent cap, a live character count, and a growing prompt box

BunsDev:
Three changes to the Research Desk mission composer. Bead `cave-r3vmj`.

## What changed

| | |
|---|---|
| **Cap** | `RESEARCH_INTENT_MAX_LENGTH` 10,000 → **25,000** |
| **Count** | live `1,234 / 25,000` inside the box, amber past 90% |
| **Size** | grows with the brief, capped at ~12 lines, was a fixed 3 rows |

A note on the request's wording: it asked to raise the *minimum*. `RESEARCH_INTENT_MIN_LENGTH` is 8 and exists to block one-word launches that still spend a familiar session — a 25,000 minimum would make the desk unusable — so the cap is what was being hit. Flagging the interpretation rather than burying it.

The counter is deliberately **not** a live region; announcing every keystroke would drown the composer. The textarea also gained `maxLength`, so the browser now enforces what the server validator already did. The height cap is expressed in `lh` so it tracks the line-height if the type scale moves, with a px fallback. Resizing runs on every intent change, not just typing, so slash completions and `/improve` rewrites resize it too.

## A real ceiling, deliberately accepted

Raising the cap surfaced a genuine constraint rather than a stale test:

```
CopilotPromptTransportError: Copilot flow prompt is too large for a safe Windows
launch (42434 UTF-16 command-line units; safe limit 30000).
```

Copilot takes its flow prompt through **argv only**, and `flow-copilot-session.ts:299` bounds that at 30,000 UTF-16 units **on `win32`**. The scaffolding around the intent costs ~17,400 units, so the practical ceiling for that one path is ~12,500 characters.

So a maximum-length brief cannot launch through **Copilot on Windows**. The existing guard fails closed with a message naming the cause and the remedy — it does not truncate a prompt. **Every other platform, and any runtime with stdin prompt support, is unaffected.**

`flow-copilot-session.test.ts` previously asserted that a maximum-length intent always fits. That is no longer true, so it now pins the honest contract:

- the plain maximum exceeds the argv budget and **throws on `win32`**
- the **same input is fine on `darwin`** — the ceiling is a Windows argv limit, not a product-wide one
- an **8k brief still launches**, so the ceiling is a real size rather than a blanket refusal

Moving Copilot to stdin is what would lift this properly; that is separate work and worth its own bead.

## Verification

`tsc --noEmit` clean · lint clean including the design tokenizer (0 drift) · `ui-consistency` ok · `check-tests-wired` 1,662 · six research suites at exit 0 · `flow-copilot-session` 32 passing.

One implementation note: the composer already had an `intentRef` holding the intent *string* for the enhance race rule, so the new element ref is `intentBoxRef` — the existing race guard is untouched.

### DIFF
--- a/src/components/role-surfaces/research-mission-composer.tsx
+++ b/src/components/role-surfaces/research-mission-composer.tsx
@@ -32,6 +32,7 @@ import {
 } from "@/lib/research-mission-routing";
 import {
   RESEARCH_BOUND_LIMITS,
+  RESEARCH_INTENT_MAX_LENGTH,
   RESEARCH_INTENT_MIN_LENGTH,
   RESEARCH_MISSION_MODES,
   type CreateResearchMissionInput,
@@ -274,6 +275,18 @@ export function ResearchMissionComposer({
   const plan = useMemo(() => defaultResearchPlan(effectiveMode), [effectiveMode]);
   const trimmedIntent = intent.trim();
   const intentTooShort = trimmedIntent.length > 0 && trimmedIntent.length < RESEARCH_INTENT_MIN_LENGTH;
+  // Grow the box with the brief instead of scrolling three fixed rows. The CSS
+  // caps it at ~12 lines, so `auto` then scrollHeight settles at the smaller of
+  // content height and that ceiling; past it the element scrolls as before.
+  // Measured on every intent change, including programmatic ones (slash
+  // completions, /improve rewrites, restored drafts), not just typing.
+  const intentBoxRef = useRef<HTMLTextAreaElement | null>(null);
+  useEffect(() => {
+    const el = intentBoxRef.current;
+    if (!el) return;
+    el.style.height = "auto";
+    el.style.height = `${el.scrollHeight}px`;
+  }, [intent]);
 
   // Strength + assembled brief are pure reads of the textarea, so they track
   // hand edits made after the builder closed.
@@ -477,6 +490,8 @@ export function ResearchMissionComposer({
           <label htmlFor="research-intent" className="sr-only">What should we investigate?</label>
           <textarea
             id="research-intent"
+            ref={intentBoxRef}
+            maxLength={RESEARCH_INTENT_MAX_LENGTH}
             value={intent}
             onChange={(event) => {
               setIntent(event.target.value);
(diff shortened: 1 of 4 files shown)


BunsDev:
Closing — superseded by #4624, which strictly contains this change and extends it.

#4624 carries every part of this PR: `RESEARCH_INTENT_MAX_LENGTH` 10,000 → 25,000, the live counter with `toLocaleString()`, the `* 0.9` near-limit class, `maxLength` on the textarea, the autosize effect, the ~12-line CSS ceiling, and the `win32`/`darwin` split in the copilot transport test — down to the `intentBoxRef` name chosen here after the collision with the pre-existing `intentRef`.

It then does what this PR didn't: applies the same counter to the research **direction** field, converts the generation-directions warning from an absolute `- 200` to the proportional `* 0.9`, and shares a `GITHUB_REVIEW_BODY_MAX_LENGTH` cap between the review UI and its route. That's the coherent version of the same idea.

Keeping both open would mean two PRs editing the same four files — that manufactures conflicts rather than resolving them.

**Two things worth carrying forward into #4624's review**, both verified here:

1. **The 25k cap has a real ceiling on one path.** Copilot takes its flow prompt through argv only, and `flow-copilot-session.ts:299` bounds that at 30,000 UTF-16 units on `win32`. A maximum-length intent assembles to ~42,400, so a full-length brief cannot launch through **Copilot on Windows**; the guard fails closed with an actionable message rather than truncating. Every other platform and any stdin-capable runtime is unaffected. `cave-cwjof` tracks moving Copilot to stdin, which is what lifts it.

2. **The test that documents that ceiling was rewritten deliberately.** It previously asserted a maximum-length intent always fits, which stopped being true. It now pins the honest contract — throws on `win32`, fine on `darwin`, and an 8k brief still launches — so please don't "restore" the old assertion if it looks odd.

This PR never ran CI: it was force-pushed/rebased six-plus times by the GitHub web UI (`committer=GitHub <noreply@github.com>`) faster than a ~30 minute run could start, so every queued run was superseded. Worth knowing if #4624 stalls the same way — the fix is to leave the branch alone rather than keep clicking "Update branch"; `strict: false` means being behind `main` never blocks a merge here.

PR #4622, opened 2026-08-14
feat(code): collapsible sessions rail that keeps project context when collapsed

BunsDev:
## Summary

Wraps the Code Workshop session list in the shared `SurfaceRail`, so the sessions rail collapses and resizes like every other left rail and remembers its state per surface. Collapsed rows stay icon-only without losing what they identify.

## Why

The Code Workshop was the last surface whose left rail could not collapse — the list was a fixed 256px column, so a Room split beside another page gave the workbench whatever was left. Reusing `SurfaceRail` makes the behaviour and the persistence identical to the other surfaces instead of Code-specific.

Bead: `cave-24d2r.1`.

## Changes

- `code-view.tsx` renders `SurfaceRail` (`storageKey="cave:code:sessions-rail"`, `defaultWidth={CODE_ROOM_RAIL_WIDTH_PX}`) **only when the measured Room fits it**. `codeRoomFitsRail` is preserved, so a narrow Room keeps the list-first drill-in and Back navigation rather than collapsing into an unusable sliver.
- `code-session-rail.tsx` gains `open` / `onExpand` and renders a collapsed variant: a 32px icon button per session with the activity dot pinned to its corner.
- Collapsed rows keep their meaning without color: `aria-label` carries session title, project group, and activity, with the title also in `title=`. That was a second-review finding — the first collapsed implementation dropped project context from the accessible name, which made two sessions from different projects indistinguishable to a screen reader.
- Clicking a collapsed row **expands before selecting** (`onExpand?.()` then `onSelect`), so picking a session never leaves you in a collapsed rail looking at a workbench you cannot navigate back from.
- `selectSession` is extracted in `code-view.tsx` and clears `workbenchTarget`, so a manual switch drops any pending file focus instead of replaying it into the newly picked session.

## Verification

Verified in full on 2026-08-08 at base `7c7e5aaac` (per `cave-24d2r.1`): focused `code-surface-mode`, `surface-rail` 10/10, `code-surface` 17/17; `pnpm test:app` 1137 files; `pnpm test:api` 357 files; `pnpm typecheck`; `pnpm lint` with zero design drift and zero ESLint warnings; `check:tests-wired` 1570 files; `git diff --check`.

`code-surface-mode.test.ts` adds source-text assertions for each behaviour above — the rail wiring and storage key, the fits-vs-narrow branch, the collapsed accessible name, and expand-before-select.

## Current status

⚠️ Not mergeable yet, for reasons outside this change:

- **Conflicts with `main`** in `code-session-rail.tsx` and `code-surface-mode.test.ts` (the branch's merge-base is `7c7e5aaac`; `main` has since taken the Coding Room rebuild and more).
- **`main` is red** — `a633d22db` merged the `cave-x6rw` split feature directly to `main` with no PR and no CI, breaking 20 test files. PR #4618 is repairing them, and it touches `code-surface-mode.test.ts` too, so the conflict here is best resolved *after* #4618 lands rather than twice.

### DIFF
--- a/src/components/code-session-rail.tsx
+++ b/src/components/code-session-rail.tsx
@@ -122,7 +122,7 @@ export function CodeSessionRail({
                       onSelect(row.id);
                     }}
                     aria-current={selected ? "true" : undefined}
-                    aria-label={open ? undefined : `Open ${title}, ${activity}`}
+                    aria-label={open ? undefined : `Open ${title} in ${group.label}, ${activity}`}
                     title={open ? undefined : title}
                     className={
                       open
--- a/src/components/code-session-rail.tsx
+++ b/src/components/code-session-rail.tsx
@@ -51,11 +51,20 @@ export type CodeSessionRailProps = {
   selectedId: string | null;
   onSelect: (sessionId: string) => void;
   onNewSession?: () => void;
+  open?: boolean;
+  onExpand?: () => void;
 };
 
-export function CodeSessionRail({ sessions, selectedId, onSelect, onNewSession }: CodeSessionRailProps) {
+export function CodeSessionRail({
+  sessions,
+  selectedId,
+  onSelect,
+  onNewSession,
+  open = true,
+  onExpand,
+}: CodeSessionRailProps) {
   const groups = groupCodeRailSessions(sessions);
-  const newButton = onNewSession ? (
+  const newButton = open && onNewSession ? (
     <div className="px-2 pb-1">
       <button
         type="button"
@@ -68,6 +77,9 @@ export function CodeSessionRail({ sessions, selectedId, onSelect, onNewSession }
     </div>
   ) : null;
   if (groups.length === 0) {
+    if (!open) {
+      return <nav aria-label="Coding sessions" className="flex h-full min-h-0 flex-col" />;
+    }
     return (
       <div className="flex h-full flex-col py-2">
         {newButton}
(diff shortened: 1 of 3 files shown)


BunsDev:
## This is already on `main` — merging it would regress two later fixes

I rewrote the title and description from the diff, then went to resolve the conflict. The conflict turns out to be the finding: **`main` already contains this entire feature**, and has moved past it.

Present on `main` today:

- `storageKey="cave:code:sessions-rail"` wiring in `code-view.tsx` ✓
- the rail assertion in `code-surface-mode.test.ts` ✓
- the collapsed `aria-label` carrying session + project + activity ✓

So there is nothing here left to land. Worse, `git diff origin/main HEAD` shows this branch would **remove** work that arrived after it:

- **`const openRailClassName = onExpand ? "py-2" : "overflow-y-auto py-2"`** and its explanatory comment, plus the assertion pinning it. That is the `cave-iixug` fix: `SurfaceRail`'s own `.surface-rail__content` already scrolls, so owning `overflow-y-auto` here too nests a second scroll container. Merging this PR reintroduces that.
- the doc-comment rename **"Coding Room" → back to "Coding Desk"**.

Net effect of merging: −21 lines from `main`, +3, all of it undoing later refinements.

### Recommendation

**Close without merging.** The owning bead `cave-24d2r.1` is already `CLOSED`, which is consistent — the work landed by another route. The branch's two commits (`eebb23727`, `94498f63c`) remain reachable, and I can archive-tag them if you want a named ref before the branch is retired.

Nothing was pushed to this branch. I did attempt the merge locally to inspect the conflict, and restored the worktree to exactly the state I found it — same branch, same commit `94498f63c`, clean.

BunsDev:
Closing unmerged: the feature is already on `main`, and this branch would regress the later `openRailClassName` fix (cave-iixug) and the Coding Room rename — see the analysis above.

Both commits are retained at `archive/fix-cave-24d2r-1-code-sessions-rail-2026-08-14` (pushed), so nothing is lost if the branch is later reaped. The owning bead `cave-24d2r.1` is already closed.

PR #4638, opened 2026-08-14
fix(tests): the source-text pin repairs, split out so main can go green

BunsDev:
`main` has been red since 2026-08-14T09:50Z. Every branch cut since inherits it, and three PRs are blocked behind it. Bead `cave-64t9u`.

## Why a second PR

The repairs live on **#4618**, which has grown to 28 commits: 17 pin fixes plus nine commits of unrelated work — a workspace deep-link feature, a drag-to-split restore *and its own revert*, three commits titled "Potential fix for pull request finding", a WIP snapshot.

That branch cannot converge. Six CI runs, **zero completions**: the head moves roughly every 20 minutes while a full run needs about 30, so it never accumulates a passing required check.

This PR is the pin repairs **only**, cut fresh from `main`, single-purpose and small enough to finish one uninterrupted run.

## What's in it

20 commits: the 17 `fix(tests)` pin repairs, plus three that don't say `fix(tests)` but are required — found by running the suite rather than by reading prefixes:

| Commit | Why it's needed |
|---|---|
| `de209cf62` | registers `beads-delivery-source` with the alias loader; without it that file dies at import and masks everything behind it |
| `9ca8ee4e3` | labelled `fix(workspace)` but is mostly pin migration — re-points `code-surface-mode`, `grimoire-view`, `rituals-tabs`, `workspace-alias-modes` at the registry |
| `e7983801a` | labelled `revert(page-drag)`; only its **test** half applies here, re-pinning `journal-redirect` to match exclusions x6rw retired deliberately |

All cherry-picked cleanly with **no conflicts** — evidence the repairs really are self-contained.

## Root cause, for the record

One refactor (`cave-x6rw`, "embed registered split pages") broke seventeen source-text pins. It reached `main` through `a633d22db`, a local merge with **no pull request**, so the required check never ran on it. Not one of the seventeen caught a defect; every one broke on a correct change. The convention written afterwards is [`docs/source-text-pins.md`](docs/source-text-pins.md) (#4625).

## Verification

Local `pnpm test:app` advanced **716 → 878 → 1,176 passing** as each prerequisite was found. It now stops only at `automation-runner.test.ts`, which is the macOS `/var` → `/private/var` symlink artifact:

```
actual:   /private/var/folders/.../codex.exe
expected: /var/folders/.../codex.exe
```

Confirmed **pre-existing and local-only** — it fails identically on clean `main`, and CI runs Ubuntu where `/var` is not a symlink. `check-tests-wired` passes at 1,662 files.

## Note for #4618

Nothing here is lost from that branch — these commits remain on it. Once this lands, #4618 rebases onto a green `main` and carries only its feature work, which can then be reviewed on its own merits instead of blocking the repository.

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -1721,6 +1721,11 @@ const STRIP_TYPES_MJS = new Set([
 // Tests whose import graph reaches the "@/..." path alias and therefore need
 // the alias-resolving loader (`scripts/test-alias-register.mjs`).
 const ALIAS_LOADER = new Set([
+  // Imports "@/lib/..." at module scope, so it cannot resolve without the
+  // alias register. It was wired into SUITES without this entry, so the runner
+  // launched it bare and it died at import with ERR_MODULE_NOT_FOUND — hiding
+  // every test after it (cave-ktvy0).
+  "src/lib/server/beads-delivery-source.test.ts",
   // onboarding diagnostics and core tools resolve shared server/API aliases.
   "src/lib/server/onboarding-diagnostics.test.ts",
   "src/lib/server/onboarding-core-tools.test.ts",
--- a/src/components/browser-polish.test.ts
+++ b/src/components/browser-polish.test.ts
@@ -224,9 +224,13 @@ assert.match(
   /import \{ deactivateAllNativeBrowserWebviews \} from "@\/lib\/native-browser-lifecycle"/,
   "Workspace uses the shared native browser cleanup helper",
 );
+// cave-x6rw normalized split payloads into WorkspacePaneRequest, collapsing the
+// old discriminated union (kind === "browser" | kind === "page" && mode ===
+// "browser") into a single pageId. Same contract: the browser counts as visible
+// from the primary mode OR any split pane (cave-ktvy0).
 assert.match(
   workspace,
-  /mode === "browser" \|\|[\s\S]{0,180}target\.kind === "browser" \|\| \(target\.kind === "page" && target\.mode === "browser"\)/,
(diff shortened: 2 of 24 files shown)


BunsDev:
Closing — superseded. `main` is green without it.

While this split was being built and verified, the repairs landed independently: **#4626** (the `beads-delivery-source` alias-loader registration, same content as `de209cf62` here) and **#4637** ("Restore the green frontend baseline"). A full `pnpm test:app` against current `main` now passes **1,209 test files with zero failures**, including the macOS `/var` → `/private/var` `automation-runner` artifact that had been failing locally on every branch tonight.

This PR existed for one reason — #4618 could not accumulate a completed CI run (six runs, zero completions, head moving every ~20 minutes against a ~30 minute run) while `main` stayed red behind it. That reason is gone.

Nothing is lost. Every commit here was cherry-picked from #4618 and remains on that branch.

**One durability note for whoever touches these next.** `main`'s version of the `initialTab` pin encodes the *new* literal:

```
/initialTab=\{mode === "familiar-work-queue" \|\| variant === "queue" \? "queue" : "tasks"\}/
```

This branch (and #4618) widened it instead:

```
/initialTab=\{[^}]*mode === "familiar-work-queue"[^}]*\? "queue" : "tasks"\}/
```

Both are green today. The literal form will break again the next time that expression gains a condition — which is exactly how tonight's seventeen-pin cascade started. The convention is written up in [`docs/source-text-pins.md`](docs/source-text-pins.md) (#4625): a source-text pin may assert a contract, never incidental syntax. Worth folding the widened form in next time that file is touched, rather than reopening this.

PR #4671, opened 2026-08-15
Exercise every React hook lint source glob

BunsDev:
## Summary

Extend the executable Rules of Hooks gate to prove ESLint rejects conditional hooks in every configured source glob: `ts` and `tsx` across `src/components`, `src/app`, and `src/lib`.

## Context

PR #4668 landed the underlying Rules of Hooks enforcement. PR #4672 then landed and expanded the original behavioral test from this PR while #4671 was waiting on CI. This follow-up preserves that stronger `main` version and closes the remaining runtime-coverage gap by exercising all six source glob variants.

## Changes

- share representative invalid TypeScript and TSX fixtures
- run each fixture through the real ESLint config in all three source areas
- retain the clean control and severity assertions from `main`

## Verification

- `node scripts/rules-of-hooks-gate.test.mjs`
- `node scripts/eslint/react-hooks-gate.test.mjs`
- `pnpm lint`
- `pnpm typecheck`
- `pnpm test:app` (1,213 files)
- `pnpm test:api` (379 files)
- `pnpm check:tests-wired` (1,673 files)

Bead: `cave-hmltt`

### DIFF
--- a/scripts/eslint/react-hooks-gate.test.mjs
+++ b/scripts/eslint/react-hooks-gate.test.mjs
@@ -34,42 +34,31 @@ async function hookErrorsFor(relativePath, source) {
 
 // A hook after an early return — the exact shape that crashed the Relations
 // graph (cave-qxq4l) with "Rendered more hooks than during the previous
-// render". One case per linted scope; the src/lib fixture is a custom hook in a
-// .ts file rather than a component, because .ts cannot carry JSX.
-const OFFENDERS = [
-  [
-    "src/components/__react-hooks-gate-fixture.tsx",
-    `import { useState } from "react";
-
-export function ConditionalHook({ enabled }: { enabled: boolean }) {
-  if (!enabled) return null;
-  const [value] = useState(0);
-  return <div>{value}</div>;
-}
-`,
-  ],
-  [
-    "src/lib/__react-hooks-gate-fixture.ts",
-    `import { useState } from "react";
+// render". One case per configured source glob proves both extensions are
+// enforced in every linted scope.
+const TS_OFFENDER = `import { useState } from "react";
 
 export function useConditional(enabled: boolean) {
   if (!enabled) return null;
   const [value] = useState(0);
   return value;
 }
-`,
-  ],
-  [
-    "src/app/__react-hooks-gate-fixture.tsx",
-    `import { useState } from "react";
+`;
+const TSX_OFFENDER = `import { useState } from "react";
 
-export default function Page({ enabled }: { enabled: boolean }) {
+export function ConditionalHook({ enabled }: { enabled: boolean }) {
   if (!enabled) return null;
   const [value] = useState(0);
   return <div>{value}</div>;
 }
-`,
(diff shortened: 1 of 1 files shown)


BunsDev:
⚠️ **cave-hmltt already landed** — flagging before this goes further.

It merged as #4668 (`7ed5d2f2c`) and the bead is closed. Apologies for the collision: I didn't see `chore/cave-hmltt-react-hooks-lint` until after mine had merged, or I'd have coordinated instead of duplicating.

We independently made the same calls — `eslint-plugin-react-hooks@7.1.1`, the same swap of the stub for the real plugin, and the same `useHarnessForReply` → `applyHarnessToReply` rename for the naming false positive. This PR now overlaps `main` on every file it touches (`eslint.config.mjs`, `package.json`, `pnpm-lock.yaml`, `scripts/run-tests.mjs`, `group-chat-view.tsx`, `harness-fix-actions.test.ts`), so expect conflicts on a rebase.

**One way what landed goes further.** `lint` now runs a single `lint:source` pass over six globs — `src/lib`, `src/app`, `src/components` (`.ts` and `.tsx`). That part is load-bearing, not taste: ESLint only lints the paths the CLI is handed, so widening the config's `files` block alone is inert. On this branch `lint` still runs `lint:design`, which passes only `src/components/**/*.tsx`, so `rules-of-hooks` would never reach `src/lib` — where `use-inline-slash-menus.ts` and `cave-familiar-images.ts` live. Measured 0 violations across the wider paths, so the bigger gate cost nothing.

**One way this PR is better, and worth re-landing.** `scripts/eslint/react-hooks-gate.test.mjs` runs ESLint and asserts the resulting severity is 2 — it proves the rule *fires*. What I merged (`scripts/rules-of-hooks-gate.test.mjs`) only pins config and script text: it catches a re-stub or an unhooked script, but not a subtler way of silencing the rule. Yours is the stronger check.

If you salvage one thing from this PR, salvage that test. A small PR adding it alongside the source-text pin needs no rebase of the rest and closes a real gap in what I merged.

Nothing here is at risk either way — `ac47d92be` is retained as `retention/chore-cave-hmltt-react-hooks-lint-a95515d5a`. I haven't touched the branch, worktree, or this PR's state; the call on closing it is yours.

BunsDev:
Superseded by #4668 and #4672, both now merged.

- #4668 landed the real plugin and a broader lint surface across `src/components`, `src/app`, and `src/lib`.
- #4672 explicitly salvaged this PR's behavioral ESLint gate, extended it to all six TS/TSX scope combinations, and added a clean control.

Together they are a strict superset of this branch, so merging #4671 would duplicate the dependency/config/callback changes. Closing without merge.

For completeness, this PR's parallel validation lanes all passed; the protected aggregate was cancelled only after E2E's dev server became unresponsive and 44 unrelated navigation tests cascaded into `ERR_ABORTED` until the 60-minute job timeout. Since the implementation is already landed through the superseding PRs, a rerun here would provide no delivery value.

PR #4763, opened 2026-08-20
docs: document model- and harness-level execution analytics

BunsDev:
## What

Adds `docs/execution-analytics.md` — a Living doc for the familiar **execution analytics** surface (`src/lib/server/familiar-execution-analytics.ts`, `/api/familiars/execution-analytics`, and the Familiar Analytics UI) — and indexes it in `docs/README.md`.

This came out of a review of the existing analytics docs and sections with a focus on **model-level and harness-level** analytics. The finding: that layer is fully implemented and **entirely undocumented**, so the update is a new doc rather than an edit to an existing one.

## Why it matters

Several behaviours here read as bugs from the UI but are deliberate (or structural), and nothing recorded which was which. The doc pins each one to the code:

- **Three model identities collapse to one slice key.** `requested`, `forwarded`, and `confirmed` are captured separately, but aggregation keys on a single resolved model — so model *substitution* is invisible in the aggregate.
- **`harnessVersion` is never recorded.** No writer populates it, so its coverage row is permanently 0% *by construction*, not because data is missing. Anyone reading that row today concludes the pipeline is broken.
- **There is no live capture path.** Every attempt is a transcript backfill, which is why `attemptNumber` is always 1.
- **Coverage is the honesty layer** — 15 declared fields against 6 actually emitted, `coverage.knownFields` read by nothing, and two dead UI labels (`firstOutput`, `quality`).
- **What the aggregate discards:** model controls are captured but never aggregated, there is no coven-level model/harness rollup, cache tokens are excluded from totals, and reads scale with lifetime history.

## Scope

Docs only. **No source or behaviour changes** — the diff is two files, `+261 / -0`.

## Verification

- `scripts/docs-index.test.mjs` — passes (50 documents, 63 links), so the new `docs/README.md` entry resolves.
- `src/lib/server/familiar-execution-analytics.test.ts` — passes 5/5 (unchanged; run to confirm the doc describes green code).
- `check-conflict-markers` — clean.
- Every claim in the doc was read back against the implementation in-tree rather than inferred from the UI.

## Follow-ups (not in this PR)

The doc names them; each is a candidate for its own change:

- Retire or populate the two dead coverage labels.
- Decide whether `harnessVersion` should be captured or its coverage row removed.
- Decide whether model substitution deserves its own slice dimension.

Bead: `cave-eptdn`


BunsDev:
Closing as superseded — both halves of this PR are already delivered elsewhere.

**The docs half has landed.** `docs/execution-analytics.md` is on `main` today, byte-identical to the version in `fe126ada3` (13,417 bytes, empty diff). Nothing in that commit is outstanding.

**The test half is carried, more strictly, by #4768.** What remains in the diff against the merge base is only `navigator-surface.test.ts`, repairing the same stale source-contract guards that PR #4757 invalidated. #4768 fixes them alongside two other breaks that deadlock the same required gate, and it pins the contract harder:

- This PR loosens the `setDependency(overlay, stepId, needs)` guard to a bare `setDependency(overlay, stepId, ` prefix — which also matches the pre-#4757 parent-writing form the assertion exists to forbid. #4768 instead pins the canonical `setDependency(overlay, stepId, null)` positively **and** adds a negative guard so the old form cannot come back unnoticed.
- #4768 also keeps the `body: JSON.stringify({…dependsOn` negative guard and pins the canonical write path (`overlayImportPlan`, `{ dependencies: [...] }`).

No work is lost by closing: the docs are on `main`, and the stricter test repair is in #4768.

For the record, this is the third independent session to hit the same break (also `cave-m448b` and `cave-2s8b2`). The underlying cause is that push runs to `main` skip the frontend validation matrix, so this rot accumulated invisibly until it blocked the whole PR queue at once.

PR #4767, opened 2026-08-20
test: restore main to green — three guards left stale by 2026-08-20 merges

BunsDev:
## What

`main` is red. Three independent test breaks landed on 2026-08-20, each one a
stale guard left behind by a merge that changed the code it pins. None of them
is a product regression — every underlying behaviour is intact — but together
they fail `pnpm test:app` and `pnpm test:api`, which reddens `PR checks` on
every open PR and masks real failures behind them.

## The three breaks

**1. `src/components/role-surfaces/navigator-surface.test.ts`** — stale since
#4757 (`5a09173153`), which rewrote `navigator-surface.tsx`:

- `announce(failure, "assertive")` is now ``announce(`${failure}${detail}`, "assertive")``
- `setDependency(overlay, stepId, needs)` is now `setDependency(overlay, stepId, null)` —
  dependencies became canonical on `Card.dependencies`, so the legacy overlay is
  only ever *cut*, never written
- the reset copy was rewritten

The test's leading comment also claimed "the board has no dependency column",
which #4757 made false (the source writes `{ dependencies: kept }`). Corrected
rather than left as a lie next to a passing assertion.

**2. `scripts/dev-app-teardown.test.mjs`** — stale since #4759 (`ae27407a91`),
which made readiness an *authenticated* probe: `dev-app-origin-health.mjs` now
sends `x-coven-cave-readiness-token` and accepts only a response carrying
`x-coven-cave-readiness: 1`, which `src/proxy.ts` stamps for a matching probe.
#4759 updated `dev-app.test.mjs` and `dev-app-origin-health.test.mjs` but not
this fixture, whose stub dev server still answered a bare `204`. Readiness could
therefore never be satisfied and the launcher timed out.

The stub now mirrors the `src/proxy.ts` contract instead of stamping the header
unconditionally, so the fixture still fails if the launcher ever stops minting
and exporting `COVEN_CAVE_DEV_PROBE_TOKEN` for the dev server it owns.

**3. `src/components/familiar-menu-bar.test.ts`** — stale since `387e9b1b5b`
("Reuse board resource and fix split pane sizing"), which moved the open-task
card read in `workspace.tsx` from `fetch("/api/board")` to
`readSurfaceResource("board:cards")`. The behavioural contract the guard exists
to protect is byte-identical — the same `.filter((c) => c.status !== "done")`
and `.map((c) => ({ familiarId: c.familiarId ?? null }))` still feed the Tasks
badge — only the transport moved. Re-anchored on the surface-resource read.

## Verification

- `navigator-surface.test.ts` — 23 pass / 0 fail
- `dev-app-teardown.test.mjs` — ok (and non-vacuously: it only goes green
  because the launcher really exports the token and the probe really sends it)
- `familiar-menu-bar.test.ts` — ok
- Full `pnpm test:app` / `pnpm test:api` on this branch, plus CI

## Scope

Test-side only. No product code is touched; every change re-points a guard at
what its own source now does, or corrects a comment that had become false.

Bead: `cave-2s8b2`

## Relationship to #4768

Another session opened #4768 against breaks 1 and 2 while this branch was in
flight. Its versions of both files were stronger than the ones here, so they
have been adopted verbatim (`a1fa67070d`) and this PR is now a strict superset.

Neither PR can go green on its own: the required `Frontend build` check
aggregates the whole validation matrix (`FRONTEND_VALIDATION_RESULT`), so app
tests must pass to merge. #4768 does not carry break 3, and this branch did not
carry its improvements — hence the consolidation rather than two PRs racing.

### DIFF
--- a/scripts/dev-app-teardown.test.mjs
+++ b/scripts/dev-app-teardown.test.mjs
@@ -83,8 +83,19 @@ echo "server_token $COVEN_CAVE_AUTH_TOKEN" >>"${pidFile}"
 node -e '
   const http = require("http");
   const fs = require("fs");
+  // Mirror src/proxy.ts: only an authenticated readiness probe earns the proof
+  // header. Stamping it unconditionally would keep this fixture passing even if
+  // the launcher stopped minting and exporting COVEN_CAVE_DEV_PROBE_TOKEN for
+  // the dev server it owns, which is the contract this test relies on.
+  const expected = (process.env.COVEN_CAVE_DEV_PROBE_TOKEN || "").trim();
   const server = http.createServer((request, response) => {
-    response.writeHead(204);
+    const url = new URL(request.url, "http://127.0.0.1");
+    const ready = expected.length > 0
+      && request.method === "GET"
+      && url.pathname === "/"
+      && url.searchParams.get("__devShellProbe") === "1"
+      && request.headers["x-coven-cave-readiness-token"] === expected;
+    response.writeHead(204, ready ? { "x-coven-cave-readiness": "1" } : {});
     response.end();
   });
   server.listen(${port}, "127.0.0.1", () => {
--- a/scripts/dev-app-teardown.test.mjs
+++ b/scripts/dev-app-teardown.test.mjs
@@ -83,19 +83,19 @@ echo "server_token $COVEN_CAVE_AUTH_TOKEN" >>"${pidFile}"
 node -e '
   const http = require("http");
   const fs = require("fs");
-  // Mirror src/proxy.ts: only an authenticated readiness probe earns the proof
-  // header. Stamping it unconditionally would keep this fixture passing even if
-  // the launcher stopped minting and exporting COVEN_CAVE_DEV_PROBE_TOKEN for
-  // the dev server it owns, which is the contract this test relies on.
-  const expected = (process.env.COVEN_CAVE_DEV_PROBE_TOKEN || "").trim();
   const server = http.createServer((request, response) => {
+    // Mirror the readiness contract in src/proxy.ts: only an authenticated
+    // dev-shell probe earns the proof header. Answering it unconditionally
+    // would let this test pass against a launcher that never sends a token.
     const url = new URL(request.url, "http://127.0.0.1");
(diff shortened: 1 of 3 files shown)


BunsDev:
Closing in favour of #4768, which carries everything here and more. No work is lost.

**The three repairs are the same, and two files are byte-identical.** Comparing this branch against `fix/cave-m448b-navigator-surface-tests`:

- `scripts/dev-app-teardown.test.mjs` — identical, zero diff.
- `src/components/role-surfaces/navigator-surface.test.ts` — identical, zero diff.
- `src/components/familiar-menu-bar.test.ts` — the only difference, and #4768's version is the stronger one.

On that third file, this branch asserts only the consumer side — that `workspace.tsx` reads `readSurfaceResource("board:cards")`. #4768 additionally reads `src/lib/surface-warmup-registry.ts` and pins `defineResource("board:cards", (signal) => json(signal, "/api/board")`. That matters: without it, the indirection could be repointed at a different endpoint and the consumer-side assertion would still pass, which is exactly how the original break got through. #4768's regex is also whitespace-tolerant, so a formatter cannot break it.

**#4768 also fixes four Playwright specs this branch does not touch.** Repairing the three unit contracts lets the e2e leg actually execute for the first time in a while, and it immediately fails on ten pre-existing stale specs (`cave-1cdzz`). Since the required `Frontend build` gate aggregates the whole validation matrix, a PR fixing only the unit contracts still cannot go green — so the e2e repairs have to land in the same PR. #4768 carries verified fixes for `canonical-memory.spec.ts`, both failing `chat-boot-landing.spec.ts` tests, and `dashboard-bento.spec.ts`.

**Your commits are retained.** The head of this branch is on `origin`, and the earlier tip is additionally archived at the pushed tag `archive/fix-cave-2s8b2-chart-room-stale-assertions-2026-08-20`.

Worth recording that four separate PRs (#4763, #4767, #4770 and #4768) independently attacked these same three breaks. That duplication is the predicted cost of `cave-ph4ga`: push runs to `main` skip the frontend validation matrix, so a break lands green, is attributed to nobody, and is then re-diagnosed from scratch by every session that trips over it.

PR #4769, opened 2026-08-21
Enforce local remote-tracking hygiene so Desktop's branch list stays honest

BunsDev:
GitHub Desktop renders every remote-tracking ref and every upstream it finds, so
foreign fork remotes, stray `refs/remotes/pull/*` heads left behind by a PR
checkout, and upstreams pointing at a branch the local branch is not a view of
all surface as branch-list noise. All of it is local-only state, but nothing
audited it — so it accumulated and got cleared by hand sweeps that re-derived
the rules each time.

## What lands

`scripts/remote-hygiene.mjs`, wired as `pnpm remotes:audit` / `remotes:audit:json`
/ `remotes:fix`. Five rules:

| rule | fix |
| --- | --- |
| exactly one remote, named `origin` | `git remote remove <fork>` |
| `fetch.prune = true` | set it |
| `fetch.pruneTags` never `true` | set false (tags are the retention store) |
| no tracking refs outside `refs/remotes/origin/` | delete — only when the tip is held elsewhere |
| no **bogus** upstream | `git branch --unset-upstream` |

It never touches the remote — no push, no branch deletion, no fetch. `origin`
branches other sessions own are reported for information only. A stray ref whose
tip is on **no other ref** is refused rather than deleted, and reported with the
archive-tag command instead.

## The load-bearing part: which upstreams count as bogus

- **Bogus → unset.** `branch.<X>.merge` naming a *different* branch — what
  `git worktree add -b X <path> origin/main` wrote until `cave-t57kr`. Renders
  "behind N" against a ref the branch is not a view of. Same for an upstream
  naming a remote that is not configured.
- **Accurate → left alone.** `branch.<X>.merge == refs/heads/X`. This is not
  noise, and `branch.<X>.remote` is one of three anti-resurrection signals
  `worktree-retention-push.mjs` reads (`cave-xjuup`) — the only one that
  survives a `fetch --prune`. Stripping it from a genuinely pushed branch makes
  a merged, server-deleted head read as "never pushed" and the hook re-creates
  it; that failure was measured at 9 of 36 remote branches.

`CLAUDE.md` previously said flatly that only `main` and the dolt branch should
have an upstream, with a bare `for-each-ref | grep -v ' -> $'` as the audit.
That listing flags every accurate self-tracking branch as a violation — **8 in
this checkout, all correct** — so following it would have traded branch-list
tidiness for resurrected merged heads. The section is rewritten around the
distinction.

## Verification

- `node --test scripts/remote-hygiene.test.mjs` — **10/10 pass**, including a
  regression test asserting `--fix` leaves an accurate self-tracking upstream
  intact, and one asserting an unretained stray ref survives `--fix`.
- `node scripts/check-tests-wired.mjs` — `✓ all 1745 test files wired into CI`.
- `pnpm exec eslint scripts/remote-hygiene.mjs scripts/remote-hygiene.test.mjs` — clean.
- Run against this checkout: clean — one remote, pruning on, no stray refs, no
  bogus upstreams; 12 self-tracking branches correctly not flagged.

## Already applied locally (not part of the diff)

`refs/remotes/pull/4753/head` — a stray fetched head from merged PR #4753 —
was deleted from the primary checkout; its tip was held by
`origin/ci/cave-7kix8-8-signed-promotion`, so nothing was lost.

Refs: cave-u426u


CompleteDotTech:
From #4785, which is blocked behind `main` and watching this PR as the closest thing to a repair: **one assertion still outstanding on `chat-view.tsx`.**

`Frontend validation (app tests)` on this head stops at `src/lib/session-debug.test.ts` after 800 files, expecting:

```
/<ChatToolActivityLayout[\s\S]*activity=\{otherTools\.length \? <ToolGroup tools=\{otherTools\} \/> : null\}[\s\S]*content=\{[\s\S]*<MessageBubble[\s\S]*editCards=\{/
```

That lines up with the other symptom `main` shows — `TS6133: 'ChatToolActivityLayout' is declared but its value is never read`. The merge appears to have dropped the **usage** while leaving the import, so the unused-import error and this assertion are one defect, not two.

For what it's worth, mapping the whole `chat-view.tsx` merge damage as seen from the outside, it looks like three separate drops with three separate source-scanning tests guarding them, and the two open repair PRs each cover a different subset:

| Dropped | Guarded by | Covered by |
| --- | --- | --- |
| `bubbleSegments` declaration | typecheck (`TS2304`) | #4784 |
| `ChatToolActivityLayout` usage | `session-debug.test.ts` + `TS6133` | neither yet |
| `!turn.pending` → `!pending` in the `<ArtifactComments>` guard | `artifact-comments.test.ts:54` | neither yet |

So neither PR alone turns `main` green today. Details on the third row are in a comment on #4784 — briefly, `chat-view.tsx:8865` defines `const pending = turn.role === "assistant" && Boolean(turn.pending)`, which is strictly narrower than the `turn.pending` that assertion still expects, so it is a behaviour change rather than a rename.

Deliberately not touching `chat-view.tsx` from #4785 — it belongs to whoever is repairing it, and two sessions duplicating a fix on one file is how conflicts get made. Flagging only because a red `main` is blocking the entire PR queue.

Issue #4777, opened 2026-08-21
[Chat v1 P7 Gate] Production v1 release accepted


CompleteDotTech:
# [Chat v1 P7 Gate] Production v1 release accepted

Close the program only after signed artifacts, public packages, compatibility canaries, OS acceptance, staged rollout, and rollback readiness are verified.

Owner repository: OpenCoven/coven-cave (program tracker)
Execution lane: program-coordination
Plan of record: OpenCoven/chat/docs/superpowers/plans/2026-08-15-phase-7-release-rollout.md

Deliverables:
- Signed artifact and published package inventory.
- Compatibility and acceptance evidence.
- Rollout decision and rollback readiness record.

Acceptance criteria:
- All Phase 7 beads and every prior phase gate are closed.
- Production v1 completion criteria are satisfied.
- Approved v1 non-goals remain out of scope.

## Bead metadata

- **Bead ID:** `cave-ilh1h`

- **Status:** open

- **Priority:** P0

- **Surface:** Shared

- **Authoritative owner:** Unassigned

- **Labels:** `area:gate`, `gate`, `needs-human`, `phase:7`, `program:chat-v1`, `release-blocker`, `repo:cross-repo`, `surface:shared`, `verification-required`

- **Parent:** None

- **Dependencies:** `cave-as76u`, `cave-563z7`, `cave-gcb0i`, `cave-mbekl`, `cave-udcn7`, `cave-j65ie`

## Source of truth

Beads is authoritative. This draft card is a GitHub Teamwork visibility mirror and should be reconciled from the Bead rather than edited as an independent task.

BunsDev:
Closing this accidental repository-issue conversion. The canonical task remains the Bead mirrored as a draft card on the Teamwork project; Beads remains the source of truth.

Issue #4776, opened 2026-08-21
[Chat v1 P7] Build signed cross-platform Chat packages and updater


CompleteDotTech:
# [Chat v1 P7] Build signed cross-platform Chat packages and updater

Configure production identity, CSP, protocol, icons, capabilities, platform installers, signatures, updater metadata, checksums, and OS-native CI.

Owner repository: OpenCoven/chat
Execution lane: chat-release
Plan of record: OpenCoven/chat/docs/superpowers/plans/2026-08-15-phase-7-release-rollout.md

Deliverables:
- Signed/notarized macOS artifacts, signed Windows MSI, and Linux AppImage.
- Updater signatures/metadata and package verification script.
- Minimum/latest/main Cave compatibility jobs.

Acceptance criteria:
- Artifacts install, launch, update, and uninstall cleanly.
- No shell or arbitrary-filesystem grant is packaged.
- Signing/updater credentials are available only to approved release jobs.

## Bead metadata

- **Bead ID:** `cave-gcb0i`

- **Status:** open

- **Priority:** P0

- **Surface:** Desktop

- **Authoritative owner:** Unassigned

- **Labels:** `area:desktop-release`, `needs-human`, `phase:7`, `program:chat-v1`, `release-blocker`, `repo:chat`, `surface:desktop`, `verification-required`

- **Parent:** None

- **Dependencies:** `cave-b6wsl`, `cave-j65ie`

## Source of truth

Beads is authoritative. This draft card is a GitHub Teamwork visibility mirror and should be reconciled from the Bead rather than edited as an independent task.

BunsDev:
Closing this accidental repository-issue conversion. The canonical task remains the Bead mirrored as a draft card on the Teamwork project; Beads remains the source of truth.

PR #4782, opened 2026-08-21
Gate rollout on a verified rollback target (cave-ilh1h)

CompleteDotTech:
Advances the Phase 7 gate in #4777 (bead `cave-ilh1h`) by building the one
deliverable it names that this repository owns no code for: **rollback
readiness**.

`scripts/release-promotion.mjs` proves the release being shipped was validated.
Nothing proved the release we would ship *back* to still works — and that is
the half a staged rollout actually depends on. Pausing a bad rollout is only a
remedy when the previous version is still installable and the updater can still
serve it.

## What lands

- `scripts/release-rollback-readiness.mjs` resolves the newest published,
  non-draft, non-prerelease release strictly below the one being shipped, and
  refuses unless it is a complete rollback target:
  - `.dmg`, `.msi`, and `.AppImage` installers, so any supported OS can reinstall
  - `SHA256SUMS`, so a rollback artifact can be checked before it is run
  - `latest.json` whose `version` matches the baseline, whose `platforms{}` is
    non-empty, and whose every entry carries a `url` and a `signature`
  - every `platforms{}` url still resolving to an asset on that release — a
    manifest pointing at deleted assets reads healthy and rolls nobody back
- A `rollback-readiness` job in `release.yml`, downstream of promotion
  authorization so a cut fails before it spends build time. `updater-manifest`
  now depends on it, so no install is moved forward on an unprovable rollback.
- The record is published as job outputs (`baseline-tag`, `baseline-version`,
  `baseline-url`, `rollback-platforms`) and a run summary, so the rollout
  decision has an artifact rather than an assumption.
- `docs/workflows/release-rollback-readiness.md`, linked from the release
  runbook in `branching.md`.

## Two details worth reviewing

- `updater-manifest` names the gate explicitly in its `if:`. That job runs under
  `!cancelled()`, which waives the default "every dependency succeeded" rule for
  *every* dependency — adding `rollback-readiness` to `needs` alone would have
  let a failed gate publish anyway. `scripts/release-rollback-readiness.test.mjs`
  pins both the dependency and the condition.
- `--allow-missing-baseline` waives only the case where no prior stable release
  exists at all (a repository's genuine first release). It is deliberately not
  wired into the workflow; every other shortfall stays fatal.

Verified against the live release history: `v0.3.7` and `v0.3.6` both carry the
full asset set and a well-formed `latest.json`, so this gate passes on the
current baseline rather than blocking the next cut.

## Validation

```
node --test scripts/release-rollback-readiness.test.mjs   # 10 pass
node --test scripts/release-promotion.test.mjs scripts/release-promotion-workflow.test.mjs scripts/ci-paths.test.mjs
node scripts/docs-index.test.mjs
node scripts/check-tests-wired.mjs   # new test wired; see caveat below
```

## ⚠️ Draft because `main` is red, not because this change is unfinished

`origin/main` (`b5f40e54d`) fails CI on its own, so the required `Frontend
build` context cannot go green for any PR branched from it. Reproduced locally
in a clean worktree at that commit, before any change here:

- `pnpm typecheck` — `src/components/chat-view.tsx` references an undefined
  `bubbleSegments` (lines 9001, 9051, 9069) and passes a `durationMs` prop that
  does not exist. Merge damage from `5771a5558`
  (`origin/feat/cave-20wrn-calm-streaming-chat`), which reached `main` directly
  rather than through a PR.
- `pnpm check:tests-wired` — 8 test files from that same merge are not wired
  into any suite (`chat-result-markers`, `chat-tool-verification`,
  `streaming-markdown-blocks`, `streaming-turn-view-model`,
  `use-stick-to-bottom`, `chat-assistant-output`, `streaming-chat-wiring`,
  `streaming-turn-response`).

The last four `CI push` runs on `main` failed the same way (typecheck, bundle,
app/API/mobile tests, test wiring, `Frontend build`). None of it is touched by
this branch — the repair belongs in its own PR, and this one should be marked
ready once `main` is green again.

Closes nothing on its own: #4777 is a gate that also requires its six
dependency beads to close.


CompleteDotTech:
The red `main` this PR is blocked behind is now partly addressed in #4784, which restores the declarations `5771a5558` dropped and wires the eight orphaned test files — 4 of the 7 failing jobs.

The remaining three (`app`, `mobile`, `api`) are source-text pins the streaming-chat feature regressed on its own, and they need that author's call rather than a unilateral fix. Nothing in either PR touches this one's scope; this stays draft until `Frontend build` can report green.

CompleteDotTech:
Reviewed at `1af65f397` and pushed four fixes; head is now `7858f383b`. The gate does what it claims — it is the only thing in this repository that proves the release you would roll *back* to is still installable, and `release-promotion.mjs` genuinely does not cover that. I would land it once `main` is green.

## Fixed

**1. The API token was sent to the asset download — `e0617229b`**

`requestJson` attached `Authorization: Bearer` to every request, and it makes two kinds: the `api.github.com` listing (needs it) and the baseline's `browser_download_url` (does not). That second url is `github.com/…/releases/download/…` and redirects to `objects.githubusercontent.com`.

I checked whether that actually leaks rather than assuming either way. Node 24's undici **does** strip `Authorization` across an origin change on redirect — verified locally on v24.18.0 with a two-origin redirect server, second hop saw `null` — so on the runner this workflow pins it was not a live leak. But the protection was the runtime's, not this script's: the token was still sent unnecessarily on the first hop, and a Node bump or a change of asset host would move the boundary silently. The token now goes only to requests whose origin matches the resolved API origin, pinned by a test that asserts both hosts are reached and only one sees the header.

**2. Manifest urls were matched by filename, so a stale manifest passed — `abb519069`**

This was the sharpest one. `manifestAssetName` took the last path segment and checked it against the baseline's asset-name set. Every release publishes the same filenames, and `generate-latest-json.mjs` writes `…/releases/download/<tag>/<name>` — so a `latest.json` left over from the *previous* cut differs from a correct one only in the `<tag>` segment, which the check could not see. The gate would report the baseline complete while its manifest pointed every updater at the previous release's binaries: precisely the "reads healthy, rolls nobody back" failure the header comment says it is guarding against.

Comparison is now origin + decoded path, which pins repository, tag and name at once and treats two encodings of the same url as equal. Refusals name the whole url now, since the filename no longer identifies what was wrong.

**3. A truncated release listing was returned silently — `b88e290f7`**

`listReleases` returned what it had when `MAX_API_PAGES` was exhausted with a full page still coming back. Yes, this needs >2000 releases to trigger — but the failure mode is a *pass*, not a miss, and that is what makes it worth closing. GitHub orders `/releases` by creation, not version, so the newest stable release below the rollout is not guaranteed to be near the front (a backported patch is created after the minor that supersedes it). A cut-short listing can therefore verify an older release and report the rollout ready. It now throws and names the constant to raise. The partner test also walks a genuine two-page listing — the old fixture returned everything on page one, so the loop's second iteration had never run.

**4. Nothing pinned that CI does not pass `--allow-missing-baseline` — `7858f383b`**

Confirmed by reading: the flag waives only `baseline === null`, artifact and manifest shortfalls stay fatal, and the workflow does not pass it. But the contract test asserted the job *runs* the script without asserting what it runs it *with*, so adding the waiver later would have been a one-word edit that every test passed. It now pins the exact command, the flag's absence anywhere in the job, and the two env values the gate reads.

## Checked, not a defect

- **`needs.rollback-readiness.result == 'success'` on `updater-manifest`** — correct, and correctly reasoned in the comment. `!cancelled()` does waive the needs-succeeded rule for every dependency, so naming the gate explicitly is required. The hyphen parses: `ci.yml` already ships `needs.frontend-validation.result` on every PR. `rollout is gated…` pins both the `needs` entry and the `if` clause.
- **Nothing else is downstream of the gate.** No job `needs: updater-manifest`, so there is no second job with the same `!cancelled()` hole. (`homebrew` uses the pattern but hangs off `checksums` — see below.)
- **Idiom** — `RELEASE_TAG: ${{ github.event.inputs.tag || github.ref_name }}` matches `authorize-release-promotion` and `source-version` exactly, and `tag` is `required: true` on dispatch, so the tag is always present. Importing `parseFinalTag` is the right seam: the alternative is a second copy of the tag grammar that could drift from the one authorization enforces.

## Needs a decision — not changed

**Gate placement.** The gate blocks `updater-manifest` but not `build` or `checksums`, so artifacts still get uploaded to a release with no verified rollback target. I think the current placement is right and would leave it: `build` uploading a `.dmg` moves nobody, and the artifacts are what a post-mortem needs. But `homebrew` (line 1816) is worth a second look — `brew upgrade` moves macOS installs forward exactly like the updater does, and it hangs off `checksums`, which is not gated. If the principle is "no install is moved forward on an unprovable rollback", Homebrew is a second forward path with no gate on it. Adding `rollback-readiness` to its `needs` would change release semantics, so I have not.

**`windows_diagnostics_only`.** The new job runs in that recovery mode and can fail the run, even though diagnostics mode publishes nothing (`updater-manifest` and `checksums` both exclude it). This is consistent with existing precedent — `daemon-package`'s npm gate and `source-version` also run unconditionally there — so I left it, but a recovery run failing on an unrelated baseline problem is a real cost if that path is ever used under pressure.

## Coordination

`scripts/run-tests.mjs` — heads up that `feat/cave-gcb0i-chat-signed-packages` (#4776) and #4784 **both** wire the same eight orphaned streaming-chat tests, at different offsets in `SUITES.app`. `git merge-tree` on those two branches merges cleanly and produces each path **twice**; nothing in `check-tests-wired.mjs` or `run-tests.mjs` guards against duplicates, so the eight would just run twice. This PR's own one-line addition to that file is unaffected, and its `release.yml` hunks are far from #4776's (line ~150 and ~1703 vs ~1732). Filed separately as the duplicate-guard gap.

CompleteDotTech:
One more, pushed after the review above: **`19759593b`** — head is now `19759593b`.

`writeCliResult` appends `rollback-platforms=<keys>` to `GITHUB_OUTPUT`, and those platform names are the only part of the readiness record sourced from a fetched document rather than a validated tag — `baseline-tag`, `baseline-version` and the rollout tag have all been through `parseFinalTag` by then. `GITHUB_OUTPUT` is newline-delimited `key=value`, so a key carrying a newline appends outputs of its own choosing, including overwriting `ready` and `baseline-tag` in the very record the rollout decision is read from.

Getting there requires already being able to replace a release asset, so this is defence in depth rather than a live hole. But the fix is a shape check against exactly the four keys `generate-latest-json.mjs` emits (`darwin-aarch64`, `darwin-x86_64`, `linux-x86_64`, `windows-x86_64` — all plain identifiers), so nothing legitimate is turned away, and it fails closed like every other shortfall here.

**Coverage note, so this is not read as more than it is:** `scripts/release-rollback-readiness.test.mjs` is wired at line 1614 of the `api` suite, and that suite halts at line 1458 on the inherited `chat-attention-persistence.test.ts` failure — so CI has never actually executed this gate's tests on this branch. All 15 pass locally (`node --test`, exit 0), along with `release-promotion` + `release-promotion-workflow` (59), `docs-index`, and `check-conflict-markers`. They will not have run under CI until `main` is green.

CompleteDotTech:
Correcting my own gate-placement note above, because I checked the Homebrew path afterwards and it does not support the suggestion I made.

Adding `rollback-readiness` to the `homebrew` job's `needs` would not actually gate anything. That job is only an *instant* notification — it dispatches `coven-cave-release` to `OpenCoven/homebrew-tap`, is `continue-on-error: true`, and swallows every failure with the same note: *"OpenCoven/homebrew-tap's scheduled update-cask workflow will pick up $TAG within 6 hours."* The tap updates itself from the published release on a 6-hourly schedule whether or not this job ever runs.

So Homebrew users get moved forward by the release *existing*, not by anything gated here. That means the honest version of the placement question is sharper than I put it:

- The updater is the only forward path this workflow can gate, and this PR gates it correctly.
- Every other forward path — Homebrew's schedule, and a direct download of a published `.dmg`/`.msi` — is opened by `build` and `checksums`, which are not gated. Gating those is the only way to actually enforce "no install is moved forward on an unprovable rollback", and it costs you the ability to publish artifacts for a post-mortem.
- Or accept that the guarantee is scoped to the updater, and say so in `docs/workflows/release-rollback-readiness.md` — right now the doc reads as a general promise ("before a release reaches anyone") that the placement does not deliver.

I would take the third option and leave the workflow alone; it is a one-paragraph doc change rather than a semantics change. But that is your call, and my earlier suggestion to add the gate to `homebrew` was wrong — it would have added a dependency edge that enforces nothing.

CompleteDotTech:
## Round-two adversarial review — 4 findings, all fixed in `671e8171`

Method: re-derived the gate's failure modes from scratch, then mutation-tested every assertion by injecting the regression into the production code and checking the suite actually goes red. **37 mutations injected; 5 survived the original suite; all 5 now caught (37/37).**

### 1. The draft / prerelease / unpublished exclusions were pinned by nothing — `671e8171`

`selectRollbackBaseline` skips drafts, prereleases and rows with no `published_at`. Removing any one of those three clauses left **every test green**.

The cause is the fixture, not the assertion: the one test that fed those rows used `v0.3.6` for all three against a `v0.3.7` stable baseline, so the newest stable release below `v0.3.8` was `v0.3.7` whether or not the filter held.

This is the most severe of the four. A draft release's assets are **not publicly downloadable** and its tag may not exist on the remote — certifying one as the rollback target promises a rollback that 404s for every user, and the gate would have reported `ready=true`.

Fixed by making the excluded rows *newer* than the answer (so admitting any one changes it), plus a dedicated test stating each rule on its own so a regression names what it broke.

### 2. A degraded GitHub reached the operator as a bare `SyntaxError` — `671e8171`

`requestJson` returned `response.json()` unwrapped. A 2xx carrying an edge error page, a truncated CDN body, or a captive-portal interstitial produced exactly:

```
release-rollback-readiness: Unexpected token '<', "<!DOCTYPE "... is not valid JSON
```

No request named, no release named, no retry advice — and not even a `RollbackReadinessError`, which the rest of the module treats as its contract. This directly contradicts the doc's promise that a failure which never reached a verdict *"says so and names the retry explicitly, because at 2am the two are otherwise indistinguishable"*. It reads as a bug in the gate rather than a GitHub that never answered, which is the misreading that advice exists to prevent.

A listing that comes back as a JSON *object* rather than an array (a degraded or rate-limited read) was the same condition and also lacked the advice. Both now carry request context and the retry advice, and both are asserted.

### 3. `runCli`'s waived branch was executed by no test — `671e8171`

`baseline-waived=true` and the `- Rollback target: none — waived as a first release` summary line were unreachable from the suite: hard-coding `baseline-waived` to `false` survived. The single run that reaches this branch is a repository's *first release*, so nobody is watching when it executes. Now covered end-to-end through `runCli`.

### 4. The blast radius was overclaimed in the docs and the job comment — `671e8171`

Both said the gate proves the rollback target *"before a release reaches anyone"*. Only `updater-manifest` depends on this job. Verified against the job graph:

| Still happens when the gate fails | Blocked when the gate fails |
| --- | --- |
| `build` creates the Release and uploads every installer | `updater-manifest` publishes `latest.json` |
| `checksums` publishes `SHA256SUMS` | in-app auto-update |
| `homebrew` bumps the tap, so `brew install --cask` serves it | |

The scoping is *correct* — the shortfall a red gate reports is in the **previous** release, so blocking the new artifacts would let a broken baseline block its own fix — but an operator reading the old wording would have believed a red gate was containment. Docs now carry the table and say plainly that stopping the new release takes a manual unpublish.

### Checked and found sound (no change)

- **The `stamp-release.test.mjs` pin widening (`9f35acc3`).** Mutation-tested through 9 variants: dropping `build`, dropping `source-version`, swapping their order, and inserting a new gate *before* `build` or *between* the two all still fail the pin; only entries appended *after* `source-version` are tolerated, which is the widening's purpose.
- **The merge resolution.** `scripts/run-tests.mjs` differs from the merge base by exactly the one added line. `check-tests-wired` passes (1758 files); the new test is listed once, in `api` only. The 9 within-suite duplicates in `app` are pre-existing on `main` and untouched here.
- **The gate against real repository state.** Dry-run with `RELEASE_TAG=v0.3.8` against `OpenCoven/coven-cave` exits 0 and resolves `v0.3.7` with all four platforms — so this does not brick the next cut. 193 releases = 2 pages, well inside `MAX_API_PAGES`.
- **`windows_diagnostics_only` recovery.** Not stranded: `updater-manifest` was already excluded for that input, and no other job depends on `rollback-readiness`.
- **`baseline-url` into `GITHUB_OUTPUT`.** The one value not shape-checked, but `html_url` is server-constructed from `owner/repo` plus a tag that already passed `parseFinalTag`, so it is not attacker-reachable the way a fetched manifest's platform keys were.

CompleteDotTech:
## CI state after `671e8171` — red, and the red is inherited from `main`

Run [32470724942](https://github.com/OpenCoven/coven-cave/actions/runs/32470724942): 8 pass, 4 fail, iOS skipped.

**Passing:** Select validation, lint, typecheck, API tests, mobile tests, protocol conformance, test wiring — including `Frontend validation (API tests)`, which is the suite this PR's tests actually live in.

**Failing:** `Frontend validation (app tests)`, `Frontend bundle`, `PR checks`, `Frontend build`.

Neither failure is reachable from this branch. Evidence:

1. **`main` fails identically, on its own runs.** [32470797261](https://github.com/OpenCoven/coven-cave/actions/runs/32470797261) (`d272ed0b1`, tip-minus-one) and [32470633689](https://github.com/OpenCoven/coven-cave/actions/runs/32470633689) (`dec4f93f9`) both fail the same four jobs with the same two root causes:
   - `AssertionError: crew trigger border uses exact 1px color-mix recipe` → `✗ FAILED: src/components/workspace-context-switcher.test.ts`
   - `✗ bundle-budget: root-layout CSS 616 KB exceeds budget 615 KB` (three Turbopack retries, all three the same)
2. **This branch touches no `src/` file at all** — the full changed set against `origin/main` is `.github/workflows/release.yml`, `docs/workflows/branching.md`, `docs/workflows/release-rollback-readiness.md`, `scripts/release-rollback-readiness.mjs`, `scripts/release-rollback-readiness.test.mjs`, `scripts/run-tests.mjs`, `scripts/stamp-release.test.mjs`.
3. **`src/components/workspace-context-switcher.test.ts` passes at this head** locally. `pull_request` CI builds the merge with `main`'s tip, which is how the regression arrives here.
4. **The previous head `9f35acc3` was 11/11 green** at 09:38 ([32468853591](https://github.com/OpenCoven/coven-cave/actions/runs/32468853591)). `#4791` landed on `main` at ~10:00; this push went out at ~10:01. Nothing in the diff between those two heads is in `src/`.

`main`'s current tip `8aaaf2f2` is docs-only, so the tree is still carrying both regressions — a rerun here will reproduce them until `main` is repaired. Not touching either from this branch: fixing an unrelated CSS budget and a chrome-refresh test regression inside a release-gate PR would bury both.

Everything this PR owns is verified green: `pnpm check:tests-wired` (1758 files), `release-rollback-readiness.test.mjs` (16/16), `stamp-release.test.mjs`, and every other test that parses `release.yml` (`release-macos-signing`, `ios-build-ci`, `ci-recovery-workflow`, `check-{grok,opencode}-registry-release`, `check-x-app-release`, `app-version`, `docs-index`).

Issue #4778, opened 2026-08-21
[Chat v1 P7] Packaging, compatibility, publishing, and rollout


CompleteDotTech:
# [Chat v1 P7] Packaging, compatibility, publishing, and rollout

Track Phase 7: Packaging, compatibility, publishing, and rollout.

Owner repository: OpenCoven/coven-cave (program tracker)
Execution lane: program-coordination
Plan of record: OpenCoven/chat/docs/superpowers/plans/2026-08-15-phase-7-release-rollout.md

Deliverables:
- Repository-owned implementation beads remain independently claimable.
- Cross-repository conformance evidence is attached to the phase gate.

Acceptance criteria:
- All Phase 7 implementation and verification beads are closed.
- The Phase 7 gate records passing commands and artifacts.

## Bead metadata

- **Bead ID:** `cave-j65ie`

- **Status:** open

- **Priority:** P1

- **Surface:** Shared

- **Authoritative owner:** Unassigned

- **Labels:** `area:program`, `phase:7`, `program:chat-v1`, `repo:cross-repo`, `surface:shared`, `tracking`, `verification-required`

- **Parent:** None

- **Dependencies:** `cave-udcn7`, `cave-ilh1h`, `cave-mbekl`, `cave-563z7`, `cave-as76u`, `cave-k0aqq`, `cave-gcb0i`

## Source of truth

Beads is authoritative. This draft card is a GitHub Teamwork visibility mirror and should be reconciled from the Bead rather than edited as an independent task.

BunsDev:
Closing this accidental repository-issue conversion. The canonical task remains the Bead mirrored as a draft card on the Teamwork project; Beads remains the source of truth.

PR #4785, opened 2026-08-21
feat(client-v1): serve release compatibility metadata on a health route

CompleteDotTech:
Chat v1 Phase 7, Task 1 — the Cave half of the packaging/compatibility work tracked by #4778 (bead `cave-j65ie`, plan of record `OpenCoven/chat:docs/superpowers/plans/2026-08-15-phase-7-release-rollout.md`). Owned locally by bead `cave-0wg`.

## The gap

Chat, the SDK, and the CLI have to know whether this Cave will talk to them **before** they pair. Today they cannot. `src/lib/server/client-v1/contract.ts` declares the API version, minimum client version and capability set, but nothing serves it — `git grep client-v1` across `src` and `scripts` returns only the module, its exporter, its CI path class, and tests. The module says so itself: *"Phase 0 fixture governance stays foundation-only … no route DTOs."*

## What this adds

`GET /api/client/v1/health`, returning the six facts a client needs. Three ride the existing envelope — `apiVersion`, `minimumClientVersion`, `capabilities` — rather than being repeated in `data`, so one response can never carry two different answers to the same question. The other three are new:

| Field | Source | Why |
| --- | --- | --- |
| `releaseVersion` | `APP_VERSION` | lets a compatibility matrix name the build it tested |
| `instanceId` | persisted random UUID | clients cache pairings against a Cave; this separates "restarted" from "different Cave, same port" |
| `pairingRequired` | contract constant | there is no unpaired mode to report, so adding one must change the contract in review |

`instanceId` identifies an installation, never a person or machine — a plain UUID with no hostname, username, or hardware value mixed in. It is deliberately public on an unauthenticated endpoint, so a test asserts the response body is exactly those three keys and contains no filesystem path.

## Fixture placeholders are deliberate

`contract-fixture.json` keeps `0.0.0` and a zero UUID rather than the running values. The fixture exists to prove the contract *shape* did not change; stamping the real version in would make every release rewrite it and the diff stop meaning anything. `keeps the release version out of the contract fixture` enforces that.

## Smoke scope — read this before reviewing the script

`scripts/client-v1-release-smoke.mjs` probes a running release build, because that is where release-only breakage lives: a route dropped from the bundle, an env-dependent field that is empty under `next build`, a version that never got stamped. Its assertions are pure and exported, so they are unit-tested without a server, and it accumulates failures rather than stopping at the first — one rebuild per failure is too slow a loop.

The plan also specifies a **pair / read / revoke / expect-401** leg. It is deliberately **not** here: the routes it drives do not exist. On this commit `src/app/api/client/v1/` contains only `health`, and there is no client-v1 pairing route at all (the only pairing code on `main` is the mobile surface). Asserting against routes that 404 would make the smoke fail for a reason unrelated to the release. That leg is tracked on `cave-0wg` and lands with the pairing routes.

## Verification

Run in `.worktrees/cave-0wg-client-v1-release-compat`:

- `node scripts/export-client-v1-contract.mjs --check` — exit 0, committed bytes match
- `contract.test.ts` — 23 passed (4 new)
- `route.test.ts` — 4 passed
- `instance-id.test.ts` — 4 passed, 1 skipped (needs POSIX permissions)
- `client-v1-release-smoke.test.mjs` — 9 passed
- `ci-paths.test.mjs` — 9 passed; `export-client-v1-contract.test.mjs` — 6 passed
- `pnpm lint` — clean
- `pnpm typecheck` — **zero errors in any file this PR touches**

---

## ⚠️ CI status: one real finding (fixed), the rest inherited from a red `main`

**CI caught a genuine problem in this branch, and `2f58992b3` fixes it.** `api-contracts.test.ts` asserted that *no* `/api/client/v1` route may exist — a Phase 0 gate from when the client v1 contract was a module nothing served — and it separately requires every route to hold a contract entry. The first push tripped both.

The gate is now **narrowed rather than deleted**: it pins the exact allowed list, so a second, unreviewed client v1 route still fails until someone adds it here on purpose. That keeps what the gate was actually protecting against. The client v1 response builders are also inlined into `effectiveRouteSource`, the same way the OAuth and onboarding routes already are — client v1 routes never construct a `Response` by hand, so without the inline the suite reads `route.ts`, finds no `Response.json`, and concludes the route returns nothing. 287 route contracts pass.

Worth recording why local runs missed it: `pnpm test:api` aborts on this Windows box at `scripts/worktree-guard.test.mjs` — a worktree-registration failure that does not reproduce on Linux CI — only 10 files in, while `api-contracts` sits at file 64 of 390. The targeted suites all passed, which is exactly how a whole-suite gate slips through.

Everything below is a different matter, and is inherited.

The base commit `b5f40e54d` **is `main`, and `main`'s own CI run fails** ([run 32454161667](https://github.com/OpenCoven/coven-cave/actions/runs/32454161667)). An incomplete merge landed there — the app-tests log still shows `error: could not apply 826167b... branch side` — and this branch inherits the wreckage. Every failing check traces to it, and each already has a fix in flight:
Every failing check was matched against `main`'s own run at the same commit. Where both fail inside a test suite, the file **and the count of files that passed first** are identical — that is the tightest evidence available that this branch reaches main's failure point and no earlier:

| Failing check | This branch stops at | `main` stops at | Fixed by |
| --- | --- | --- | --- |
| typecheck, PR checks | `bubbleSegments` undeclared in `chat-view.tsx` | same | #4784 |
| Frontend bundle | `'ChatToolActivityLayout' is declared but its value is never read` | same | #4784 |
| mobile tests | `chat-view-mobile-command-center`, scroll-bottom markup | same | #4784 |
| app tests | `artifact-comments.test.ts` — **428 passed before it** | `artifact-comments.test.ts` — **428 passed before it** | #4784 |
| API tests | `chat-attention-persistence.test.ts` — 157 passed before it | same file — 154 passed before it | #4784 |
| test wiring | eight unwired streaming-chat files, **none of them mine** | same eight | #4784 |
| iOS build | `AppModel.swift:1730`, missing `:` label | pre-existing; `main` skips the job | #4769 |
| Frontend build | requires the validation matrix above | — | consequence of the rest |

The API-tests row is the one number that *should* differ: 157 versus 154 is exactly the three suites this PR adds, all passing before it reaches main's wall.

Already green here: **lint**, **protocol conformance** (the client-v1 CI class this PR's work belongs to), and **Select validation**.

Two notes on that table:

**The iOS failure is pre-existing but only this PR surfaces it.** `scripts/ci-paths.mjs` is in `IOS_PATH`, so touching it runs the iOS build that every sibling PR skips. The one-character fix is already in #4769, so it is deliberately **not** duplicated here — landing it twice is how two sessions produce a conflict over one typo. Verified against the declaration at `AppModel.swift:1667` and the two correct call sites at 1679 and 1725.

**Why not drop the `ci-paths.mjs` change to dodge the iOS build?** Because that check is not wrong — it is reporting a real broken build. Removing the line would also leave `scripts/client-v1-release-smoke.mjs` ungated by the CI class that owns it, which is a coverage regression traded for a green square.

This stays in **draft** until #4784 and #4769 land on `main`; then it rebases and re-runs. `mergeable` is already `MERGEABLE` — no conflicts today — but expect to re-check `scripts/run-tests.mjs` after #4784 merges, since it edits that same file.

Refs #4778


CompleteDotTech:
### CI round 2 — a second real finding, also fixed

`f196762aa` fixes the one remaining failure that belonged to this branch.

`src/lib/server/client-v1/instance-id.test.ts` and `src/app/api/client/v1/health/route.test.ts` both died in CI with `Cannot find package '@/lib'` while passing locally. Adding a test to a `SUITES` array in `scripts/run-tests.mjs` is only half the wiring: the runner passes `--import ./scripts/test-alias-register.mjs` **per file**, to files listed in `ALIAS_LOADER`. `instance-id.ts` reaches `@/lib/coven-paths`, and the route reaches that plus `@/lib/app-version` and the client v1 response builders, so both belong in the set.

Local runs hid it exactly because I had typed the loader flag on the command line by hand — which is the difference `ALIAS_LOADER` exists to record. Re-verified by generating each file's argv from the runner itself rather than writing a command:

```bash
node -e "import('./scripts/run-tests.mjs').then(m=>console.log(m.nodeArgsFor('<file>').join(' ')))"
```

and running exactly that: contract **23 pass**, instance-id **4 pass**, health route **4 pass**, release smoke **9 pass**, api-contracts **287 route contracts pass** — all with zero manual flags.

One further check worth naming, since unit tests would not have caught it: the smoke's assertions were run against the **real route's output** rather than only a synthetic envelope, by invoking `GET()` and feeding the result to `checkHealthEnvelope` / `checkInstanceStability`. Real body `{"instanceId":"…","pairingRequired":true,"releaseVersion":"0.3.8"}`, zero failures — so the checker and the route genuinely agree on the shape.

The remaining red checks are still the inherited ones documented in the description (#4784 for the `chat-view.tsx` merge damage and the eight unwired streaming tests, #4769 for `AppModel.swift:1730`). Both of those PRs are themselves still failing checks, so this stays in draft until `main` is repaired.

CompleteDotTech:
### Round 3 — both of this branch's own defects are now confirmed fixed in CI

The API tests job on `f196762aa` runs **157 files** before failing, versus 64 when `api-contracts` was tripping. Everything this PR adds passed on Linux:

```
api-contracts.test.ts: 287 route contracts passed
✔ mints an instance id once and reuses it
✔ serves the six compatibility fields a client needs before pairing
✔ keeps the instance id stable across requests
client-v1-release-smoke contract: ok
```

Plus **Frontend validation (lint)** and **Frontend validation (protocol conformance)** — the client v1 CI class this work belongs to — both green.

**The remaining API failure is byte-for-byte main's own.** This branch stops at `src/app/api/chat/send/chat-attention-persistence.test.ts` (157 passed before it); `main`'s run [32454161667](https://github.com/OpenCoven/coven-cave/actions/runs/32454161667) stops at the *same file* (154 passed before it). The three-file delta is exactly the suites this PR adds.

So the inventory is now settled — nothing red here belongs to this branch:

| Check | Owner |
| --- | --- |
| API tests | `chat-attention-persistence`, same file main fails on |
| typecheck, PR checks, Frontend bundle, mobile tests | `chat-view.tsx` merge damage → #4784 |
| test wiring | the eight unwired streaming-chat files, none of them mine → #4784 |
| iOS build | `AppModel.swift:1730` → #4769 |

Test-merged clean against both `main` and #4784's head (`75136cc2d`) — zero conflict markers — so the rebase after they land should be uneventful. Staying in draft until then.

CompleteDotTech:
## ⚠️ Head-on collision with `origin/phase1a/cave-pairing-authority` — needs a human decision

Surfaced by an automated review pass. Flagging before this branch goes any further, because no amount of work *on this branch* resolves it.

`origin/phase1a/cave-pairing-authority` (`287497dd3`, 11 commits, Val Alexander, pushed 2026-08-20) **has no open PR** and builds the same surface this one does. A test-merge of the two heads reports **11 conflict markers across all 10 shared files** — which is every file this PR touches except the four it adds alone:

```
scripts/ci-paths.mjs                              src/lib/server/client-v1/contract.ts
scripts/ci-paths.test.mjs                         src/lib/server/client-v1/contract.test.ts
scripts/run-tests.mjs                             src/lib/server/client-v1/contract-fixture.json
src/app/api/api-contracts.test.ts                 src/lib/server/client-v1/contract-fixture.sha256
src/app/api/client/v1/health/route.ts             src/app/api/client/v1/health/route.test.ts
```

### The two branches disagree about what health *is*

`phase1a`'s version of the identical file:

```ts
export async function GET(): Promise<Response> {
  return clientV1SuccessResponse({ status: "ok" });
}
```

That is a liveness probe. This PR's is a compatibility handshake carrying `releaseVersion`, `instanceId`, and `pairingRequired` — and a `grep` for those three fields against `phase1a`'s route returns **0**, so it does not satisfy Phase 7 Task 1's acceptance criteria, which name them explicitly.

They are not rival attempts at one thing; they are two different phases that happen to land on the same path. `phase1a` is the pairing authority (admin credentials, pairing requests, exchange — a real route table). This PR is the release-compatibility metadata. **Both are wanted. Neither subsumes the other.**

### Why this is worth resolving before either lands

`phase1a` also resolves two things left open here, which is the strongest argument for reconciling rather than racing:

- **The capabilities over-promise.** This PR publishes `capabilities` with nine entries while `src/app/api/client/v1/` contains only `health`, so a client that branches on that list — the reason a capability list exists — calls `/api/client/v1/conversations` and gets a 404. `phase1a` actually ships routes, so the list stops being fiction.
- **The smoke's missing leg.** `scripts/client-v1-release-smoke.mjs` documents that pair → read → revoke → 401 is unimplemented because the routes do not exist. On `phase1a` they do.

It also narrows the same Phase 0 gate in `api-contracts.test.ts` to an 8-route allow-list using the *identical* construct this PR uses for a 1-route list — same author, same intent, two independent implementations. And it repairs `usesJsonResponse` by widening its regex where this PR inlines `responses.ts` into `effectiveRouteSource`. Both work; they cannot both land unreconciled.

### Suggested reconciliation, for whoever decides

Land `phase1a` first, then rebase this PR onto it and keep **its** route table with **this PR's** health body — that ordering leaves exactly one health route, a capability list that is true, and a smoke that can finally grow its pairing leg. The reverse order works too but reconciles the same ten files under more pressure.

This is the orphaned-work pattern `docs/multi-session-coordination.md` describes: both branches build clean, both pass their own tests, and nothing in git warns you until the second merge. Holding this PR in draft until someone rules on the ordering.

CompleteDotTech:
## Evidence against this PR's health design, surfaced by the review loop

Recording this because it argues against my own change and belongs in front of whoever rules on `cave-7yo`.

A review pass traced the Phase 0 gate to #4717 and found that its author **deliberately removed a route-shaped `health` DTO** while building the contract. From that commit:

> Replace the broad, route-shaped success DTOs (**health**, credential, familiar, project, conversation list/detail, stream event) with generic Phase 0 foundation primitives: a shared status record type … since Phase 0 does not define Phase 1-4 route response DTOs.

`ClientV1StatusRecord` was introduced *as the replacement for* the health DTO. And `phase1a/cave-pairing-authority` — same author — ships exactly that:

```ts
return clientV1SuccessResponse({ status: "ok" });
```

So `phase1a`'s health is not a stub someone left behind. It is the author's intended design, using the primitive built to replace the shape this PR reintroduces.

**The honest tension.** This PR adds `ClientV1Health` — a bespoke, route-shaped success DTO for `/health`, which is the category #4717 removed on purpose. Against that:

- The removal is scoped in its own words to *"Phase 0 does not define Phase **1-4** route response DTOs."* This is Phase **7**, and Phase 7 Task 1 names the six fields explicitly — `apiVersion`, `minimumClientVersion`, capabilities, instance ID, pairing requirement, release version. So the DTO is defined by a plan of record, not guessed, which is the thing #4717 was guarding against.
- A `{status:"ok"}` health cannot satisfy those acceptance criteria. A client cannot learn it is too old from a liveness probe.

Both designs are defensible and they cannot both own `/api/client/v1/health`. My read is that Phase 7's requirement is the later and more specific instruction and should win on the *body*, while `phase1a` owns the route table around it — but I am the author of one side, so I am not the right one to call it.

Two things I'd ask a reviewer to weigh that I cannot settle:

1. Should the compatibility metadata live on `/health` at all, or on a sibling (`/version`, `/compatibility`) — leaving `health` the liveness probe #4717 designed and removing the collision entirely?
2. If it stays on `/health`, `ClientV1Health` should probably be justified in the contract as a Phase-7-defined DTO rather than appearing as the kind of shape Phase 0 deleted.

Tracked on `cave-7yo`. Nothing here changes the code on this branch; it stays in draft.

CompleteDotTech:
## The plan of record settles the merge order — and corrects my earlier suggestion

Following up on the `phase1a` collision and the #4717 design evidence above. I went to the Phase 7 plan itself, and its Task 1 file list is decisive in a way none of the previous discussion noticed. It distinguishes *Modify* from *Create* deliberately:

```
- Modify: src/lib/server/client-v1/contract.ts
- Modify: src/app/api/client/v1/health/route.ts        ← Modify
- Modify: src/app/api/client/v1/health/route.test.ts   ← Modify
- Create: scripts/client-v1-release-smoke.mjs          ← Create
- Create: scripts/client-v1-release-smoke.test.mjs     ← Create
```

**Task 1 was written expecting the health route to already exist.** It is a *modify* step. This PR created it, which means it did a predecessor's work as well as its own — and that predecessor is `phase1a/cave-pairing-authority`, which creates exactly this file returning `clientV1SuccessResponse({ status: "ok" })`.

Three things follow, and they resolve most of what was open:

**1. The merge order is not a judgement call.** `phase1a` lands first; this PR then *modifies* its health route to carry the compatibility metadata. That is the sequence the plan describes, and it matches the #4717 evidence: `{status:"ok"}` on `ClientV1StatusRecord` is the Phase 0 author's health, and Phase 7 is the phase that extends it.

**2. My sibling-route suggestion was wrong — please disregard it.** I proposed moving the metadata to `/version` or `/compatibility` to dissolve the collision. The plan explicitly puts these fields on `/health`, so that would have traded a merge conflict for a deviation from the plan of record. Withdrawing it.

**3. It explains the smoke's missing leg.** Task 1 Step 1 also requires the smoke to *"pair a synthetic client, read one resource, revoke it, and receive 401 on the next read."* This PR documents that leg as unimplementable because the pairing routes do not exist. Under the intended ordering they do — `phase1a` ships them. The gap is not a scoping decision after all; it is a symptom of building Task 1 before its predecessor landed.

**What this PR should probably become**, for whoever picks it up: rebase onto `phase1a` once that has a PR and merges, reduce the health-route change from a create to the modify the plan calls for, and grow the smoke's pairing leg against the routes that then exist. The contract, fixture, instance-id, and release-smoke work all survive that rebase unchanged — it is the health route file itself, and `api-contracts.test.ts`'s allow-list, that fold into `phase1a`'s.

Still draft. `cave-7yo` updated with this.

CompleteDotTech:
## Review loop converged — five rounds, 14 defects fixed

Head `804ef0959`, 17 commits. Each round adversarially reviewed the round before it, because round 2 found that round 1's fix carried a regression and that turned out to be the rule rather than the exception.

| Round | Found | Notable |
| --- | --- | --- |
| 1 | 6 | sync disk read on every unauthenticated request (340 µs, 92% of the handler); two doc comments describing behaviour the code did not have |
| 2 | 3 | **a regression inside round 1's fix** — memoising removed the only thing that ever retried the store write |
| 3 | 3 | **round 2's hardening was a fail-open** — an equality check against an unvalidated expectation, so `undefined === undefined` passed |
| 4 | 3 | **round 3 repeated that defect one level up** — refusing the whole envelope discarded five checks that never consult the expectation (7 failures → 1) |
| 5 | 1 | `unusable` Set proved correct across all 8 completeness states; only a CLI-typo diagnostic remained |

The severity decays cleanly — shipped-behaviour bugs → a fail-open → incompleteness in a correct fix → an unhelpful error string on operator input error — and round 5 cleared all three of round 4's commits on the merits. That is why it stops here.

**Three findings are worth a reviewer's attention specifically**, because each is a case where something looked like protection and was not:

- `instance-id.ts` promised a stable per-process id on an unwritable disk; it minted a fresh UUID **per request**, so a client would re-pair on every call — the exact failure `instanceId` exists to prevent. The test guarding it chmodded `<home>/cave` while the store lives at `<home>/client-v1-instance.json`, so it passed for the wrong reason, and being POSIX-only it was skipped on the authoring machine.
- The race comment promised "the loser adopts the winner's file"; a plain `writeFileSync` did the opposite. Memoisation made it worse by removing the convergence that had hidden it, so the two changes had to ship together.
- The smoke accepted any non-empty `minimumClientVersion` — the field deciding whether a client may pair at all — so a build shipping a lower one would admit every too-old client while printing `ok`.

**Verification at `804ef0959`** (all via the runner's own `nodeArgsFor` argv, never a hand-written command): contract **23 pass** · instance-id **10 pass** · health route **4 pass** · release smoke **18 pass** · api-contracts **287 route contracts passed** · `export-client-v1-contract.mjs --check` exit 0 · lint clean · typecheck reports only the inherited `chat-view.tsx` errors · 0 conflict markers against `main`.

**Still blocked, and not by this branch.** Every red check is inherited from a `main` that went red across six direct-to-main merges (`b4c737dc1..b5f40e54d`); two of them match main's own run down to the identical count of files passing first. And per the comment above, the plan's Task 1 is a *Modify* step on `health`, so the real next move is `phase1a` opening a PR and landing first — this PR then reduces to the modify the plan describes.

CompleteDotTech:
## ✅ Green — all twelve checks pass at `461fc00c1`

`main` was repaired by #4769 (`b9856a7ed`), which turned out to be the broad fix despite its "remote hygiene" title: `chat-view.tsx`, `AppModel.swift`, `session-debug.test.ts`, `chat-attention-persistence.test.ts`, and the test wiring. Merged it in — zero conflicts, and a merge rather than a rebase to avoid a force-push in a shared checkout.

```
pass  Frontend build                         pass  PR checks
pass  Frontend bundle                        pass  Select validation
pass  Frontend validation (API tests)        pass  iOS build
pass  Frontend validation (app tests)        pass  Frontend validation (lint)
pass  Frontend validation (mobile tests)     pass  Frontend validation (typecheck)
pass  Frontend validation (protocol conformance)
pass  Frontend validation (test wiring)
```

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

Every one of the nine previously-failing checks was inherited, and all nine cleared without a line of this branch's code changing — including `iOS build`, the one this PR uniquely triggers by touching `ci-paths.mjs`, which needed #4769's `AppModel.swift` fix. The 17 review-loop commits went in on top and nothing regressed.

**Correcting myself:** my earlier comments on #4784 and #4769 mapped the failures to the wrong owners and claimed neither PR would fix `main` alone. #4769 fixed all of it on its own. The individual diagnoses in those comments hold; the ownership table did not, because I inferred it from titles and a glance at changed files rather than from what each PR actually repaired.

**Deliberately still a draft.** Per the plan-of-record comment above, Task 1 is a *Modify* step on `src/app/api/client/v1/health/route.ts` — it expected the route to exist. `phase1a/cave-pairing-authority` is what creates it, and still has no PR. Marking this ready would invite a merge that lands the health route in the wrong order and forces `phase1a` to reconcile ten files under pressure. Tracked as `cave-7yo`; green is necessary here, not sufficient.

CompleteDotTech:
**Marked ready for review** at the owner's direction, superseding the "deliberately still a draft" note above.

State: `461fc00c1`, **12/12 checks pass**, `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. `main` has since advanced 15 commits (now `f4f7e45de`) with **zero conflicts** against this branch; protection is `strict: false`, so being behind is not a merge blocker and the green run stands.

One caveat carried forward from the plan-of-record comment, unchanged by this status flip and worth a reviewer's eye before merge: Phase 7 Task 1 lists `src/app/api/client/v1/health/route.ts` as **Modify**, not Create. `phase1a/cave-pairing-authority` is what creates it and **still has no PR open**. Merging this first means `phase1a` reconciles ten files afterwards rather than this branch reducing to the modify the plan describes.

That is a sequencing judgement, not a defect — the code here is green and independently verified. Tracked as `cave-7yo`.

Issue #4775, opened 2026-08-21
[Chat v1 P5] Encrypted replaceable offline read cache


CompleteDotTech:
# [Chat v1 P5] Encrypted replaceable offline read cache

Implement an AES-GCM, keychain-derived, versioned, instance-scoped read cache with atomic replacement, caps, migration, and explicit corruption diagnostics.

Owner repository: OpenCoven/chat
Execution lane: chat-native-frontend
Plan of record: OpenCoven/chat/docs/superpowers/plans/2026-08-15-phase-5-native-lifecycle-offline-sdk-tooling.md

Deliverables:
- Encrypted cache envelope and Rust read/write/clear commands.
- Instance ID, revision, timestamp, schema version, and size bounds.
- Explicit purge/reset path for incompatible or undecryptable entries.

Acceptance criteria:
- Offline mode is read-only and clearly labeled.
- Secrets and attachment bytes are absent from cache files.
- A different Cave instance never receives another instance's cache as current data.

## Bead metadata

- **Bead ID:** `cave-f1k8n`

- **Status:** open

- **Priority:** P1

- **Surface:** Desktop

- **Authoritative owner:** Unassigned

- **Labels:** `area:offline-cache`, `phase:5`, `program:chat-v1`, `repo:chat`, `surface:desktop`, `verification-required`

- **Parent:** None

- **Dependencies:** `cave-gylsl`, `cave-0567z`

## Source of truth

Beads is authoritative. This draft card is a GitHub Teamwork visibility mirror and should be reconciled from the Bead rather than edited as an independent task.

BunsDev:
Closing this accidental repository-issue conversion. The canonical task remains the Bead mirrored as a draft card on the Teamwork project; Beads remains the source of truth.

PR #4786, opened 2026-08-21
Encrypted replaceable offline read cache (cave-f1k8n)

CompleteDotTech:
Closes the cache half of #4775 (bead `cave-f1k8n`, Chat v1 P5).

## Why

Chat's only cache is in memory — `src/lib/conversation-cache.ts` keeps the last few payloads so switching threads doesn't flash a skeleton, and it dies with the window. A Cave whose daemon is down therefore shows an empty room rather than the history it loaded five minutes ago. This adds the durable tier underneath it.

## What the format guarantees

`src-tauri/src/offline_cache.rs` writes one AES-256-GCM envelope per entry:

```
magic          8 bytes   b"CVOFFL01"
header_len     4 bytes   u32
header         N bytes   JSON, authenticated (AAD) but not encrypted
ciphertext     rest      AES-256-GCM(payload)
```

- **The key never touches disk.** HKDF-SHA256 over a 32-byte root secret held in the OS keychain, salted with the instance id and bound by the hashed `scope`/`key` pair. Moving a file to another name, scope, or instance directory makes it undecryptable rather than merely mislabelled.
- **Nothing identifying is in plaintext.** Scope and key are hashed before they reach the filesystem, so a file discloses neither its contents nor its subject. A test asserts the stored bytes contain neither the payload, the key, nor the scope.
- **A different instance is never served as current data.** The instance id is in the authenticated header *and* is the HKDF salt, so a foreign entry fails both checks.
- **Replacement is atomic.** Stage to a unique sibling, then rename. A torn write leaves the previous entry intact. This deliberately does not reuse `reliability_metrics::write_private_atomic`: that helper keeps a backup because losing a measurement loses evidence, whereas a cache entry is reconstructible and a failed swap should simply keep what was already there.
- **Everything unusable is purged, not served.** Past schema generation, foreign instance, rotated key, tampered header, truncated file, non-UTF-8 plaintext. Each is recorded as a classified `OfflineCacheFault` with a *static* detail string — a cache diagnostic must not become the leak the encryption exists to prevent — and surfaced through `offline_cache_status` alongside occupancy.
- **Bounded.** 1 MiB per entry, 256 entries, 32 MiB total, evicted oldest-first; abandoned foreign instance directories are reclaimed after 30 days; stale staging files are swept at startup.

## Read-only, and sanitized before it is stored

`src/lib/offline-cache.ts` is the only writer. It sanitizes first: credential-shaped keys (`token`, `apiKey`, `Authorization`, `cookie`, `privateKey`, …), embedded file bytes (`bytes`, `base64`, `data:` URLs, long base64 runs), class instances and typed arrays, and prototype-polluting keys are all dropped before serialization. Reads come back with `readOnly: true` from the native layer and are refused if that flag is missing, so a surface can label an offline copy without inferring it from a timestamp.

Outside the desktop shell every call is a no-op returning `null`/`false`. There is no keychain in the browser or on mobile, so the honest answer is "no cache" rather than an unencrypted fallback.

## Permissions

The four commands are granted to **both** the local app origin (`permissions/default.toml`) and the sidecar loopback origin the packaged UI actually runs on (`capabilities/loopback-browser.json`) — granting only the former would have made the cache unreachable in a real install. `desktop-permissions.test.mjs` pins the registration, the local-data-dir scoping, the instance check, and the client's command surface.

## Platform note

On Linux the keychain backend is the kernel session keyring (`keyring`'s `linux-native`), chosen so the build pulls no D-Bus stack. The secret does not survive a logout; the cache then reads as undecryptable, purges itself, and refills. Degraded, never wrong. macOS and Windows use their persistent native stores.

## Verification

| Gate | Result |
|---|---|
| `cargo test --lib` | 184 passed, 0 failed (20 of them new) |
| `cargo check --locked --lib` | clean, no new warnings |
| `node --test src-tauri/permissions/desktop-permissions.test.mjs` | 16 passed |
| `src/lib/offline-cache.test.ts` | 15 passed |
| `pnpm lint` | clean |

⚠️ **`pnpm typecheck` and `pnpm check:tests-wired` currently fail on `main` itself**, from the `cave-20wrn` streaming merge — `src/components/chat-view.tsx` references an undeclared `bubbleSegments`, and eight streaming test files landed unwired. Neither file is touched here. #4784 is the fix; this PR's CI cannot be green until that lands, and this stays a draft until it is verified against a repaired base.

## Not in this change

Showing and labelling the cached copy in the chat surface. The data layer asserts read-only; the UI that says so out loud is a follow-up.


CompleteDotTech:
Every red check on this head (`4177630d1`) reproduces on untouched `main`. Recording the evidence so the next reader doesn't re-derive it.

Run [32456206523](https://github.com/OpenCoven/coven-cave/actions/runs/32456206523) fails in exactly four places, and each one is the `cave-20wrn` streaming merge rather than anything in this diff:

| Job | Failure | Same failure on #4784 |
|---|---|---|
| Frontend validation (app tests) | `src/components/artifact-comments.test.ts` | yes |
| Frontend validation (API tests) | `src/app/api/chat/send/chat-attention-persistence.test.ts` | yes |
| Frontend validation (mobile tests) | `src/components/chat-view-mobile-command-center.test.ts` | yes |
| Frontend bundle / typecheck / test wiring | `chat-view.tsx` — `bubbleSegments` undeclared, `ChatToolActivityLayout` unused, 8 unwired streaming tests | yes |

Checked out `origin/main` at `b5f40e54d` with none of this branch applied and ran the three test files directly — all three fail there too. `pnpm typecheck` and `pnpm check:tests-wired` also fail on that bare checkout.

What this branch's own gates do:

- `cargo test --lib` — 184 passed, 0 failed (20 new)
- `cargo check --locked --lib` — clean, no new warnings
- `node --test src-tauri/permissions/desktop-permissions.test.mjs` — 16 passed, including the new offline-cache pin, which **also passed in CI** on this head
- `src/lib/offline-cache.test.ts` — 15 passed
- `Frontend validation (lint)`, `protocol conformance`, `Select validation` — green in CI on this head

Merging #4784 into this branch locally resolves cleanly — including `scripts/run-tests.mjs`, which both branches append to — and on that combined tree `pnpm typecheck` and `pnpm check:tests-wired` both pass.

Staying a draft until `main` is repaired, then re-running to confirm green.

CompleteDotTech:
## Review of `4177630d1`, with fixes applied

Reviewed against the nine claims and the twelve author doubts. Nine commits on top of the original; final head `ecee18885`.

---

### 1. Applied

| Commit | Fix |
|---|---|
| `bd7935782` | **Claim 2 was false: the cache path was recomputable without the keychain.** `scope_id`/`entry_id` were plain SHA-256 over the caller's strings, so both are one line of code away for anyone holding this source. `OfflineCacheScope` is a single-variant union, which makes that directory name a *published constant*; and an observer with a list of candidate session ids could confirm, one digest at a time, exactly which conversations an install had read offline — the precise disclosure the encryption exists to prevent. Both ids are now HKDF-SHA256 (HMAC under the hood) under the same keychain root secret, salted by the instance id, domain-separated from each other and from the content key. Generation bumped to `v2`; `purge_incompatible_generations` already discards a non-current generation wholesale, so that is the entire migration, and nothing consumes the cache yet so it costs one refetch in a build nobody runs. |
| `e5b33674f` | **Doubt 6 substantiated — foreign-instance reclamation deleted live installs.** A directory's mtime moves only when a child is created or removed, and an established cache does neither: `write_entry_file` renames over an existing entry, touching the *scope* directory and the file. So an instance directory carries the timestamp of the last time it gained a scope. A second install rewriting the same conversations lost its whole cache on day 31 while in use. Recency is now the newest mtime anywhere in the (two-level, bounded) tree. The old test read "a live instance is left alone" only because its tree was seconds old. |
| `7d46778fc` | **Claim 6 was write-side only.** `read_entry` called `std::fs::read` on whatever was at the path and measured the payload afterwards, from a header it had already parsed. Anything that grew by another route — corrupt filesystem, a build with a larger ceiling, anything with write access — was loaded into memory in full first. Size now comes from the directory entry before any read, classified `Oversized` and purged. |
| `523f11e4f` | **Doubt 7 — the rotation branch had no coverage,** and it is the one that empties the cache on launch. The decision is now `secret_from_stored`, a pure function; the keychain call is only the I/O around it. Pinned: a readable 32-byte value is adopted without rotating (including uppercase hex), and missing/short/long/non-hex/whitespace-padded values mint fresh keys. **The path cannot fire spuriously** — a locked or denied keychain returns `Err` and fails the read rather than minting. |
| `8646f9daa` | **Three fault paths the test suite could not have reached.** `LengthMismatch` and `PayloadMalformed` guard fields inside the AAD, so tampering always lands on `Undecryptable` first — they looked dead but are the check on *this module's own writer*. A helper now seals an envelope under the real entry key with a chosen header, the only way to produce bytes that authenticate and still lie. Third test: the entry key is deterministic in scope+name, so every rewrite reuses it, and **nonce reuse under a reused AES-GCM key surrenders the plaintext** — nothing here would have noticed a fixed nonce, since a round trip still succeeds and the payload is still absent from the file. All three fail when their guard is removed (verified by mutation, not assumed). |
| `d835a82bd` | **Every client-side bound was measured in the wrong unit.** The native side enforces UTF-8 bytes (`str::len`); the client checked `String.length`, UTF-16 code units. A CJK payload at half the nominal budget is already 1.5× the real one; a 128-character key is a 384-byte key. One-directional, so nothing oversized ever got stored — but "refused" arrived as a failed IPC round trip instead of the local check that exists to prevent it. |
| `ad24c623b` | **Doubt 1 — the `dropped` array had no observer.** `writeOfflineCache`, the only writer, destructured just the value. The write returned `true` and the cached copy quietly was not what the caller passed. Now reported through the existing `warn` dependency. Paths only, never values. |
| `3237e2ebe` | **Doubt 4 — `instanceId` removed from the status.** A stable per-install fingerprint, reachable from the sidecar loopback origin, that nothing consumed. Same shape as #4194. The id keeps every job it actually has (directory, HKDF salt, authenticated header field); it is simply not handed out. A test serializes the status and asserts the word is absent. |
| `ecee18885` | The Cargo comment breaks off mid-sentence — *"Desktop-only because the module is —"* — and the clause after it, about a sidecar cache, is not the reason. The reason is that `keyring` ships no iOS/Android backend. |

---

### 2. Rejected

**Doubt 9 — atomic-write divergence. The module's argument is correct.** Rust's `std::fs::rename` on Windows is `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING`; it does replace an existing destination. More to the point, `reliability_metrics`'s backup dance renames the *destination away first* (`rename(path, backup)` then `rename(temp, path)`), which opens a window where nothing is at the path at all. That is the right trade when losing a measurement loses evidence, and the wrong one for a cache entry. Using the hardened helper here would make claim 4 weaker, not stronger.

**Doubt 5 — no deadlock, and serialization is not a problem at these sizes.** `inner` and `faults` are separate mutexes and are never held nested: `record_fault` is called *after* `with_context` returns in `offline_cache_read`, and `offline_cache_status` reads `faults()` outside the closure. No command re-enters `with_context`. Each held critical section is one small file operation.

**Doubt 8 — once-per-launch is enough.** `configure()` nulls the context at setup, so both routines run on every launch. Staging debris only exists if a process died mid-write, and each launch sweeps it. (One cosmetic gap: `sweep_staging_files` walks only `instance_dir` and matches only `.bin.tmp-`, so a crashed `instance-id` write leaves `instance-id.tmp-…` at the cache root forever. Not worth a commit.)

**Doubt 11 — Cargo hygiene is clean.** The lock adds `linux-keyutils` and **no** `dbus`/`secret-service`, so the "no D-Bus stack" claim holds. `windows-native`/`apple-native`/`linux-native` are the correct keyring 3.x feature names, and the `macos`/`linux` target blocks exclude iOS/Android by construction. Moving `sha2` is safe: `sidecar_archive` is Windows-only and the desktop block is a superset of Windows. `cargo check --locked` passes at `rust-version = "1.77.2"`.

**Doubt 12 — permissions are complete and nothing overgrants.** Four commands, four permissions, both `default.toml` and `loopback-browser.json`, all four pinned in the test. The loopback capability is already scoped to `webviews: ["main"]` plus loopback remote URLs — the same scoping as the pty and browser grants beside it.

**Doubt 10, mostly.** The sort is deterministic (mtime desc, then path); an empty directory yields an empty eviction list; "oldest-first" is genuinely what it produces. A single entry over `MAX_TOTAL_BYTES` would evict everything, but it is unreachable — 1 MiB per entry against a 32 MiB total. The one real edge is filed below.

**Claims 1, 3, 4, 5, 7 verified as written.** Also verified: the existing opacity test would *not* pass with encryption removed (it asserts the secret's absence from the file), so it is not the empty test it might look like.

---

### 3. Open

**Doubt 2 — `readOnly` is a label, not a control, and the code is already honest about it.** The native layer sets it unconditionally and the client refuses an entry without it, so in production the check can never fail. What actually prevents a write-back is that no write path exists — the flag prevents nothing. But this is not #4194: the field *is* consumed, as the type-level marker on `OfflineCacheRead` that lets a surface say "offline copy" without inferring it from a timestamp, and the doc comments already describe it in exactly those terms rather than as enforcement. The client-side check is defensive redundancy against a mocked `invoke` or a future native change. Left as-is; deleting it would weaken claim 8 rather than de-theatre it.

**Doubt 3 — nothing consumes the API, and I think the split is right.** No surface reads or writes the cache. The one genuinely speculative piece of surface area was `instanceId`, and it is now gone; the rest (four commands, the sanitizer, the read-only framing) is the minimum a consumer needs. The value of landing it alone is that the schema is still free to change — `bd7935782` is proof of that, since keying the paths would have been a migration rather than a version bump if a surface were already writing.

**Doubt 1, on the threshold itself: defensible, kept.** `looksLikeEmbeddedBytes` is anchored (`^…+$`), so an inline blob inside a larger message does not match; only strings that are *entirely* base64 alphabet for 8 KiB+ are dropped. A minified JS snippet contains `{`, `(`, `;`, `.` and never matches. A base32 dump or a long hex run would — and those genuinely are bytes. The failure mode was that a wrong drop was invisible; `ad24c623b` fixes that rather than loosening the rule.

**Still untested, deliberately.** The `Io` fault has no portable trigger. The keychain functions are now covered at the decision layer but not against a real store — that needs a platform integration test, not a unit test.

**`with_context` ignores a failed `clear_entries` on rotation** (`let _ = …`). Now that paths are keyed, orphaned pre-rotation entries are unreachable rather than merely undecryptable, so a failed clear leaves bytes that only the cap eviction will ever reclaim. Bounded, so not worth a fix, but worth knowing.

**Filed:**
- `cave-pr3` (P2) — `aes-gcm`'s `zeroize` feature is off, so the expanded AES round-key schedule survives in freed heap even though the root secret and every derived key are `Zeroizing`. One feature flag plus a lock refresh; kept out of this PR to avoid lock churn on a shared branch.
- `cave-2pa` (P3) — mtime ties are broken by path, which is a keyed hash, so at the cap a write can be followed by the eviction of the entry it just wrote. Low impact at conversation write rates; the existing test passes either way.

---

### 4. Gate results

Run from the worktree on `ecee18885`.

```
$ cargo check --locked --lib
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.53s
    (9 warnings, all pre-existing and in unrelated modules; offline_cache adds none)

$ cargo test --locked --lib
test result: ok. 192 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 7.95s
    (up from 184 at 4177630d1: 8 new offline-cache tests)

$ pnpm lint
[tokenize-tsx-design] checked — 0 file(s) with drift
eslint --max-warnings=0 — clean

$ node --experimental-strip-types … src/lib/offline-cache.test.ts
ℹ tests 18   ℹ pass 18   ℹ fail 0

$ node --test src-tauri/permissions/desktop-permissions.test.mjs
ℹ tests 16   ℹ pass 16   ℹ fail 0
```

Two gates fail on this branch and both fail identically on bare `origin/main`; neither touches a file in this PR:

```
$ pnpm typecheck
src/components/chat-view.tsx(50,1): error TS6133: 'ChatToolActivityLayout' is declared but its value is never read.
src/components/chat-view.tsx(9001,22): error TS2304: Cannot find name 'bubbleSegments'.
… 5 more, all in chat-view.tsx

$ pnpm check:tests-wired
✗ 8 test file(s) on disk are not wired into any CI test suite
    (all 8 are the cave-20wrn streaming files)
```

**Both are green with #4784's repair merged in.** On a throwaway branch carrying this PR plus `pull/4784/head`:

```
$ pnpm typecheck
(no output — clean)

$ pnpm check:tests-wired
✓ all 1757 test files wired into CI
```

So this PR contributes nothing to the red baseline, and adds no new failure on top of the repair. The scratch branch was deleted; `chat-view.tsx` and the chat send pipeline were not touched.

CompleteDotTech:
Closed the two follow-ups the review left open, rather than carrying them as debt against a branch that is already waiting on the base.

**`44e9d6c3c` — eviction could undo the write that triggered it (`cave-2pa`).** `enforce_caps` sorted newest-first by mtime and cut everything past the caps. Filesystem mtime is coarse, so entries written in the same tick tie, and the tiebreak was path order — a keyed derivation, which is to say arbitrary. At the cap that made it a coin flip whether a write survived its own cap enforcement. `write_entry` now names the entry it created and that entry takes its budget slot before anything competes for one, so it neither loses its place nor lets the cache run one entry over. A file larger than `MAX_ENVELOPE_BYTES` is also reclaimed ahead of the budget instead of counted into it — `read_entry` already refuses one from its metadata, so those bytes could never be served and were only crowding out entries that could. The invariant the loop rests on, that one entry always fits the total budget, is now a `const` assertion instead of arithmetic a reader has to redo.

**`bbeece248` — the one piece of key material that was not scrubbed (`cave-pr3`).** `aes-gcm`'s default features leave `zeroize` off, so the expanded AES round-key schedule inside `Aes256Gcm` was left in freed heap on every drop, in a module that wraps everything else in `Zeroizing`. The feature costs one line of lockfile because `zeroize` is already a direct dependency. Same gap either side of the keychain call, where the root secret's hex form was an ordinary `String` both coming back from the store and going into it; both are `Zeroizing` now.

Both new eviction tests were verified fail-first by mutation — neutralizing the protection makes `eviction_never_removes_the_entry_the_write_just_created` fail, and neutralizing the servability guard makes `an_unservable_file_is_reclaimed_before_it_can_crowd_out_a_servable_one` fail.

Gates on `bbeece248`:

```
cargo check --locked --lib   Finished, 9 pre-existing warnings, none from offline_cache
cargo test --locked --lib    ok. 194 passed; 0 failed   (was 192)
pnpm lint                    clean
offline-cache.test.ts        tests 18   pass 18   fail 0
desktop-permissions.test.mjs tests 16   pass 16   fail 0
```

`cave-pr3` and `cave-2pa` are closed against these two commits. The base is still `b5f40e54d` and still red; #4784 has pushed again and its checks are re-running.

CompleteDotTech:
**CI is green.** `main` was repaired by #4769 landing at `b9856a7ed`; merging it in as `ba66f0dd0` gave a clean run.

| Check | Result |
|---|---|
| **Frontend build** *(required)* | ✅ — including `cargo check --locked` + `cargo test --locked --lib` and both Playwright passes |
| PR checks | ✅ |
| Frontend validation — lint / typecheck / test wiring / protocol conformance | ✅ |
| Frontend validation — app / API / mobile tests | ✅ |
| Frontend bundle | ✅ |
| Select validation | ✅ |
| iOS build | skipped (path-aware) |

`mergeStateStatus: CLEAN`, `mergeable: MERGEABLE`, no conflicts.

Both new suites ran on Linux rather than only on the author's Windows box:

- `cargo test --locked --lib` — `test result: ok. 163 passed; 0 failed`, every `offline_cache::tests::*` among them
- `src/lib/offline-cache.test.ts` — ran in the API suite, e.g. `✔ credential-shaped keys never reach the payload`, `✔ prototype-polluting keys are refused`, `✔ an over-budget payload is refused without touching the native side`, `✔ nothing is read or written outside the desktop shell`
- `src-tauri/permissions/desktop-permissions.test.mjs` — `✔ the offline read cache is registered, instance-scoped, and served read-only`

Still a draft, and the one thing that would change my mind about that is in the PR body: nothing consumes this cache yet. It is a complete, tested data layer waiting on the surface work that shows and labels an offline copy. Ready for review whenever someone wants to take it.

Issue #4779, opened 2026-08-21
[Chat v1 P6] Enforce shell, list, stream, route, bundle, and CLI budgets


CompleteDotTech:
# [Chat v1 P6] Enforce shell, list, stream, route, bundle, and CLI budgets

Create reproducible performance fixtures and enforce warm/offline shell, 10k-list, 10k-event, Cave read-route, bundle, package, and doctor deadlines.

Owner repository: OpenCoven/coven-cave (program tracker)
Execution lane: cross-repo-hardening
Plan of record: OpenCoven/chat/docs/superpowers/plans/2026-08-15-phase-6-production-hardening.md

Deliverables:
- Reference-machine shell and offline timing harness.
- Large list/event memory and input-blocking benchmarks.
- Cave route, bundle, package, and CLI deadline checks.

Acceptance criteria:
- All approved production budgets are machine-enforced.
- Streaming and pagination remain bounded in memory.
- SDK imports perform no I/O.

## Bead metadata

- **Bead ID:** `cave-o8gc4`

- **Status:** open

- **Priority:** P1

- **Surface:** Shared

- **Authoritative owner:** Unassigned

- **Labels:** `area:performance`, `performance`, `phase:6`, `program:chat-v1`, `release-blocker`, `repo:cross-repo`, `surface:shared`, `verification-required`

- **Parent:** None

- **Dependencies:** `cave-z2af3`, `cave-rbikx`

## Source of truth

Beads is authoritative. This draft card is a GitHub Teamwork visibility mirror and should be reconciled from the Bead rather than edited as an independent task.

BunsDev:
Closing this accidental repository-issue conversion. The canonical task remains the Bead mirrored as a draft card on the Teamwork project; Beads remains the source of truth.

CompleteDotTech:
Record correction: this was closed as `NOT_PLANNED`, but the work was delivered.

PR #4788 squash-merged 2026-08-21T18:17:10Z as `571efb4ff` on main (exact-head match on `f7e354c51`), 10 files, +2321/-21. Delivered: the performance-budget catalogue with fail-closed evaluation (6 enforced, 9 delegated, 4 pending), a relative ratio budget, phase-6 10k fixture profiles, a Windows-hermetic benchmark, cold-scan measurement, report and workflow wiring, and docs.

[...]

PR #4788, opened 2026-08-21
feat(performance): enforce phase-6 production budgets (cave-o8gc4)

CompleteDotTech:
Closes part of #4779 (bead `cave-o8gc4`).## WhyEvery budget in this repository was a size or a count, each enforced by its own script with its own private constants — bundle bytes, standalone file counts, sidecar closure entries. Nothing enumerated them, so the question this issue actually asks — *are all approved production budgets machine-enforced?* — could not be answered without reading every gate. Worse, a surface with no gate at all was indistinguishable from one that passed.## What`src/lib/performance-budgets.ts` is the single catalogue. Each entry names the gate that enforces it, so the catalogue stays a directory rather than becoming a second gate over numbers `postbuild` already owns:| `gate` | Meaning | Can it fail a run? || --- | --- | --- || `performance-report` | Evaluated against the metrics that run produced | Yes || `postbuild` | Already enforced by a build gate; recorded, not re-checked | No || `pending` | Value approved, no fixture yet; printed every run | No |**Missing measurements fail closed.** An enforced budget whose metric is absent is `unmeasured`, and `unmeasured` fails the run. A benchmark that crashed emits no metric, and treating "no measurement" as "no breach" would turn every such outage into a green run — a silent failure that looks exactly like success. Same asymmetry the maintenance gate takes on a missing plane entry.**Every `source` is machine-resolved.** A `performance-report` source names a fixture profile that must exist; a `postbuild` source names a gate constant that is imported and resolved; a `pending` source may be prose *only* when nothing in this repository owns the number, and one naming a repository file must use the same `<file> (<CONSTANT>)` form so the limit can be asserted equal to the declaration. Prose that merely mentions a `.ts`/`.mjs` path is rejected.**Completeness is derived from the gates, not asserted.** All nine build-gate numbers that exist today are recorded, and the unit suite reads the constants out of `bundle-budget.mjs`, `standalone-budget.mjs` and `sidecar-runtime-closure.mjs` and requires an entry for each.`pending` entries are the honest half: the warm/offline shell deadlines, the 10k-event stream ceiling, and the `coven doctor` deadline are approved numbers with no harness behind them yet, so they appear in every report instead of being mistaken for coverage.## Two fixture defects fixed on the way- **The fixture was not hermetic on Windows.** The benchmark pinned `HOME`, but `caveHome()` falls back to `os.homedir()`, which reads `$HOME` only on POSIX — on Windows it reads `USERPROFILE` and ignores `$HOME`. So the fixture isolated correctly on Linux CI while writing its conversations into the developer's real `~/.coven/cave/conversations`, where the temp-dir cleanup could never reclaim them: 10,000 stray files at the phase-6 scale. It now pins `COVEN_CAVE_HOME` and refuses to run if `CONV_DIR` resolves outside its temp home.- **A blank `CAVE_BENCH_*` env var was coerced to zero** rather than treated as absent, so a workflow writing `${{ inputs.iterations }}` on a scheduled run would select zero iterations and read off the end of an empty array. Blank now falls back to the profile; malformed is rejected outright.The benchmark also now measures Cave's own cold scan, cleared per iteration. The `before` figure is a raw sequential `readdir` + `JSON.parse` control loop — it carries no ceiling of its own, since budgeting it would police the harness, but it *is* the divisor in the relative budget below.## What review found after the first commitEach of these was reproduced by execution before it was fixed, and each has a test that fails without the fix.- **The catalogue restated a gate's own number and got it wrong** — 900 KB for home first-load while `bundle-budget.mjs` defaults to 2800 KB. `postbuild` entries now carry `limit: null`, and the unit suite enforces that.- **A delegated `source` named a symbol that does not exist** — `STANDALONE_BUDGETS.bytes`; the real key is `unpackedBytes`. The unit suite now imports each gate and resolves the dotted path, because grepping for the string passes: `bytes` occurs in that script's own comments.- **Three separate doors into "a smaller-than-seeded run certifies the 10k budgets".**  - *The profile door.* `pnpm performance:report` with no profile ran the 100-conversation `default` fixture and printed `10k conversation list, cold metadata scan p95 | 38.07 ms | ≤ 3000.00 ms | pass`. The report now defaults to the seeded profile, and a measured profile that differs reads `unmeasured`.  - *The override door.* `CAVE_BENCH_PROFILE=phase-6-list-10k CAVE_BENCH_CONVERSATIONS=25` still reported `profile: "phase-6-list-10k"`. A dimension differing from its profile now appends `(overridden: <VAR>)`, which matches no catalogue entry.  - *The fixture door.* Shrinking `phase-6-list-10k` to 200 conversations kept the profile *name*, so both guards above passed and the report certified `30.07 ms | ≤ 3000.00 ms | pass` with exit 0. The seeded `fileCount`, `transcriptBytes` and `iterations` are pinned in the unit suite, so re-scaling the fixture has to reseed the limits in the same commit.- **The enforced timing budgets graded a "p95" that was really a maximum.** `percentile(values, 0.95)` returns the largest sample for every n ≤ 20, and the fixture runs 5 iterations — so "cold metadata scan p95 ≤ 3,000 ms" was "slowest of five ≤ 3,000 ms", decidable by one descheduled iteration. Measured: a busy box produced a cold p95 of **16,058 ms against a p50 of 1,840 ms in the same run**, a 5x breach from scheduling noise with the workload unchanged. The enforced ms budgets now grade the **median**; the p95 stays measured, reported and baseline-compared.- **The flagship ms ceiling could not catch the regression it was written for.** See the next section — this is the largest single change since the first commit.- **The catalogue was silently partial: 2 of the 9 build-gate budgets.** `BUNDLE_MAX_SHELL_KB`, `BUNDLE_MAX_CHUNK_KB`, `BUNDLE_MAX_ROOT_CSS_KB`, `BUNDLE_MAX_HOME_CSS_KB`, `STANDALONE_BUDGETS.fileCount` and both sidecar runtime closure ceilings were absent, so a reader trusting the directory concluded the bundle surface had one ceiling when it has five. The only completeness check was "every *surface* carries at least one budget", which a one-of-five surface satisfies. All nine are recorded now, and the check derives the constants from the gates themselves — `bundle-budget.mjs` by the exact `Number(process.env.X)` shape its five knobs use (it exits at import time, so it cannot be imported), the two frozen `*_BUDGETS` objects by importing them — with the gate list cross-checked against the files the catalogue delegates to so neither side can drift alone. Verified by deleting one entry: the suite fails naming the constant it lost. Two of the nine are file *counts*, which the `ms | bytes | percent` unit union could not express; the union grew a `count` rather than the entries being dropped with a note. (`cave-21w`)- **A pending `source` asserted against nothing.** `cli.doctor.p95-ms` explained itself in prose — "matches `EXEC_TIMEOUT_MS` in `src/app/api/coven/exec/route.ts`" — so moving that constant would have left the catalogue stating a stale ceiling while still reading like a live delegation.- **A fixture profile that describes no workload ran green and silent.** A profile missing `transcriptBytes` wrote `"x".repeat(undefined)` — 10,000 empty transcripts — and exited 0 with every budget inside its ceiling. The benchmark refuses such a profile at load; the unit suite validates every committed one.- **Nothing pinned the one line that turns a breached budget red.** The workflow step runs the report under `set +e` so a failing run still writes its job summary, which makes `exit "$report_status"` the whole enforcement. The capture and the propagation are asserted now.- **The new doc was not in the docs index.** `scripts/docs-index.test.mjs` pins `docs/README.md` as a complete classification of every top-level `docs/*.md` and is wired into the `app` suite.- A blank `CAVE_BENCH_PROFILE` crashed the benchmark instead of falling back; the delegated-source regex used named capture groups this project's `tsc` target rejects; and a test hard-coded an iteration count the fixture owns.## The cold-scan budget, and what it is really measuring**The mechanism two earlier drafts described was wrong.** The cold scan does *not* skip parsing: `readConversationSummary` reads each file whole and `JSON.parse`s it exactly as the control loop does, then derives signals on top. What the shipped path adds is an 8-way read pool (`CONVERSATION_LIST_READ_CONCURRENCY`) and a stat-keyed summary cache. So "a cold scan that regressed back to parsing every transcript" was never a describable regression, and `cold-scan.bytes` cannot separate the two loops for a simpler reason than the one previously given: both read all 43,608,890 bytes, and always did.**The describable regression is the cold scan costing what the naive sequential loop costs, and 5,000 ms does not catch it.** Measured by causing it — `CONVERSATION_LIST_READ_CONCURRENCY` set to 1, `phase-6-list-10k`:| platform | cold p50 | control p50 | ratio || --- | ---: | ---: | ---: || windows | 4,397.85 ms | 3,926.19 ms | 112.01% || linux (wsl2) | 3,460.77 ms | 2,718.62 ms | 127.30% |Both cold medians are *inside* the 5,000 ms ceiling. `cold-scan.bytes` sees nothing, `cache-hit-rate` is taken from the warm loop, and the 20% baseline comparison does classify it as a regression but the nightly runs without `--fail-on-regression` — so that run exits 0 with every budget green. Tightening is not the repair: the slowest legitimate median on record is 2,603 ms, so any limit under the ~3,900 ms control cost reinstates the noise-driven red nightly that moving off p95 removed.**So the catalogue grew a relative budget kind (`cave-4e1`).** `ratioOf` names a numerator and a denominator metric and the value is graded as a percentage; `conversation-list.cold-scan.share-of-full-parse-pct` divides the cold median by the control median from the *same run*. That cannot be defeated by a fast machine, because the machine sets both numbers. It fails closed on a missing numerator, a missing denominator, or a denominator that is zero, negative or non-finite — `x / 0` is `Infinity`, which a ceiling reads as a breach and a floor as a pass, both by accident.**It is platform-sensitive, and seeding it here alone would have shipped a nightly close to red.** The read pool only pays off where per-file I/O overhead dominates, and it dominates far less on Linux. The nightly runs on `ubuntu-24.04`, so it was measured there too:| run | cold p50 | control p50 | ratio || --- | ---: | ---: | ---: || windows, idle | 994.70 ms | 4,096.10 ms | 24.28% || windows, three concurrently | 1,809.08 ms | 4,718.10 ms | 38.34% || linux (wsl2), idle | 1,759.80 ms | 2,787.94 ms | 63.12% || linux (wsl2), three concurrently | 1,639.41 ms | 3,117.21 ms | 52.59% |Contention moves the ratio in *opposite* directions on the two platforms — up on Windows, where the read-bound cold loop suffers more, and down on Linux, where the parse-bound control loop does — so the worst healthy reading is a *quiet* Linux run. The first draft of this budget was 75%, seeded on Windows alone; the nightly's own platform reads 63%.The ceiling is **90%**, near the middle of the 63.12%–112.01% band separating every healthy run measured from every collapsed one. The unit suite pins it from both sides: every healthy pair must pass, every collapsed pair must breach. Verified by moving the limit — at 50 the Linux idle run goes red, at 120 the Windows collapse goes green. Re-seeding from a real `ubuntu-24.04` nightly is filed as `cave-wxq`.## Seeding the absolute ceilings`phase-6-list-10k`, 43.6 MB scanned every time:| run | cold p50 | cold p95 | warm p50 | warm p95 || --- | ---: | ---: | ---: | ---: || first seeding, 2026-08-21 | — | 1,039 ms | — | 82 ms || windows, idle | 1,364 ms | 1,398 ms | 156 ms | 157 ms || windows, three concurrently | 2,603 ms | 2,686 ms | 156 ms | 159 ms || linux (wsl2), idle | 1,760 ms | 1,800 ms | 198 ms | 213 ms || linux (wsl2), three concurrently | 1,639 ms | — | 177 ms | — |The p95/p50 ratio is 1.02–1.04 whenever nothing stalls, which is why the original 3,000 ms ceiling read as 3x headroom — against an idle single run it was. Against three concurrent runs the *median* already reaches 2,603 ms, leaving 15%.The cold ceiling is **5,000 ms** — 1.9x the slowest median measured. Warm stays **750 ms**. Both hold on Linux with room (cold 1,760 of 5,000, warm 198 of 750), so neither moves; the Linux rows are recorded anyway, because a limit whose entire justification comes from a platform CI never runs is a claim nobody has checked.Raising the sample count is not an alternative to the median: n ≥ 21 merely stops p95 being the maximum, ~100 gives it resolution, and one iteration costs 8–10 s, so 100 would be ~17 minutes inside a 30-minute job that also seeds 10,000 files. `src/lib/performance-budgets.test.ts` asserts the shipped limits clear the *slowest* readings recorded, and enforces the rule rather than the instance: every enforced ms budget must be a `.p50-ms` metric labelled "median", every operand of a relative budget must be one too, and `iterations` is pinned because a median is only robust while five samples sit under it.Fixture scale lives in `fixtures/phase-6/performance-fixtures.json` so the workload and the budget it is judged against move together.## Verification- `src/lib/performance-budgets.test.ts` — 29 tests: registry invariants, ceiling/floor semantics, boundary equality, fail-closed on absent and on `NaN`, delegated-constant resolution, gate-derived completeness, pending-constant resolution and the prose ban, fixture-profile validity, the seeded-workload pin, the enforced-statistic rule, the relative-budget kind (pass, breach, both missing operands, an unjudgeable denominator, and the id-collision case), and the measured healthy/collapsed ratio band from both platforms.- `scripts/cave-performance-report.test.mjs` — 15 tests: passing verdict, breach flips status to `fail` with no baseline, a single stalled iteration does *not*, a missing cold scan fails closed, smoke-scale and swept-dimension runs fail closed, the relative budget resolved against the ids the report actually emits, the benchmark's own override stamp, the markdown table, and the workflow's shape and exit-code propagation.- Full `pnpm performance:report` on `phase-6-list-10k`, end to end: `status: pass`, 0 breaches, 0 unmeasured, exit 0; 6 enforced, 9 delegated, 4 pending. Cold median 984.72 ms against 5,000 ms (80.3% headroom), the relative budget 25.22% against 90%, warm median 106.22 ms against 750 ms (85.8%), 41.59 MiB scanned against 64 MiB, 100% cache hit rate.- `pnpm exec eslint` clean on both new files; `node --test scripts/docs-index.test.mjs scripts/ci-paths.test.mjs` passes.## Base and CI status

This branch was opened on `b5f40e54d`, which carried the direct-to-main merge `5771a5558` and the `typecheck` / `check:tests-wired` / suite failures it put on `main`. That is now history. `main` has been repaired, and `f7e354c51` merges `origin/main` at `b9856a7ed` into this branch: no conflicts, nothing dropped, and no change to any file this PR owns. `git diff origin/main HEAD` and `git diff origin/main...HEAD` are the same ten files, which is the check that a merge neither reverted someone else's work nor swallowed this branch's.

**The earlier "any failure you see is inherited" caveat no longer applies.** CI on `f7e354c51` is green on every job that has reported:

| check | result |
| --- | --- |
| Frontend validation (typecheck) | pass |
| Frontend validation (lint) | pass |
| Frontend validation (test wiring) | pass |
| Frontend validation (app tests) | pass |
| Frontend validation (API tests) | pass |
| Frontend validation (mobile tests) | pass |
| Frontend validation (protocol conformance) | pass |
| Frontend bundle | pass |
| Select validation | pass |

Reproduced locally on the same head: `pnpm typecheck`, `pnpm lint`, `pnpm check:tests-wired` (all 1758 files wired), `pnpm test:mobile`, `scripts/docs-index.test.mjs`, both new test files (29 + 15, 0 fail), and a full `pnpm performance:report` on `phase-6-list-10k` — `status: pass`, 0 breaches, 0 unmeasured, exit 0, 6 enforced / 9 delegated / 4 pending.

The catalogue's guards were mutation-tested rather than trusted: deleting the `bundle.shell.bytes` entry fails with *"scripts/bundle-budget.mjs enforces BUNDLE_MAX_SHELL_KB, but no catalogue entry records it"*; moving the relative ceiling to 50 fails with *"linux (wsl2), idle measured 63.12%, over the shipped ceiling"*; moving it to 120 fails with *"windows, read pool 1 measured 112.01% and passed"*.

Two suites still fail in a **Windows** shell only, in files this PR does not touch, and both pass on Linux CI above:

- `scripts/check-conflict-markers.test.mjs` — the CLI is gated on ``import.meta.url === `file://${process.argv[1]}` ``, which never matches on Windows (`file:///C:/…` against `C:\…`), so the gate exits 0 on a tree full of markers. Present since #4434; filed as `cave-zya`, not fixed here.
- `scripts/dev-app.test.mjs` — spawns `node` through `/bin/bash`, which has no `node` on PATH in this environment.

The PR remains a **draft**.


CompleteDotTech:
## Review pass at `2c8976ab9` → repairs landed at `0228d4586`

Four commits, each reproduced by execution before it was written.

### Changed

**`cdf871d93` — a delegated budget pointed at a constant that does not exist.**
`package.standalone.bytes` named `STANDALONE_BUDGETS.bytes`; the real key is
`unpackedBytes`. A `postbuild` entry carries `limit: null` on purpose, so
`source` is its *entire* tie to the gate that owns the number — a stale name is
a dangling pointer that renders in every run's budget table as a live
delegation. Same class as the 900 KB / 2800 KB drift fixed in `2c8976ab9`.
The unit suite now **resolves** the name instead of matching it: an exported
object is imported and the dotted path walked; an env knob is matched on the
gate's literal `process.env.<NAME>` read (`bundle-budget.mjs` runs its whole
check at import time and calls `process.exit`, so it is never imported).
A plain string search would not have caught this one — "bytes" occurs
word-for-word in `standalone-budget.mjs`'s own comments and log strings.
Verified the assertion fails on the old value.

**`9f627c436` — a smoke-scale run certified the 10k budgets.** Reproduced:
`pnpm performance:report` with no profile ran the 100-conversation `default`
fixture and printed

```text
| list | 10k conversation list, cold metadata scan p95 | 38.07 ms | ≤ 3000.00 ms | 98.7% | pass |
```

with `summary.budgetPass: true` and exit code 0. Five approved production
budgets green off a run that went nowhere near any of them, while
`docs/performance-budgets.md` already promised "the workload and the budget it
is judged against move together". The evaluation now compares the profile the
benchmark reports against the profile the budget's `source` names and records a
mismatch as `unmeasured` — which already fails the run — and a run that does not
identify its fixture at all fails the same way, matching this module's existing
treatment of absent data. Each `unmeasured` result carries a `note` so the table
and stderr say *which* way it failed.

So this does not merely make the report harder to satisfy: `performance:report`
now defaults its benchmark to the profile the enforced budgets were seeded
against, so an ad-hoc run measures what it is about to be graded on. A test pins
the workflow's `profile` input and job env to that same catalogue value, so the
two cannot drift without the unit suite failing instead of the nightly.

**`46bacd271` — a blank `CAVE_BENCH_PROFILE` crashed the benchmark.**
`dimension()` exists in that file precisely because `FOO: ${{ inputs.foo }}`
sets an empty string on a scheduled run and `??` only catches `undefined` — but
the profile selector directly above it was left on `??`, so a blank value looked
up a profile named `""` and died on `unknown benchmark profile ""` (reproduced).
Not live today; the `iterations` input next to it is written in exactly the
unguarded shape, so the next copy of that line reintroduces it.

**`0228d4586` — my own regression in `cdf871d93`.** The matcher used
`(?<name>…)`; `node --test` and ESLint were both clean but this project's tsc
target predates ES2018, so `pnpm typecheck` went from the 7 inherited
`chat-view.tsx` failures to 9. Caught by the differential check, fixed with
positional groups.

### Investigated and rejected

- **Are the seeded limits too loose to ever fire?** Measured
  `phase-6-list-10k` on a fast local NVMe box: cold p95 **997–1,114 ms** across
  runs (seed 1,039), warm p95 **87–123 ms** (seed 82). Local run-to-run spread
  alone is already ~1.5x on the warm number. With **no CI datapoint yet**,
  tightening 3,000 / 750 risks a red nightly on runner variance rather than on a
  regression. Reseat from the first scheduled runs, not from a guess.
- **Fixture cost in CI.** Full `pnpm performance:report` at `phase-6-list-10k`,
  including the daemon-reliability benchmark: **31.9 s** end to end here (10,000
  files, 43.6 MB, 5 iterations × three scan loops). Not remotely marginal
  against the 30-minute timeout.
- **The blank-env fix.** `inputs` is empty for a non-dispatch event, so
  `CAVE_BENCH_ITERATIONS: ${{ inputs.iterations }}` is `""` on a schedule and
  `dimension()` falls back; an explicit dispatch value still wins. Correct as
  written, and now pinned by a test.
- **Windows hermeticity.** `caveHome()` reads `COVEN_CAVE_HOME` first, the
  benchmark overwrites it before importing `cave-conversations.ts`, and
  `benchHome` is a fresh `mkdtemp`, so a pre-set env cannot defeat the guard and
  there is no pre-existing path inside it to symlink. I could not construct a
  defeat. (Nit, not fixed: the check is `startsWith` without a trailing
  separator, so it means "same prefix", not "inside".)
- **Baseline backward compatibility.** Executed against a synthetic
  pre-change `report.json` (no `cold` metrics, no `budgets` block, no
  `fixture.profile`): `buildPerformanceReport` does not throw, history chains to
  2 points, and `controls.conversationFixture?.profile` degrades to `unknown`
  both when `profile` is missing and when `fixture` is absent entirely.
- **`run-tests.mjs` conflict risk.** Both added lines still apply; `mergeable`
  is `MERGEABLE`; `src/lib/performance-budgets.test.ts` is in the `app` suite
  and `check:tests-wired` does not name it.

### Deliberately left alone

The red `pnpm typecheck` (7 errors, all `chat-view.tsx`) and
`pnpm check:tests-wired` (8 unwired streaming/chat files) are inherited from
`5771a5558` on `main` and belong to #4784. The differential check is
byte-identical between this branch's tip and its base for both gates.

### Verification

```text
node --experimental-strip-types --test src/lib/performance-budgets.test.ts \
  scripts/cave-performance-report.test.mjs      27 pass / 0 fail
pnpm test:performance-report                    11 pass / 0 fail
pnpm exec eslint <all touched files> --max-warnings=0   exit 0
pnpm typecheck        7 errors, all src/components/chat-view.tsx (unchanged vs base)
pnpm check:tests-wired  8 files, all streaming/chat (unchanged vs base, byte-identical)
pnpm performance:report   status pass, 0 breaches, 0 unmeasured,
                          fixture phase-6-list-10k (10000 conversations), 31.9 s
CAVE_BENCH_PROFILE=default pnpm performance:report   status fail, 5 unmeasured, exit 1
```

Left as a draft; not marked ready, not merged.

CompleteDotTech:
Follow-up filed: **`cave-21w`** — the catalogue records 2 of the 7 build-gate budgets that already exist (missing `BUNDLE_MAX_SHELL_KB`, `BUNDLE_MAX_CHUNK_KB`, `BUNDLE_MAX_ROOT_CSS_KB`, `BUNDLE_MAX_HOME_CSS_KB`, `STANDALONE_BUDGETS.fileCount`, and the sidecar closure). Real, but adding five entries is a scope change rather than a repair, so it stays out of this PR. The unit suite only asserts that every *surface* carries at least one budget, so a partly-populated catalogue passes today — `cave-21w` proposes tightening that alongside the additions.

CompleteDotTech:
## Review round 2 — four changes on top of `0228d4586`

Round 2 targeted the code round 1 itself added. Two of the four are behavioural fail-opens, both reproduced before being fixed.

### Fixed

**`7e66cca92` — a swept dimension inherited its profile's authority.** Round 1's scale check (`9f627c436`) compares the profile *name* the benchmark reports, and the benchmark stamped that name from `CAVE_BENCH_PROFILE` alone. Every `CAVE_BENCH_*` dimension override therefore kept the profile's authority while changing the workload out from under it:

```
CAVE_BENCH_PROFILE=phase-6-list-10k CAVE_BENCH_CONVERSATIONS=25 \
CAVE_BENCH_ITERATIONS=1 pnpm bench:conversation-list
```

reported `profile: "phase-6-list-10k"`, and feeding that payload to `buildPerformanceReport` returned `status: "pass"`, `budgetPass: true`, and cold-scan p95 `pass` against the 3,000 ms ceiling — over twenty-five conversations. That is the same smoke-certifies-10k defect the profile check exists to close, reached through the override door instead of the profile door, and it is not hypothetical: the workflow exposes `iterations` as a dispatch input and `docs/performance-budgets.md` advertises dimension sweeps as the bisect route. A dimension differing from its profile's own value now appends `(overridden: <VAR>)` to the reported profile; that string matches no fixture key, so the existing mismatch path records every enforced budget `unmeasured` and fails the run. A value *equal* to the profile's is not a sweep, so the nightly's blank `iterations` still certifies.

**`66d02065d` — a fixture profile that describes no workload was runnable.** `dimension()` rejects a malformed env override but trusted its fixture fallback absolutely, and "the profile exists" was the only assertion any test made about a committed profile. The malformed shapes do not fail equally loudly: missing `iterations` crashes in `percentile()`, missing `fileCount` breaches the cache-hit floor — but missing `transcriptBytes` writes `"x".repeat(undefined)`, i.e. empty transcripts, and exits 0 with every budget green. Demonstrated by adding `{label, fileCount: 5, iterations: 2}` to the fixture file. The benchmark now refuses such a profile at load, and the unit suite asserts every committed profile carries a positive `fileCount`/`transcriptBytes`/`iterations`.

**`b3491e966` — nothing pinned the line that turns a breached budget red.** In CI the budget verdict travels by exit code and by nothing else, and the `Generate report` step runs the report under `set +e` on purpose so a failing run still writes its job summary. A single trailing `exit "$report_status"` re-raises it. The workflow-shape test pinned `--baseline`, `if: always()` and the upload action SHA, but not that. Delete the line and every breach reports green with no failing assertion anywhere. Verified the new assertion bites by stripping the line from the parsed step body.

**`8ee6b846e`** — the new sweep test read its "unswept" iterations value from the fixture instead of a hard-coded `5`, so it keeps testing the property rather than a snapshot of it.

### Examined and rejected

- **Arithmetic/verdict edge inputs** — probed negative values in both directions, zero and negative limits, a metric appearing twice, `Infinity`, a budget id colliding with a `daemon-reliability.*` metric id, and duplicate budget ids. Every verdict is correct or conservative; `Infinity` reads `unmeasured`. The only oddity (`limit: 0` reporting `headroomPct: 0` on a breach) is unreachable — the catalogue test requires `limit > 0` — and does not affect the verdict.
- **`BUDGETED_FIXTURE_PROFILE` with zero or several enforced profiles** — resolves to `null`, no default is supplied, the benchmark runs `default`, and every enforced budget goes `unmeasured`, i.e. red. A unit test pins `enforcedFixtureProfiles()` to exactly one entry, so a second profile fails fast with an explanatory message.
- **The dynamic `import()` in the delegated-constant test** — `standalone-budget.mjs` is guarded by `process.argv[1] === fileURLToPath(import.meta.url)`, so importing it runs nothing. `bundle-budget.mjs` genuinely self-runs, which is why the no-dot branch matches on the `process.env.<NAME>` read instead. Correct as written.
- **`pending` entries asserted against nothing** — checked the one that names a constant: `EXEC_TIMEOUT_MS` in `src/app/api/coven/exec/route.ts` really is `6_000`, matching `cli.doctor.p95-ms`. No second drift of the `2c8976ab9` class.
- **Markdown table integrity** — no catalogue `label`, `source`, or generated note contains a pipe; `formatMetricValue` is unreachable with `null` (both `budgetRow` and `formatBudgetLimit` guard first). Rendered the real report and confirmed the table.
- **Exit-code propagation** — `pnpm performance:report --bogus` exits 1 through pnpm.
- **`budgetFixtureProfile()` returning `null` for a `performance-report` entry** (skipping the scale check) is the intentional profile-agnostic shape the unit tests use, and the catalogue test rejects it for shipped entries.
- Round 1's rejections (limit tightness, CI fixture cost, blank-env fallback, Windows hermeticity, baseline compatibility, `run-tests.mjs` conflict) were not relitigated — no new executable evidence overturned them.

### Verification

`node --test` on both files: **30 pass, 0 fail**. ESLint clean on all five touched script/lib files. `pnpm performance:report` → `status: pass`, 0 breaches, 0 unmeasured, budget table rendered correctly.

Differential against the inherited `main` breakage: `pnpm typecheck` reports the same 7 errors, all in `src/components/chat-view.tsx`; `pnpm check:tests-wired` reports the same 8 unwired streaming/chat files. Neither names a file this PR touches — inherited from `5771a5558`, owned by #4784, deliberately untouched here.

Staying a draft until `main` is green.

CompleteDotTech:
## Review round 3 — one defect found and fixed, and the accumulated guards verified not to over-fire

Reviewed at `8ee6b846e`; head is now `095128c9f`. Everything below was confirmed by execution, not by reading.

### Fixed

**`095128c9f` — the fixture profile's *contents* were unpinned, only its name was.**

A budget's `source` binds it to a profile by name, and the name outlives any edit to what the profile holds. I shrank `phase-6-list-10k` from 10,000 conversations to 200, ran the benchmark and fed its real output through `buildPerformanceReport`:

```text
benchmark reported fixture: {"profile":"phase-6-list-10k","fileCount":200,"transcriptBytes":4096,"iterations":5}
summary: {"status":"pass", ... "budgetPass":true,"budgetBreachCount":0,"budgetUnmeasuredCount":0}
 pass  10k conversation list, cold metadata scan p95 = 30.07 (limit 3000)
 pass  Cave conversation read route, warm cache p95 = 2.25 (limit 750)
 pass  10k conversation list, cold scan bytes = 871890 (limit 67108864)
```

That is the same certified-smoke-run defect `9f627c436` and `7e66cca92` each closed, reached through a third door. Neither existing guard can see it: the profile check compares *names*, and the swept-dimension stamp only watches `CAVE_BENCH_*` overrides — a committed fixture edit trips neither, and it is the door most likely to be opened in good faith by someone trimming nightly CI cost.

The scale a limit was approved against is as much part of the budget as the limit, so the unit suite now pins the seeded `fileCount` and `transcriptBytes`. Verified both ways: 18/18 on the shipped fixture, and the new test is the single failure against the shrunk one. `docs/performance-budgets.md` gained the matching paragraph.

### Verified, not changed

- **A healthy scheduled run still goes green.** Full `pnpm performance:report` on `phase-6-list-10k`, end to end, exit 0: `status: pass`, 0 breaches, 0 unmeasured, 5 enforced / 2 delegated / 4 pending. Cold scan 1,328 ms against 3,000; warm cache 134 ms against 750; cold bytes 41.59 MiB against 64; hit rate 100% against 95. Nine commits of fail-closed guards, and none of them fires on a healthy run.
- **The seeded limits also hold on a machine slower than the reference one.** Three consecutive 10k runs here: cold p95 1,370 / 1,523 / 1,361 ms, warm p95 143 / 145 / 147 ms.
- **The `(overridden: …)` stamp produces neither a false positive nor a false negative on the nightly path.** On a `schedule` event `inputs` is empty, so `CAVE_BENCH_ITERATIONS` is `""` and `dimension()` reads blank as absent — no stamp; `CAVE_BENCH_PROFILE` falls back to `phase-6-list-10k`; nothing else sets a `CAVE_BENCH_*` var, and `runJsonBenchmark` only fills the profile when it is unset. Validation of the profile runs before the first `dimension()` call, so the `value !== fallback` comparison is always against a checked positive number.
- **Delegated sources still resolve.** `BUNDLE_MAX_HOME_KB` is read at `bundle-budget.mjs:43`; `STANDALONE_BUDGETS.unpackedBytes` exists on the frozen export.
- **`cli.doctor.p95-ms` = 6,000 does match `EXEC_TIMEOUT_MS`** (`src/app/api/coven/exec/route.ts:30` is `6_000`). Correct today, but nothing checks it — see below.
- Both new test files are wired into the `app` suite, which `ci.yml` runs.
- `pnpm typecheck` output is byte-identical to `8ee6b846e`: seven errors, all in `chat-view.tsx`, all inherited from merge `5771a5558`. `pnpm check:tests-wired` output is byte-identical too. `pnpm exec eslint` clean.

### Left alone, deliberately

- **`percentile(values, 0.95)` returns the maximum for any n ≤ 20**, so with `iterations: 5` every enforced "p95" budget is really "slowest of five". `Math.floor(n * 0.95) === n - 1` holds for all n ≤ 20, so raising iterations to 20 would not change that either. Three idle-machine runs showed p50 within 2% of p95, but one run taken while the box was busy produced a cold p95 of 16,058 ms against a p50 of 1,840 ms in the *same* run — a 10x tail that breached the 3,000 ms ceiling. A shared `ubuntu-24.04` runner is a busy box. Not changed here because every available fix — reseat the ceiling, or budget a statistic five samples can support — needs a real CI datapoint, and the first nightly is what supplies it. Worth deciding then rather than guessing now.
- **A `pending` entry's `source` is free prose and is asserted against nothing.** The `postbuild` sources are machine-resolved since `cdf871d93`, but `cli.doctor.p95-ms` claims "matches `EXEC_TIMEOUT_MS`" with nothing to notice if that constant moves — the same drift class as the 900 KB / 2800 KB bug, one gate short of being caught.
- `STANDALONE_BUDGETS.fileCount` and the sidecar closure budget are absent from the catalogue. Not a gap: `PerformanceBudgetUnit` is `ms | bytes | percent`, a file count is none of those, and #4779 scopes the plan to the seven surfaces the catalogue already lists.
- Nothing in `src/components/chat-view.tsx` or the streaming tests was touched.

The PR description was stale after nine commits (it still described "two fixture defects" and 11 + 9 tests); it now lists what review found and the current counts, 18 + 13.

CompleteDotTech:
## Review round 4 — three fixes, all reproduced by execution

Head `2e1156866`. Still a draft.

### 1. The enforced "p95" budgets were really "slowest of five" (`92c5d84f3`)

Executed, not inferred: `percentile(values, 0.95)` returns the maximum for every n ≤ 20, and `phase-6-list-10k` runs 5 iterations. So `cold metadata scan p95 ≤ 3,000 ms` was a ceiling one descheduled iteration could decide — and a busy box has produced a cold p95 of **16,058 ms against a p50 of 1,840 ms in the same run**, a 5x breach of that ceiling from scheduling noise with the workload unchanged. A shared `ubuntu-24.04` runner is a busy box, and a gate that refuses on noise is an outage rather than a budget (`WORKTREE_WARNING_BUDGET`, `cave-qpwx0`).

Measured on the reference machine, `phase-6-list-10k`, 43.6 MB every scan:

| run | cold p50 | cold p95 | warm p50 | warm p95 |
| --- | ---: | ---: | ---: | ---: |
| first seeding | — | 1,039 ms | — | 82 ms |
| idle | 1,364 ms | 1,398 ms | 156 ms | 157 ms |
| three benchmarks concurrently | 2,603 ms | 2,686 ms | 156 ms | 159 ms |

Two things fall out. The p95/p50 ratio is 1.02–1.04 whenever nothing stalls, which is why 3,000 ms *looked* like 3x headroom — against an idle single run it was. And under mild contention the **median** already reaches 2,603 ms, so that ceiling had 15% left, not 3x. Seeded from the quiet run, it was one busy runner away from a red nightly nobody would trust.

Raising the sample count is not the escape: n ≥ 21 merely stops p95 being the maximum, ~100 gives it resolution, and one iteration costs 8–10 s — 100 would be ~17 minutes inside a 30-minute job that also seeds 10,000 files and runs the reliability benchmark.

So the enforced ms budgets grade the **median**, which five samples do support and which absorbs two stalled iterations. The cold ceiling is **5,000 ms**: 1.9x the slowest median measured, and still under the 6,411–6,900 ms the benchmark's own full-transcript-parse control costs, so a cold scan that regressed all the way back to parsing every transcript still trips it. Warm stays **750 ms** — its median moved 156 → 156 ms under the same contention. The p95 is still measured, reported and baseline-compared; it just no longer decides the run.

The guards enforce the rule rather than this instance: every enforced ms budget must be a `.p50-ms` metric labelled "median", and `iterations` joins `fileCount`/`transcriptBytes` in the pinned workload — a median is only robust while five samples sit under it, and editing the profile to `iterations: 1` restores the noise-sensitive number with every other guard blind to it (verified: 20 tests pass on that edit without the pin).

### 2. The last `source` asserted against nothing (`198140595`)

`cli.doctor.p95-ms` recorded 6,000 ms and explained itself in prose: *"matches `EXEC_TIMEOUT_MS` in `src/app/api/coven/exec/route.ts`"*. The constant is real and the number did match, but nothing checked — moving `EXEC_TIMEOUT_MS` would have left the catalogue asserting a stale ceiling while still reading like a live delegation. That is the 900 KB / 2800 KB drift class exactly, one gate short of being caught.

`pending` sources may now be prose only when nothing in this repository owns the number (the shell-boot and stream deadlines qualify). A source naming a repository file takes the same resolvable `<file> (<CONSTANT>)` form as a `postbuild` entry, the unit suite reads the declaration and asserts the limit equals it, and prose that merely *mentions* a `.ts`/`.mjs` path is refused so the checkable case cannot be reopened by rewording. Verified by mutation both ways: `EXEC_TIMEOUT_MS = 9_000` fails with the drift message; restoring the old prose fails with the prose message.

### 3. The new doc broke a wired unit suite (`2e1156866`)

`scripts/docs-index.test.mjs` pins `docs/README.md` as a complete classification of every top-level `docs/*.md`, and it is wired into the `app` suite a frontend-touching PR runs. `docs/performance-budgets.md` landed without an index entry, so the suite failed with `docs/*.md missing from the docs/README.md index: performance-budgets.md`. Indexed under Living / Working in this repo.

## Deliberately not changed

- **The `bytes` and cache-hit budgets keep p95-free absolute limits.** They are deterministic functions of the fixture, not of the machine — 41.59 MiB scanned on every run — so contention does not move them and there is nothing to make robust.
- **The three remaining `pending` limits stay prose-sourced.** Nothing in the repository declares a warm-boot, offline-boot or stream input-blocking deadline; inventing a constant to point at would be the dangling pointer this round just removed.
- **`percentile()` itself is unchanged.** Nearest-rank p95 of five samples legitimately *is* the maximum; the defect was budgeting it, not computing it. The benchmark now says so at the definition, so the next reader does not have to re-derive it.
- Re-litigation of already-rejected items: fixture cost in CI, blank-env fallback, Windows hermeticity, baseline backward compatibility, `run-tests.mjs` conflict risk, evaluator edge inputs, the dynamic `import()` in the delegated test, markdown pipe-escaping, and the override stamp's behaviour on the nightly path.

## Verification

- `node --experimental-strip-types --test src/lib/performance-budgets.test.ts scripts/cave-performance-report.test.mjs scripts/docs-index.test.mjs` — **35 pass, 0 fail**.
- `pnpm exec eslint src/lib/performance-budgets.ts src/lib/performance-budgets.test.ts --max-warnings=0` — clean.
- Full `pnpm performance:report` on `phase-6-list-10k`, end to end: `status: pass`, 0 breaches, 0 unmeasured, exit 0; 5 enforced, 2 delegated, 4 pending. Cold median 1,301.65 ms (74% headroom), warm median 139.30 ms (81%), 41.59 MiB against 64 MiB, 100% hit rate. Ten accumulated guards still let a healthy run through.
- Differential against pristine `origin/main` (`b5f40e54d`, detached): `pnpm typecheck` and `pnpm check:tests-wired` produce **byte-identical output** on base and head. The 7 `chat-view.tsx` typecheck errors and the 8 unwired streaming tests are inherited from merge `5771a5558` and are owned by `fix/cave-20wrn-restore-streaming-merge`; nothing here touches them.

PR description updated to match the implementation. Draft until `main` is green.

CompleteDotTech:
## Review round 5 — the differential, done gate by gate, and one claim it falsified

Reviewed at `2e1156866`; head is now `5f1241013`. Still a draft.

### The named job: does this branch add a failure to any gate?

**No.** Rounds 1–3 reported "byte-identical" on `typecheck` and `check:tests-wired` while this branch was in fact failing `scripts/docs-index.test.mjs` for three rounds, so this round compared *every* gate the required `Frontend build` job runs, on this head and on a detached pristine `origin/main` (`b5f40e54d`) in the same worktree:

| gate | base | head | differential |
| --- | --- | --- | --- |
| `lint` | pass | pass | output byte-identical |
| `typecheck` | fail, 7 `chat-view.tsx` errors | fail, same 7 | output byte-identical |
| `check:tests-wired` | fail, 8 unwired | fail, same 8 | output byte-identical |
| `test:conformance` | pass, 14 files | pass, 14 files | same verdict |
| `test:app` | fail | fail | both stop at `scripts/check-conflict-markers.test.mjs`, 55 passed before it |
| `test:api` | fail | fail | both stop at `scripts/dev-app.test.mjs`, 3 passed before it |
| `test:mobile` | fail | fail | both stop at `src/components/chat-view-mobile-command-center.test.ts`, 80 passed before it |
| `scripts/docs-index.test.mjs` | ok, 50 docs | ok, 51 docs | the new doc is indexed |
| `scripts/ci-paths.test.mjs` | pass | pass | same verdict |
| `scripts/check-conflict-markers.mjs` | pass | pass | output byte-identical |

`test:app` short-circuits on its first failure, so on this machine neither side gets past file 56 of ~1,274. That gap is closed separately: the **53 wired suites that walk the repository tree** — everything matching `readdirSync|readdir(|withFileTypes` under `scripts/*.test.mjs` and `src/**/*.test.ts`, which are the only tests a newly added file can newly break — were run file-by-file on both sides. One verdict changed: `src/lib/performance-budgets.test.ts`, `ABSENT` on base and `0` here. `scripts/docs-index.test.mjs` is in that set, which is how last round's defect would have been caught.

**Not compared:** `pnpm build` (`frontend-bundle`) and Playwright. Neither is affected — nothing in `src/app` or `src/components` imports `src/lib/performance-budgets.ts`, so it is not in any bundle graph.

### Fixed

**`19f7bb0be` — the cold ceiling claimed a guarantee it does not have.** Round 4's reseeding justified 5,000 ms with an upper anchor: it sits below the 6,411–6,900 ms the benchmark's own full-transcript-parse control loop cost on the seeding runs, so *"a cold scan that regressed all the way back to parsing every transcript must still trip this."* That anchor is not machine-independent — the control loop costs whatever the machine costs. Measured, not argued, on the same reference machine while quieter:

```
cold metadata scan median      922.92 ms
full-transcript-parse median  3891.97 ms   <- below the 5,000 ms ceiling
cold scan bytes            43,608,890
full-parse bytes           43,608,890      <- identical
```

So a cold scan back at full-parse speed reads ~3,892 ms and **passes**. Nothing else closes it: `cold-scan.bytes` sees the same 43.6 MB in both loops, because the optimisation removed parsing rather than reading; `cache-hit-rate` is taken from the warm loop, so a cold regression does not move it; and the 20% baseline comparison does classify it as a `regression`, but the nightly runs without `--fail-on-regression`, so the run still exits 0 — status `regression`, `budgetPass: true`, green.

Tightening is not the repair. The slowest legitimate median measured is 2,603 ms (three concurrent benchmarks), so any limit under 3,892 ms leaves 25–34% headroom and reinstates exactly the noise-driven red nightly that moving off p95 existed to end. The durable fix is a budget expressed *relative* to the control loop measured in the same run, which needs a second budget kind in the catalogue — filed as **`cave-4e1`**. The comment and the doc now state the gap instead of denying it, because a false guarantee is worse than a documented one.

**`5f1241013` — two restatements the median migration left behind.** `budgetFixtureProfile()`'s doc still illustrated what a profile name buys with `"10k conversation list, cold scan p95 ≤ 3,000 ms"` — a budget id and a limit that exist nowhere in the catalogue that function serves. And the Fixtures section said the unit suite pins `fileCount` and `transcriptBytes`, while the section 70 lines above it says `iterations` is pinned too, so the doc disagreed with itself about what a fixture edit costs. Every other `3,000 ms` left in the tree is explicitly historical ("the original", "the old", "previously reported") and is left as written.

### Reviewed and found sound

Round 4's own new code, which nobody had reviewed:

- **The median migration is internally consistent.** `conversation-list.{cold-scan,warm-cache}.p50-ms` are emitted by `conversationMetrics()` under exactly those ids; both labels say "median"; `SLOWEST_MEASURED_RUN` was moved to the `.p50-ms` ids with p50 values, and still asserts something — `unmeasuredCount: 0` requires every id to resolve, and 2,603.31 ms against the 5,000 ms ceiling still fails if anyone tightens the limit past the measurement it was seeded from.
- **The `.p50-ms` guard.** Its tripwire keys the sample count off `SEEDED_WORKLOAD.iterations`, which is itself pinned to the fixture, so raising `iterations` to 100 makes `percentile(sample, 0.95) !== max(sample)` and the test fails with the message that says a p95 budget is defensible again. Verified arithmetically: nearest-rank p95 is the maximum for n ≤ 20 and stops being it at n = 21.
- **The `iterations` pin.** Deliberately a literal rather than read from the fixture — the opposite choice from `8ee6b846e`, and correct: its job is to pin the fixture against change, while the sweep test's job is to test a property of it.
- **`EXEC_TIMEOUT_MS` resolution.** `src/app/api/coven/exec/route.ts` declares `const EXEC_TIMEOUT_MS = 6_000;`, the test's declaration regex matches it, `6_000` normalises to `6000`, and the catalogue's limit is `6000`. Read as text, not imported, so no Next runtime enters the unit suite.

### Verification

- `node --experimental-strip-types --test src/lib/performance-budgets.test.ts scripts/cave-performance-report.test.mjs` — **34 pass, 0 fail**.
- `pnpm exec eslint` clean on `performance-budgets.ts`, `performance-budgets.test.ts`, `cave-performance-report.mjs`, `conversation-list-benchmark.mjs`.
- `node scripts/docs-index.test.mjs` — ok, 51 documents classified, 64 links resolved.
- `pnpm performance:report` end to end on `phase-6-list-10k`: `status: pass`, 0 breaches, 0 unmeasured, exit 0; 5 enforced, 2 delegated, 4 pending. Cold median 922.92 ms against 5,000 ms (81.5% headroom), warm median 109.09 ms against 750 ms (85.5%), 41.59 MiB against 64 MiB, 100% cache hit rate. Artifacts deleted afterwards.
- `pnpm typecheck` produces no error naming any file this PR touches.

### Deliberately left alone

The `chat-view.tsx` / streaming failures on `main` (owned by `fix/cave-20wrn-restore-streaming-merge`), and everything rounds 1–4 rejected with executable evidence.

CompleteDotTech:
## Review round 6 — head `51b51c1cc`

Four commits. Both beads this round was scoped to are closed; one new one is filed.

### Landed

**`cave-4e1` — the flagship budget could not catch its own regression** (`2d14b0118`, `ce340bc6a`).
The catalogue grew a relative kind: `ratioOf` names a numerator and a denominator metric and the value is graded as a percentage, so `conversation-list.cold-scan.share-of-full-parse-pct` divides the cold median by the control median measured in the *same run*. A ratio cannot be defeated by a fast machine the way the 5,000 ms anchor was.

Two things turned up while building it, both by execution:

1. **The mechanism the comments described was wrong.** The cold scan does not skip parsing — `readConversationSummary` reads each file whole and `JSON.parse`s it exactly as the control loop does. The shipped win is the 8-way read pool plus the stat-keyed summary cache. "Regressed back to parsing every transcript" was never a describable regression, and `cold-scan.bytes` cannot separate the loops for a simpler reason than the one given: both read all 43,608,890 bytes and always did. Corrected in the module, the doc and the PR body.
2. **The ratio is platform-sensitive, not just load-sensitive.** Measured cold/control medians — windows idle 24.28%, windows 3-concurrent 38.34%, **linux (wsl2) idle 63.12%**, linux 3-concurrent 52.59%. Collapsed read pool (`CONVERSATION_LIST_READ_CONCURRENCY=1`): windows 112.01%, linux 127.30%. The first draft shipped at 75% seeded on Windows alone, which would have left the nightly's own platform 12 points from red. Re-seeded to **90%**, mid-band between the worst healthy and the nearest collapse.

Note the collapse experiment is the real point: a cold scan back at naive-sequential cost measures 4,397.85 ms on Windows and 3,460.77 ms on Linux — both **pass** the 5,000 ms ceiling, with identical bytes and an untouched warm-loop hit rate. The unit suite pins the new ceiling from both sides (every measured healthy pair passes, every collapsed pair breaches); verified by moving the limit — 50 reddens the Linux idle run, 120 greens the Windows collapse.

**`cave-21w` — the catalogue was silently partial** (`e62116259`).
It recorded 2 of the 9 build-gate numbers, so the bundle surface read as covered with four of its five ceilings absent. All nine are recorded now, and the completeness check no longer asks the catalogue about itself: it reads the constants out of `bundle-budget.mjs` (by the exact `Number(process.env.X)` shape its knobs use, since that script exits at import time) and out of the two frozen `*_BUDGETS` objects, and requires an entry for each — with the gate list cross-checked against the files the catalogue delegates to. Verified by deleting an entry: the suite fails naming the constant it lost. Two of the nine are file counts, so the `ms | bytes | percent` unit union grew a `count` rather than the entries being dropped with an excuse.

**Linux rows for the absolute ceilings** (`51b51c1cc`). Every seeding row for the cold and warm ms ceilings came from a Windows box. Measured on Linux: cold 1,760 of 5,000, warm 198 of 750. Both hold, so neither number moves — recorded because a limit justified only on a platform CI never runs is a claim nobody checked.

### Filed, not fixed

- **`cave-wxq`** — re-seed the 90% ratio ceiling from a real `ubuntu-24.04` nightly. Every number behind it was measured on this machine, and WSL2 is Linux syscalls over a VM disk: the right platform family, not the right machine.

### Deliberately not done

- **Not merged/rebased onto current `main`.** `main` has been repaired since this branch's base — at `b9856a7ed`, `typecheck`, `check:tests-wired` and `test:mobile` all pass, while this head still inherits their failures from `b5f40e54d`. Bringing that in is a merge, not a review repair, and the PR is waiting on `main` anyway.
- **Nothing in `chat-view.tsx` or the streaming tests.**
- **`cold-scan.bytes` left as it is.** It is close to tautological while the fixture scale is pinned, but it still catches a scan that reads each file more than once, which is a real class.

### Differential

Every gate run at the merge-base `b5f40e54d` and at this head; exit codes identical, and after normalising timestamps and durations `lint`, `typecheck`, `check:tests-wired`, `test:conformance` and `test:api` are byte-identical. `test:app` differs only in `1273 → 1274 test file(s)` and stops at the same file with the same count before it; `test:mobile` differs only in timings and stops at the same file. `scripts/docs-index.test.mjs` and `scripts/ci-paths.test.mjs`: 10 tests, 0 fail.

Direct: `src/lib/performance-budgets.test.ts` 29 pass, `scripts/cave-performance-report.test.mjs` 15 pass, ESLint clean, and a full `pnpm performance:report` on `phase-6-list-10k` at `status: pass` — 6 enforced, 9 delegated, 4 pending, the relative budget reading 25.22% against 90%.

CompleteDotTech:
## Round 7 review — head `f7e354c51`

No code change. Six earlier rounds each turned the previous round's new code into the next round's finding; this round did not, and the checks below are why I am calling that a real result rather than a gap in the search.

### The merge is non-destructive

`f7e354c51` merges `origin/main` at `b9856a7ed`. `git diff origin/main HEAD` and `git diff origin/main...HEAD` are the **same ten files** — the check that catches both halves of a bad merge: nothing of `main`'s was reverted, and nothing of this branch's was swallowed. `main` touched `scripts/run-tests.mjs` in the same window (+10 lines); both of this branch's lines in that file survive.

### Every gate run for real, not assumed

CI on this exact head: typecheck, lint, test wiring, **app tests**, **API tests**, mobile tests, protocol conformance, Frontend bundle and Select validation all pass. Only `Frontend build` is still reporting.

Locally on the same head: `pnpm typecheck` 0, `pnpm lint` 0, `pnpm check:tests-wired` 1758/1758, `pnpm test:mobile` pass, `scripts/docs-index.test.mjs` pass, `performance-budgets.test.ts` 29/29, `cave-performance-report.test.mjs` 15/15, and a full `pnpm performance:report` on `phase-6-list-10k`: `status: pass`, 0 breaches, 0 unmeasured, exit 0 — cold median 1,062.92 ms of 5,000, relative budget 23.14% of 90, warm median 117.62 ms of 750, 41.59 MiB of 64, 100% hit rate.

### Round 6's code was mutation-tested, not read

The guards it added are the ones nobody had reviewed, so I broke each and checked the message names the loss:

| mutation | result |
| --- | --- |
| delete the `bundle.shell.bytes` entry | *"scripts/bundle-budget.mjs enforces BUNDLE_MAX_SHELL_KB, but no catalogue entry records it"* |
| relative ceiling 90 → 50 | *"linux (wsl2), idle measured 63.12%, over the shipped ceiling — a healthy run must not be red"* |
| relative ceiling 90 → 120 | *"windows, read pool 1 measured 112.01% and passed — the ceiling has been loosened past the regression it exists to catch"* |

Also re-derived rather than trusted: the delegated set really is 9 (5 `Number(process.env.X)` knobs in `bundle-budget.mjs`, `STANDALONE_BUDGETS.{fileCount,unpackedBytes}`, `SIDECAR_RUNTIME_BUDGETS.{fileCount,unpackedBytes}`), and `SIDECAR_RUNTIME_BUDGETS.fileCount` really is `readSidecarFileCountBudget()` reading `scripts/sidecar-runtime-budget.json`, so the entry's comment about naming the reader is accurate. `caveHome()` reads `COVEN_CAVE_HOME` verbatim with no `realpath` on either side of the benchmark's `CONV_DIR` guard, so no symlink can defeat it and the pinned env overwrites any pre-set value before `cave-conversations.ts` is imported. An old baseline `report.json` is only read for `.metrics`, `.history`, `.sha`, `.runId`, `.generatedAt` — never `.budgets` — so it cannot throw.

### What I changed

**The PR description only.** Its premise section had gone stale: it told a reviewer the branch was based on `b5f40e54d`, that `main` was red, and presented a differential table asserting *this head* fails `typecheck`, `check:tests-wired`, `test:app`, `test:api` and `test:mobile`. All five of those are now false, and the section's practical effect was to pre-excuse any red as inherited — which is exactly how a real failure gets waved through. Replaced with the actual base, the green CI table, and the two remaining local failures with their causes.

### Deliberately not fixed

- `scripts/check-conflict-markers.test.mjs` fails on Windows because the CLI is gated on ``import.meta.url === `file://${process.argv[1]}` ``, which never matches there (`file:///C:/…` against `C:\…`) — so a repo-wide gate exits 0 on a tree full of markers, and `pnpm test:app` cannot complete locally. Linux CI is unaffected. Present since #4434, untouched by this PR; filed as `cave-zya`.
- `scripts/dev-app.test.mjs` fails locally because it spawns `node` through `/bin/bash`, which has no `node` on PATH in this environment. Passes on CI.

Those two are just where each runner stops. Driving the same file lists past them with a continue-on-failure harness surfaces further Windows-only failures in unrelated files — `chat-project-access.test.ts` (`EPERM` on `symlinkSync`), `api/beads/route.test.ts` (needs `wsl.exe` and a beads database), `release-notes.test.mjs` — none of which this PR touches, and every one of them green in the Linux `app tests` / `API tests` jobs above.

### One observation, not a defect

`budgetFixtureProfile()` returns `null` for a `performance-report` source with no `#`, and the evaluator then skips the fixture-scale check entirely — an unidentified *run* fails closed, but an unidentified *budget* is graded unconditionally. Unreachable in the shipped catalogue: `"every budget enforced here names the fixture that produces it"` rejects any enforced source that is not `fixtures/phase-6/performance-fixtures.json#<existing profile>`, and the test helper relies on the current behaviour to exercise ceiling semantics without a fixture. Noting it so the next entry added is checked against it rather than rediscovered.

Issue #4781, opened 2026-08-21
[Chat v1 P7] Execute OS acceptance, staged rollout, and rollback drill


CompleteDotTech:
# [Chat v1 P7] Execute OS acceptance, staged rollout, and rollback drill

Run the complete no-source-checkout acceptance journey on all supported operating systems, then stage rollout with ready rollback metadata and prior artifacts.

Owner repository: OpenCoven/coven-cave (program tracker)
Execution lane: release-acceptance
Plan of record: OpenCoven/chat/docs/superpowers/plans/2026-08-15-phase-7-release-rollout.md

Deliverables:
- Install/pair/read/send/resume/restart/attachment/action/revoke/update acceptance on macOS, Windows, and Linux.
- Global opencoven doctor/pair/session/send/tail/scaffold acceptance.
- Low-percentage rollout, health thresholds, pause criteria, and rollback drill.

Acceptance criteria:
- The full release acceptance journey passes on all supported OSes.
- Prior stable artifacts and rollback metadata are verified before rollout.
- Crash, auth, duplicate-send, and data-integrity regressions stop rollout.

## Bead metadata

- **Bead ID:** `cave-udcn7`

- **Status:** open

- **Priority:** P0

- **Surface:** Shared

- **Authoritative owner:** Unassigned

- **Labels:** `area:rollout`, `needs-human`, `phase:7`, `program:chat-v1`, `release-blocker`, `repo:cross-repo`, `surface:shared`, `verification-required`

- **Parent:** None

- **Dependencies:** `cave-as76u`, `cave-j65ie`

## Source of truth

Beads is authoritative. This draft card is a GitHub Teamwork visibility mirror and should be reconciled from the Bead rather than edited as an independent task.

BunsDev:
Closing this accidental repository-issue conversion. The canonical task remains the Bead mirrored as a draft card on the Teamwork project; Beads remains the source of truth.

CompleteDotTech:
Reopened: this is unblocked and the remaining work is real.

[...]

PR #4789, opened 2026-08-21
Gate staged rollout on acceptance evidence and a bounded rollback drill (cave-udcn7)

CompleteDotTech:
Advances Phase 7 Task 12–13 for bead `cave-udcn7` (#4781): three-OS acceptance,
staged rollout, and the rollback drill.

The acceptance journey itself is manual by design — it installs published,
signed artifacts on real machines with no developer tooling and no source
checkout, which is exactly what a CI runner cannot stand in for. So what lands
here is the machinery that makes the human execution *checkable* and the
rollout decision *reproducible*, rather than a claim in a release issue.

## What lands

**`scripts/release-acceptance.mjs`** — validates the evidence record an
operator produces:

- all three operating systems present exactly once,
- every one of the twelve journey steps and seven global-CLI steps
  (`doctor`/`pair`/`session`/`send`/`tail`/`scaffold`) recorded,
- candidate version/tag/commit and artifact SHA-256 digests well formed,
- no credential-shaped text committed alongside the evidence, named by JSON
  path so it can be redacted.

Steps carry four results, and the split is load-bearing: `pending` is a gap in
coverage and owes nothing, while `blocked` and `fail` are outcomes somebody
observed and must name a diagnostic. Without that distinction a fresh template
emitted 57 spurious "needs a diagnosticId" errors, which buries the real gaps.

**`scripts/release-rollout.mjs`** — turns a rollout state file into exactly one
of `advance`, `hold`, or `rollback`:

- a **hard-stop class** (crash, auth, duplicate-send, data-integrity) is
  user-visible or data-affecting and sends the release back — this is the
  issue's "crash, auth, duplicate-send, and data-integrity regressions stop
  rollout", implemented as both reported regressions and threshold breaches
  mapped onto the class they belong to;
- an unmet observation window, a failing functional canary, acceptance that is
  merely incomplete, or a metric nobody measured **holds** the current stage;
- absent data never advances a rollout and never triggers a rollback.

Five stages (`maintainer` → `private-beta` → `stable-5` → `stable-25` →
`stable-100`); the two manual stages sit at 0% so no broad automatic update
exists before the stable stages.

It also carries the **drill**: a three-step restore plan of which exactly one
step mutates anything (republishing `latest.json`), plus a guard that refuses
any plan reaching for a `tag-move`, `artifact-overwrite`, `version-unpublish`,
or `signature-regeneration`.

**Runbooks** — `docs/workflows/release-acceptance.md` and
`docs/workflows/production-rollout.md`, linked from the existing release
procedure in `docs/workflows/branching.md`.

## Scope, and how it composes with #4782

Deliberately narrow: **"is the prior release a usable rollback target?" is not
answered here.** That belongs to #4782 (`cave-ilh1h`), which resolves the
baseline against live release history. This gate consumes its verdict as
`state.rollbackReadiness` and **holds when the verdict is missing** — silence
is not proof of a way back.

An earlier revision of this branch did re-derive rollback readiness; that
duplication was removed once #4782 was found, so the two compose rather than
compete. The remaining overlap is two additive lines in `scripts/run-tests.mjs`
and one in `docs/workflows/branching.md`.

## Validation

```
node --test scripts/release-acceptance.test.mjs   # 21 pass
node --test scripts/release-rollout.test.mjs      # 35 pass
node --require ./scripts/css-source-contract-hook.cjs scripts/release-{acceptance,rollout}.test.mjs   # exit 0, the runner's own invocation
node scripts/docs-index.test.mjs                  # index ok
pnpm lint                                         # clean
```

Both tests are wired into `SUITES.api`; `check-tests-wired` reports no unwired
file from this branch.

## ⚠️ Draft because `main` is red, not because this change is unfinished

`origin/main` (`b5f40e54d`) fails CI on its own, so the required `Frontend
build` context cannot go green for any PR branched from it. Reproduced in this
worktree before any change here, and identically in an unrelated checkout:

- `pnpm typecheck` — `src/components/chat-view.tsx` references an undefined
  `bubbleSegments` (9001, 9051, 9069) and passes a `durationMs` prop that does
  not exist.
- `pnpm check:tests-wired` — 8 test files unwired (`chat-result-markers`,
  `chat-tool-verification`, `streaming-markdown-blocks`,
  `streaming-turn-view-model`, `use-stick-to-bottom`, `chat-assistant-output`,
  `streaming-chat-wiring`, `streaming-turn-response`).

Both come from the `feat/cave-20wrn-calm-streaming-chat` merge that reached
`main` directly. #4784 is the repair; #4782 documents the same blockage
independently. Mark ready once `main` is green.

Separately, `scripts/worktree-guard.test.mjs` fails locally on Windows in this
checkout *and* in an unrelated one with these changes absent, and its files are
byte-identical to `main` — environmental, not from this branch.


CompleteDotTech:
## CI triage for `349f21d24` — run [32457208663](https://github.com/OpenCoven/coven-cave/actions/runs/32457208663)

Confirmed: **every failure is inherited from a red `main`. No failed log
mentions a file from this branch** (`grep -ic 'release-acceptance|release-rollout'`
over `--log-failed` returns 0).

| Check | Fails on | Owner |
| --- | --- | --- |
| Frontend validation (typecheck) | `src/components/chat-view.tsx` — undefined `bubbleSegments`, unused `ChatToolActivityLayout`, bad `durationMs` prop | #4784 |
| Frontend validation (test wiring) | the same 8 unwired `streaming-*` / `chat-*` files | #4784 |
| Frontend validation (app tests) | `src/components/chat-view-mobile-command-center.test.ts` | #4784 |
| Frontend validation (API tests) | `src/app/api/chat/send/chat-attention-persistence.test.ts` | pre-existing |
| Frontend validation (mobile tests) | `src/components/artifact-comments.test.ts` | pre-existing |
| Frontend bundle | same `chat-view.tsx` type error | #4784 |
| iOS build | `apps/ios/.../AppModel.swift:1730` Swift syntax error | pre-existing |
| **lint**, **protocol conformance**, **Select validation** | — | **pass** |

### One consequence worth stating plainly

`scripts/run-tests.mjs` runs each suite sequentially and exits on the first
failure. `chat-attention-persistence.test.ts` sits at line 1458 of `SUITES.api`;
this branch's two test files sit at 1614–1615. **So neither of them was reached
in this run** — they are verified locally only, including under the runner's own
invocation:

```
node --require ./scripts/css-source-contract-hook.cjs scripts/release-acceptance.test.mjs   # exit 0, 11 pass
node --require ./scripts/css-source-contract-hook.cjs scripts/release-rollout.test.mjs      # exit 0, 16 pass
```

They will not have run in CI until the API suite is green again. Re-run this
run and re-read this table before marking the PR ready.

CompleteDotTech:
## Review of `349f21d24` → fixes through `4aa65bc44`

Six fixes, each pushed as its own commit. Every finding below was reproduced
before it was fixed.

### The #4782 integration defect (`1bdb82d10`)

The gate consumed a rollback-readiness verdict `scripts/release-rollback-readiness.mjs`
(#4782) does not emit. That gate resolves to `{ tag, version, baseline, baselineWaived, platforms, ready }`
and throws `RollbackReadinessError` on every shortfall, so two of the four fields
read here never existed:

- `.baselineVersion` → `undefined`, so `rollbackTarget` came back `null` on a
  real rollback. Reproduced against #4782's actual shape:
  `rollback | target: null`. The target is `baseline.version`; the verdict's
  top-level `version` is the release being *shipped*, so reading that instead
  would have named the regression as its own remedy.
- `.blockers` → nothing, because that gate throws rather than returning a list.
  The hold reason now quotes `error`, the message an operator transcribes.
- `baselineWaived: true` / `baseline: null` is now decided rather than left to
  `undefined`: it does **not** hold, because `--allow-missing-baseline` is
  already an explicit upstream decision and re-litigating it would make a
  repository's first release un-rolloutable. Every report states there is no
  rollback target and the remedy is patching forward, and `restore-plan`
  refuses. `ready: true` with neither a baseline nor the waiver is a shape that
  producer cannot emit, so it holds.

The adapter belongs on this side: #4782 produces a general-purpose verdict, and
this gate is its only consumer. No comment left there — nothing needs to change.
The test fixture is now that producer's resolved object copied field for field.

### `Number(null) === 0` (`07beda66a`)

One coercion broke the absent-data rule in both directions at once:

```
crashFreeLaunchRate: null  -> rollback ("0 is below the 0.995 threshold")
duplicateSendCount:  null  -> advance  (0 duplicates, threshold cleared)
```

`""`, `[]` and `false` behave identically, and `"0.99"` was accepted as a rate.
Range is now checked too: a crash-free rate of `42` cleared every minimum and
`-5` duplicate sends cleared every maximum, both advancing silently. Also
closed the gate's only fail-**open** path — `regressions` was coerced with
`Array.isArray(v) ? v : []`, so `"regressions": {…}` dropped every regression
inside it and advanced.

### Threshold overrides deleted checks (`5f1f60921`)

`{ ...DEFAULT_THRESHOLDS, ...state.thresholds }` let a state file remove a
threshold rather than relax it. `{"minCrashFreeLaunchRate": null}` made the
comparison `0.1 < null` — false — so a 10% crash-free launch rate **advanced**
with no reason printed. `"lots"` disabled a maximum the same way past 900
duplicate sends.

### Acceptance secret scan (`a335d3127`)

`findSecrets` was recursive and overflowed the stack at ~5,000 levels of
nesting while `JSON.parse` accepts 100,000, so a deep record threw
`RangeError` from a function contracted to return every problem it finds. Now
iterative, with a `seen` set so a cyclic object handed to the exported function
terminates. Added the three credential shapes this record invites — the journey
pairs a client at step four and pairs the CLI again, so a pasted pairing
diagnostic is an `Authorization: Bearer` header, a `?token=` callback url, or a
Slack token. All anchored on literal prefixes, asserted directly against
`findSecrets` rather than through "the record validates clean", which would
also pass against a scanner that never ran.

### CLI options and doc drift (`658615488`), observation window (`4aa65bc44`)

`gate --dry-run state.json` ran the real gate and said nothing about the flag.
Neither CLI has options; both now say so. `production-rollout.md` promised
overridable thresholds without naming the four keys, and `release-acceptance.md`
named the OSes only as `macOS`/`Windows`/`Linux` while the validator accepts
only lowercase. And `observedHours: null` cleared `stable-100` outright, whose
required window is 0.

### Deliberately not changed

- **`pairingSuccessRate` → `auth`.** Correct as written. A pairing *rate* below
  98% means a large share of users cannot authenticate to their own daemon with
  no workaround; the pass/fail pairing canary is the functional check that
  merely holds. Code and docs agree.
- **`acceptance` outside `HARD_STOP_CLASSES`.** Coherent. That constant is the
  vocabulary for classifying an operator-reported regression in
  `state.regressions`; a reason's `class` is a label, not a membership test.
  Comment and test added rather than a change.
- **`complete` requiring `errors.length === 0`.** Correct — the record is a
  claim about which bytes were accepted, and one that cannot say which has not
  made it. Documented explicitly instead.
- **Everything in the §3 inherited table.** Re-verified rather than trusted:
  this branch touches no file under `src/`, and the eight unwired test files are
  present on `origin/main`. Owned by #4784.

### Verification

`node --test` both files: **41 pass, 0 fail**. Under the runner's own
invocation (`nodeArgsFor` resolves to `--require ./scripts/css-source-contract-hook.cjs`
for both): 14 + 27 pass. `docs-index.test.mjs` ok. `check-tests-wired.mjs`
reports exactly the 8 inherited files. `pnpm lint` clean.

**CI is not green and cannot be** until #4784 lands. Worse, these tests have
never executed in CI at all: `run-tests.mjs` returns on the first failing file,
`chat-attention-persistence.test.ts` sits at line 1458 of `SUITES.api`, and
these two sit at 1614–1615. Local runs are the only evidence that exists.
Staying a draft.

CompleteDotTech:
## Second-pass review — 4aa65bc44 → d2c6626c9

Round 2 audited each of the six round-1 fixes for correctness and completeness,
then looked past them. Five changes landed; all five are in `scripts/release-*`
and their docs — nothing outside this PR's surface was touched.

### Fixed

- **549a753d2 — a word wrapped in an array read as the word.** Round 1 stopped
  coercing the *numeric* half of both files (`Number(null) === 0`) and left the
  *word* half coercing. `String(["complete"]) === "complete"`, so
  `"acceptance": {"status": ["complete"]}` **advanced a rollout**, and a canary
  recorded as `["pass"]` counted green. Same trap in the validator: a record
  whose `os`, `result`, `candidate.version/tag/commit` and `artifacts[].sha256`
  were each wrapped that way validated `complete` with no step result recorded
  anywhere in it. That is the shape a generated state file or a `jq '[.x]'`
  produces, which is the case these scripts exist for. Both now require an
  actual string; the gate distinguishes `unreadable` from `missing`, and the
  validator prints the value as it sits in the file rather than the coerced
  form that looks correct.
- **d4f11f6a0 — `restore-plan` refused only the rarer half of its own rule.**
  Round 1 refused a *waived* baseline on the grounds that a placeholder drill
  rehearses a procedure nobody can run. A missing, empty, not-ready, or
  truncated `rollbackReadiness` still printed the three-step drill against the
  literal string `<baseline-tag>` — the same defect, in the far commoner shape.
  Now refused too, quoting the transcribed upstream message. A baseline that is
  *named but unproven* still prints: the operator knows which release they
  would put back, which is what a rehearsal is for.
- **4731b085d — a waived baseline printed two `rollback target:` lines.** A
  rollback took `describeRollbackTarget`'s line *and* the standing waiver note
  added for advance/hold, with different wording. In an incident report two
  lines sharing a prefix read as two findings.
- **cdb234f02 — the observation-window rule was enforced but undocumented**, and
  `observedHours: -5` reported "is not recorded". `stable-100` requires zero
  hours, so this is the one field where the strict reading decides whether the
  whole install base gets the update; `production-rollout.md` described it for
  the metrics only. Negative is now its own message.
- **d2c6626c9 — kept `validateAcceptanceRecord` total.** The new messages print
  the offending value and `JSON.stringify` throws on a cycle; the function
  documents itself as returning every problem rather than throwing one. Same
  reasoning `findSecrets` already carries.

### Checked and left alone

- **Round-1 fix 1** composes correctly against #4782's real shape. Confirmed by
  reading `chore/cave-ilh1h-p7-release-gate`: `baselineWaived: true` ⟺
  `baseline: null`, `ready` is only ever `true` from the producer (it throws
  otherwise), and `baselineVersion`/`blockers` genuinely do not exist. The
  waiver decision is consistent across gate (advance), both reports, and
  `restore-plan` (refuse).
- **Round-1 fix 2's boundaries all behave.** `0.995` exactly → advance,
  `0.9949999` → rollback, `0` / `1e-7` / `-0` → rollback, `1` → advance,
  `maxDuplicateSends` at `-0` → advance and at `MAX_SAFE_INTEGER`/`1e21` →
  rollback. Pairing at exactly `0.98` advances.
- **`regressions[].class` still uses `String()`.** Deliberate: coercion there
  can only make the gate *stricter* (an unreadable class holds, a coerced
  hard-stop class rolls back). Tightening it would turn a would-be rollback
  into a hold.
- Template still round-trips; the documented state-file example still gates
  `advance`; `docs-index` resolves; `check-tests-wired` reports only the eight
  inherited files; `pnpm lint` clean. Nothing in this branch traces to a `main`
  CI failure — the §3 inherited set (#4784) is untouched.

### Open

`release-rollback-readiness.mjs` never emits the JSON object the state file
wants: its CLI writes `GITHUB_OUTPUT` key/value pairs (`ready`,
`baseline-tag`, `baseline-version`, `baseline-waived`, …) and a step summary.
`production-rollout.md` documents the *resolved* shape correctly, but neither
PR says how an operator gets from those outputs to the `rollbackReadiness`
object. That is a question for #4782's side, not a defect here.

Still a draft, still blocked on #4784 — the api suite exits on
`chat-attention-persistence.test.ts` at line 1458 of `SUITES.api`, well before
these two files at 1614–1615, so local runs remain the only evidence.

CompleteDotTech:
## Round 3 review — three commits, head `233596b69`

Audited rounds 1–2's eleven commits for defects they introduced, then enumerated **every** JSON field either script reads and drove it through a hostile-value matrix (`null`, `undefined`, `""`, whitespace, `[]`, `{}`, `[value]`, `"value"`, `0`, `-0`, `NaN`, `±Infinity`, booleans, nested objects, `__proto__`/`constructor` keys) asserting no combination yields `advance` / `complete` / `ok: true` / exit 0 that should not. 1527 assertions, run from a scratch script that is not committed.

### Fixed

- **`7e062dafa` — a coerced object read as a recorded version.** The round-2 `readString` sweep reached every field matched against a pattern or a vocabulary and stopped at the five checked only for emptiness. Those still read through `String(x ?? "")`, where `{}` is `"[object Object]"` and `true` is `"true"` — non-empty, satisfying *is required*. A record whose `osVersion`/`caveVersion`/`chatVersion`/`cliVersion` were each `{}` validated **`complete`**, claiming a three-OS journey against environments nobody wrote down; and a `fail` step satisfied its `diagnosticId` demand with an empty object. All five now go through `readString`. Runbook field list corrected — it enumerated exactly the fields that already had the rule.
- **`d0fa2ef49` — the drill rehearsed the placeholder the guard exists to prevent.** `restore-plan` admits a drill on `readRollbackReadiness().target`, which reads `baseline.version` **or** `baseline.tag`; `planRollbackRestore` printed `baseline.tag` alone, through `String()`. So the mirror of the tag-only case already pinned — `{"version": "0.9.4"}` — cleared round 2's guard and printed `confirm every asset <baseline-tag>'s manifest references…`, and a non-string tag printed as `[object Object]` or `42`. The printer now has its own reader over the same two fields (tag preferred, since that is what an operator types), so guard and drill are non-null on exactly the same baselines. The equivalence is pinned directly rather than through a distant CLI assertion.
- **`233596b69` — the acceptance↔rollout seam was untested.** Every gate fixture hand-writes `{ status: "complete" }`, which asserts the gate *reads* a field, not that anything produces it. Rename `status` in `summarizeAcceptance()` and both suites stay green while rollout reads `missing` on every real record — the same consumer-reads-a-field-the-producer-never-emits shape the readiness fixtures guard against for #4782, on the one seam this PR owns at both ends. Now driven end to end for all three outcomes.

### Checked and found sound

Prototype pollution: `__proto__`/`constructor`/`prototype` as threshold keys, stage ids, canary names, step ids, OS names and artifact names all fail closed, and `Object.prototype` is never written (`Object.hasOwn` on the defaults; fixed vocabularies indexed, never the input's keys). Every remaining rollout advance path refuses hostile input. Precedence `rollback` > `hold` > `advance` has no early return that can mask a hard stop. `structuredClone` is fine on the `>=24.18.0` engine floor. `findSecrets` is iterative with a `seen` guard. No overlap with `verify-release-updater.mjs` / `release-promotion.mjs` / `check-release-version.mjs`.

### Left alone

`check-tests-wired` reports exactly the 8 inherited `streaming-*`/`chat-*` files — #4784's, untouched. `main` is still red, so the required context cannot go green here; the PR stays a **draft**.

### Verified

`node --test` both files (48 pass) and again under the runner's own `--require ./scripts/css-source-contract-hook.cjs` invocation (17 + 31); `docs-index.test.mjs` — `index ok (50 documents classified, 63 links resolved)`; `pnpm lint` clean; `template 1.0.0` round-trips through `JSON.parse`; `release:rollout stages` prints the five stages; both CLIs print usage and exit 1 with no args.

CompleteDotTech:
## Review round 4 — convergence check

Reviewed `233596b69`, left `e5599a063`. Four commits. Round 4's stated purpose was to establish whether the loop had converged; it had not, though the remaining defects were in different places than rounds 1–3 were looking.

### Round 3's three fixes — audited, all sound

- **`7e062dafa`** (readString on `osVersion`/`caveVersion`/`chatVersion`/`cliVersion`/`diagnosticId`) — rejects nothing legitimate; a plain non-empty string still validates. The doc's field list in `release-acceptance.md` matches the code exactly.
- **`d0fa2ef49`** (`baselineName` for the printed drill) — guard/printer equivalence verified. Both `readRollbackReadiness().target` and `baselineName()` are `firstNonEmptyString` over the same `{version, tag}` of the same object, so they are non-null on exactly the same baselines, in every branch. The differing field *order* is deliberate and documented, and only changes which of two names is shown.
- **`233596b69`** (the acceptance/gate seam test) — confirmed genuinely binding. Renaming `status` to `state` in `summarizeAcceptance()` fails exactly that test, and only that test, with `'hold' !== 'advance'`.

### The three items round 3 left

1. **`validateArtifacts` uses `forEach`, skipping array holes — ratified as-is.** `findSecrets` got a cycle guard because it is *exported* and its failure mode on a cycle is a hang, which breaks its documented "always returns its problems" contract. `validateArtifacts` is module-private; its failure mode on a hole is one fewer reported error on an input `JSON.parse` cannot construct. Different reachability and different severity class — the principle is applied consistently, not inconsistently.
2. **`formatGateReport`'s waived line with a baseline present — fixed** (`9d46ccf9f`), see below.
3. **`show()`'s totality comment — ratified.** A null-prototype function does escape both the `??` and the `catch`. `show()` is module-private, every call site's argument descends from `JSON.parse`, and JSON cannot produce a function of any prototype. The comment's "total" is scoped by the sentence naming the two cases it exists for (cycles, BigInt), both of which it does handle.

### Fixed

**`ef9f476fa` — the drill's own verification step reports failure on a rollback that worked.** The highest-value finding, and an operator-facing one. `verify-release-updater.mjs` §3 asserts `latest.json`.version equals the tag of `/releases/latest`. Step 2 of this drill deliberately breaks that equality: it republishes the baseline manifest while the candidate release stays published and stays `latest`, because `tag-move` and `version-unpublish` are both on `FORBIDDEN_RESTORE_OPERATIONS`. The updater endpoint is `releases/latest/download/latest.json`, so a *correct* rollback produces exactly `✗ version drift: latest.json=0.9.4 vs release=v1.0.0` and exit 1. The step said only "the served manifest must now be the baseline version" — an operator who runs it mid-incident, sees `FAILURE(S)`, and concludes the restore did not take rolls the regression forward again. Both the plan step and the runbook now name the expected drift and separate it from the failures that mean the restore really did not take.

**`9d46ccf9f` — hold on a verdict that both waives the baseline and names one.** The gate already fails closed on `ready: true` with no baseline and no waiver, reasoning that #4782 cannot emit that pairing so the file was hand-edited. The mirror pairing is unproducible for the same reason (`baselineWaived: true` is returned only from the branch returning `baseline: null` — verified against `chore/cave-ilh1h-p7-release-gate`) and was not caught. It made one file give three answers to one question: an advancing report said "none — baseline waived; the only remedy is patching forward", a rollback on the same file printed `0.9.4`, and `restore-plan` refused to rehearse a baseline sitting in the same object. The first is the harmful one — it steers an operator away from an available restore. The runbook asks operators to hand-write `rollbackReadiness` for a not-ready verdict, so transcription is a documented path here.

**`5844f4552` / `e5599a063` — the tests as a body would not have caught a rewrite.** Mutation-tested both scripts. Fourteen of the mutations left all 48 tests green; the two worst:

- Deleting `pendingSteps` from the status ternary in `validateSteps()` — a journey with steps nobody attempted then reads as `complete`, which is exactly what the rollout gate advances on. The suite missed it because every record it checks for not-being-complete is kept out by a *structural* error instead (the blank template has an empty commit and an empty digest), and the one test that builds a clean `pending` record asserts only that `errors` is empty and never looks at `status`.
- The published rollout policy itself: `stable-25` could expose 100%, the 48h windows could drop to 1h, `revoke` could stop being a canary and `signature-regeneration` could stop being forbidden — every expectation derives from those constants, so the suite moved with them.

Also unpinned and now pinned: `complete` reading `run.status !== "fail"` instead of `=== "pass"`; dropping an OS from `ACCEPTANCE_OSES`; the 64-hex and 40-hex lengths (the only malformed digest in the file failed on non-hex characters, never on length, while a truncated paste is hex all the way down); both `<` → `<=` boundaries, where the window one would make the published 24h mean "more than 24h" and cost an operator a day; and `rollbackTarget` being null unless the decision is `rollback`. Every one of those mutations is now red.

### Deliberately not changed

- `state.candidate` is read by nothing in the gate. It is provenance for the human reading a committed state file, and cross-checking it against `rollbackReadiness.tag` would be a new gate rule rather than a bug fix.
- `assertBoundedRestore` keeps `String()` coercion. Here coercion runs toward *more* refusal (`["tag-move"]` is refused), which is the safe direction, and the non-object case is pinned by an existing test.
- Inherited `main` breakage (#4784). Re-verified: no failure traces to this branch.

### Verification

56 tests pass under `node --test` and under the runner's own `--require ./scripts/css-source-contract-hook.cjs` invocation. `docs-index.test.mjs` ok (50 documents, 63 links). `check-tests-wired.mjs` reports only the 8 inherited files. `pnpm lint` clean. Template parses; `stages` prints the documented table. Doc/code cross-check is now programmatic in both directions: all 19 step ids and titles, all 5 stage rows (percentage, distribution, window), all canaries, hard-stop classes, forbidden operations, and threshold defaults match between the runbooks and the constants.

**Not claiming CI is green.** `main` is red on its own, the api suite exits before it reaches line 1614, and these tests have still never executed in CI. Local verification remains the only evidence that exists. PR stays a draft.

CompleteDotTech:
## Round 5 — convergence test

Reviewed `e5599a063`, left `5d8f481c1`. **One commit.** Round 4's verdict was
very nearly right: the code is correct and I found no defect in it. One
mutation of the twenty-eight I ran survived, and it was in a test rather than
in the code.

### 1. Round 4's central factual claim (`ef9f476fa`) — verified TRUE

Checked against `scripts/verify-release-updater.mjs` directly rather than
taking it on trust, because a wrong runbook is worse during an incident than
the bug it claims to document.

- Step 3 of that script (`=== 3. version matches latest GitHub release ===`,
  lines 100–107) fetches `/repos/<repo>/releases/latest`, strips the leading
  `v` from `tag_name`, and `fail()`s on inequality with `manifest.version`. It
  exits `process.exit(failures ? 1 : 0)`. So the non-zero exit is real.
- The failure string in the runbook is the code's, character for character:
  `` `version drift: latest.json=${manifest.version} vs release=${ghTag}` ``,
  with `ghTag` un-stripped — which is why the example reads
  `latest.json=0.9.4 vs release=v1.0.0` and not `release=1.0.0`.
- The updater endpoint in `src-tauri/tauri.conf.json` is
  `https://github.com/OpenCoven/coven-cave/releases/latest/download/latest.json`.
  That resolves to the *latest release*, so "the release the updater endpoint
  resolves to" in drill step 2 is the candidate — and `tag-move` and
  `version-unpublish` are both in `FORBIDDEN_RESTORE_OPERATIONS`, so the
  candidate stays published and stays latest. The equality therefore breaks
  exactly as claimed, and only that one check breaks: the baseline's assets are
  untouched, so the schema, asset-resolution, and signature steps all still
  pass. "Exactly one failure" is accurate.

On the "could this be read as *ignore failures here*" question: no. The doc
names the single expected line, says *"That line is the rollback, not a failure
of it"*, enumerates the other `✗` shapes as meaning the restore did **not**
take, and closes with *"Read the whole output rather than the exit code."* The
step `detail` in the code carries the same split.

### 2. Convergence evidence, re-derived rather than trusted

Full battery re-run at `e5599a063`: 56 pass under `node --test`; 21 + 35 under
the runner's own `--require ./scripts/css-source-contract-hook.cjs` invocation;
`docs-index` ok (50 documents, 63 links); `check-tests-wired` reports **only**
the 8 inherited files; `pnpm lint` clean; both CLIs print usage and exit 1 with
no args; `release:acceptance template` parses.

I then wrote **28 fresh mutations** of my own choosing across both scripts —
boundary flips (`<`→`<=` on both rate thresholds and on the observation
window), guard deletions (`isPlainObject` on regressions, the 0..1 range check,
the integer-count check, the unknown-threshold hold, the duplicate-OS check,
the unknown-step-id check, the `seen` cycle guard), vocabulary edits (a stage
percentage, a secret pattern, `DIAGNOSED_RESULTS`), coercion re-introductions
(`readString`, `readWord`), field-order swaps in `baselineName` /
`readRollbackReadiness`, CLI exit-code flattening, and a revert of `ef9f476fa`'s
own wording. **27 were caught.** Round 3's property-probe and round 4's
mutation sweep hold up.

### 3. The one survivor — `test(release): pin that only ready: true proves a way back` (`5d8f481c1`)

Loosening `verdict?.ready === true` to `verdict?.ready !== false` in
`readRollbackReadiness` left all 56 assertions green.

The code is right — it holds. The hole is that the test named for this property,
`rollback readiness must be asserted, not merely unmentioned`, only ever
exercised verdicts that omitted `ready` **and** the baseline together, so the
hold always arrived from the *"names no baseline"* branch and the reading of
`ready` was never measured. That is the same class round 4 named in
`5844f4552` — a test passing for a reason other than its name.

It is not academic. On a verdict that names a baseline but omits `ready`, the
fail-open reading does not hold, it **advances**: a named baseline satisfies
every later guard. And that is the shape this runbook invites, since it asks an
operator to hand-write `rollbackReadiness` — a transcription keeps the baseline
it can see and drops the boolean it cannot. Verified the real gate holds on it
with `rollback readiness is not proven` before adding the pin; the added
assertions cover the decision, `rollbackReady`, and the reason text, since *"not
proven"* and *"names no baseline"* send an operator to different places.

### 4. Coherence of the 19 commits as one change

The branch is **purely additive** against `origin/main` — 2805 insertions, zero
deletions, ten files, and the only two outside its own surface are two lines in
`scripts/run-tests.mjs` and one link in `docs/workflows/branching.md`. So
nothing can be vestigial-from-removal, and no commit reverts an earlier one. I
re-read both runbooks against the code end to end: every stage id, percentage,
observation window, threshold number, JSON field, subcommand, and forbidden
operation in the docs exists in the code with that spelling. The
`4731b085d` / `9d46ccf9f` pair reads coherently — the waiver prints once on a
non-rollback report, and a file that both waives and names now holds.

No re-derived rollback readiness survives anywhere in the code, tests, or docs.
`scripts/release-rollout.mjs` reads only `ready`, `baselineWaived`, `baseline`,
and the transcribed `error`; `baselineVersion` and `blockers` appear nowhere
except in the fixture assertion that pins their absence.

Also corrected one stale claim in the PR description: the validation block
still said `# 11 pass` / `# 16 pass`, which were the counts eighteen commits
ago. They are 21 and 35.

### 5. Against issue #4781 as scoped

The three deliverables are met by the machinery: all ten journey verbs
(install/pair/read/send/resume/restart/attachment/action/revoke/update) map to
`JOURNEY_STEPS`, all six global-CLI verbs to `CLI_STEPS`, `stable-5` is the
low-percentage stage, and `HARD_STOP_CLASSES` is exactly the four classes
acceptance criterion 3 names.

What is **not** here, and cannot be: acceptance criterion 1 says the journey
*passes* on all supported OSes. That is an execution on three real machines
against published signed artifacts — the issue carries `needs-human` and
`verification-required` for that reason. This PR lands the record format, the
validator, and the empty `docs/release-acceptance-results/` the run's evidence
goes into. **The bead cannot close on this merge**; it closes when a validated
record for a real candidate is committed there.

### Not touched

Everything in the §3 inherited table (#4784's). Confirmed independently that
none of it can trace here: `tsconfig.json`'s `include` has no `**/*.mjs`, so
`pnpm typecheck` never sees these four files, and `lint:source` globs only
`src/**`.

Still a draft, still `BLOCKED` on a red `main`, and this branch's tests still
sit at lines 1614–1615 of `SUITES.api` behind the first-failure exit at 1458 —
so CI has still never executed them. Local runs remain the only evidence.

CompleteDotTech:
## Review loop complete — converged at `2e648b951`

Six review rounds, 21 commits, 19 defects fixed. Round 6 returned **zero findings and zero commits**, which is what ended the loop. Test count went 27 → 56, hardened by a 1,527-assertion property probe over every JSON input field and three independent mutation sweeps (12/12, 28/28, and 14 previously-surviving mutations, all now caught).

The substantive fixes were fail-open bugs in a gate whose entire job is to refuse:

- `Number(null) === 0` — an unmeasured crash-free rate read as a total outage, an unmeasured duplicate-send count read as zero duplicates. One coercion breaking the safety rule in both directions.
- `String(["complete"]) === "complete"` — a state file with `"status": ["complete"]` **advanced a rollout with no acceptance evidence at all**.
- A `null` or `"garbage"` threshold override silently *deleted* the check rather than relaxing it (`0.1 < null` is false).
- The `#4782` shape mismatch (`.baselineVersion` / `.blockers` do not exist) — a rollback decision that could not name what to roll back to.
- `findSecrets` blew the stack at ~5k nesting inside a function contracted never to throw.
- The drill's own verify step reports `version drift` on a **successful** rollback, because the candidate stays `releases/latest` while `tag-move` is forbidden. Now documented, so an operator seeing `FAILURE(S)` mid-incident does not conclude the restore failed and roll the regression forward again.

## New evidence on the `main` blocker

Merged `fix/cave-20wrn-restore-streaming-merge` (#4784) into this branch **locally** to test the interaction, then discarded it (`git reset --hard`; worktree is pristine at `2e648b951`, zero diff from HEAD):

- **The merge is clean** — including `scripts/run-tests.mjs`, where both PRs add suite entries. No conflict between this branch and the `main` repair.
- **The repair works**: with it applied, `pnpm test:mobile` passes all 85 files, including `chat-view-mobile-command-center.test.ts`, which is the app-suite failure this PR has been blocked behind.

A full local CI simulation is **not** achievable on this Windows checkout: `scripts/worktree-guard.test.mjs`, `scripts/worktree-autolock.test.mjs`, and `src/lib/server/research-landing.test.ts` all fail here for environment reasons, independently of this branch (the guard test fails identically in an unrelated checkout with these changes absent). So the standing caveat holds unchanged — **these two test files have still never executed in CI**, and local runs under the runner's own invocation remain the only evidence.

Nothing blocks this PR other than the inherited red `main`. `mergeable: MERGEABLE`, no conflicts, no review threads, no AI attribution on any of the 21 commits.

CompleteDotTech:
## CI is green — head `8afe293df`

Every check passes, including the required `Frontend build`:

```
pass  Frontend build                              pass  PR checks
pass  Frontend bundle                             pass  Select validation
pass  Frontend validation (API tests)             pass  iOS build
pass  Frontend validation (app tests)             pass  Frontend validation (lint)
pass  Frontend validation (mobile tests)          pass  Frontend validation (typecheck)
pass  Frontend validation (protocol conformance)  pass  Frontend validation (test wiring)
```

`mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

### The standing caveat is now closed

Through six review rounds this PR carried the same disclaimer: its two test files had **never executed in CI**, because the api suite exited at its first failure on line 1458 while these sit at 1614–1615. `main` is repaired, so the suite now runs to completion — `✓ 394 test file(s) passed [api]` — with both files in it. They have finally run, and they pass.

`check:tests-wired` also reports `✓ all 1766 test files wired into CI`, rather than the 8 orphans it named on every previous run.

### One flake, re-run and cleared

The first attempt failed a single Playwright spec, `tests/right-chat-panel.spec.ts:184` (mobile focus-trapped drawer), on `pixel-5` — while the **same** spec was reported flaky on `iphone-13` in that run, i.e. it failed then passed on retry. 286 passed, 1 failed, 2 flaky. This branch touches no file under `src/` or `tests/`; the e2e lane runs at all only because `package.json` is in its path filter. Re-running the failed job alone turned the run `success` with no code change.

Left as a **draft** deliberately, per the original request. The bead `cave-udcn7` should **not** close on merge — it closes when a validated three-OS acceptance record lands in `docs/release-acceptance-results/`, which is human execution this PR cannot perform.

PR #4792, opened 2026-08-21
Report and attribute a red main after a direct-to-main push

CompleteDotTech:
## The cause

Between 00:29 and 01:22 CDT on 2026-08-21, seven feature branches were merged
into `main` **locally** and pushed direct:

| commit | merged branch | had a PR? |
| --- | --- | --- |
| `b5f40e54d` | `fix/cave-frsmb-dev-sidecar-token` | #4748 |
| `8bcaf33b8` | `fix/cave-cma81-remove-ios-terminal` | no |
| `5771a5558` | `feat/cave-20wrn-calm-streaming-chat` | no |
| `86ccc8b57` | `fix/cave-w7wzu-research-video-e2e` | no |
| `0f875c5da` | `fix/cave-w3hx6-empty-self-report` | no |
| `1257258ce` | `fix/cave-atox4-marketplace-logo-colors` | no |
| `b4c737dc1` | `feat/cave-kex69-performance-reports` | no |

Six of the seven never had a pull request at all. Branch protection gates pull
requests, not pushes, and `enforce_admins` is off by the owner's standing
instruction — so nothing checked any of them.

CI *did* run on each push. Nobody read it. `main` has failed on every push since
`1257258ce`, and the first thing to notice was a session reading the commit list
seven hours later. The damage that landed unchecked:

- `1257258ce` — its own `tests/marketplace-logos.spec.ts` fails.
- `5771a5558` — a conflict resolution kept three uses of `bubbleSegments` and
  dropped its declaration, so typecheck, build, and app tests all fail.
- `5771a5558` — eight test files landed wired into no CI suite.

## What this PR changes

Not the permission gap — the **observation** gap. `enforce_admins` is the
owner's setting and this PR does not touch it, argue about it, or work around it.

`scripts/main-health.mjs` walks `main` back from HEAD to the last commit CI
judged green and names the **oldest** failing commit in the streak — not the
head, which after a burst of direct pushes is usually just downstream of the
break — then says how that commit landed.

```console
$ GITHUB_TOKEN="$(gh auth token)" GITHUB_REPOSITORY=OpenCoven/coven-cave pnpm main:health
main-health: main is RED at b5f40e54d.
  oldest failing commit 1257258ce Merge remote-tracking branch 'origin/fix/cave-atox4-marketplace-logo-colors'
  landed: pushed directly to the branch — a local merge of fix/cave-atox4-marketplace-logo-colors, no PR
  last green b4c737dc1 Merge remote-tracking branch 'origin/feat/cave-kex69-performance-reports'
  would open-issue
```

That is the live repository, and it reproduces by tool the diagnosis above that
was made by hand.

`.github/workflows/main-health.yml` runs the same script with `--apply` after
every CI run on `main`, plus an hourly sweep because `workflow_run` deliveries
drop here — the same gap `ci-recovery.yml` exists for. Apply mode keeps exactly
**one** open issue labelled `main-red`, deduplicated by culprit SHA: it retargets
that issue if an earlier culprit turns up, and closes it when `main` is green.
The job holds `issues: write` and nothing else — it never repairs, reverts, or
pushes.

## Two deliberate refusals to guess

A wrong accusation is worse than silence, so:

- **A merged PR is credited only on an exact `merge_commit_sha` match.**
  `/commits/{sha}/pulls` also lists open PRs that merely *contain* the commit,
  which every branch cut from `main` afterwards does. Matching loosely would
  accuse a PR-landed commit of being a direct push.
- **An unavailable association stays unavailable.** That connection degrades
  under throttling (the worktree patrol hits the same thing), so a failed lookup
  is reported `undetermined`, never as "no PR".

And a `cancelled` run is neither blame nor clearance — pushes supersede each
other's concurrency group, so `main` cancels healthy runs by design. Only
`success` is green; commits CI never judged are listed rather than dropped.

## Verification

- 17 tests, all passing, run with the runner's own argv
  (`node --require ./scripts/css-source-contract-hook.cjs --test`).
- Two real bugs the tests caught while building it: the plan closed the tracking
  issue on a head CI had not judged yet, and a `cancelled` head read as green.
- `pnpm lint` passes. `pnpm check:tests-wired` confirms both new tests are wired.
- Report-only mode verified against the live GitHub API; it makes no mutations.

## Status: rebased onto a repaired `main`

`main` is green again as of `b9856a7ed` (#4769), which turned out to carry all
three outstanding fixes — the `bubbleSegments` declaration, the eight orphaned
test files, and the iOS colon below. This branch has been rebased onto it and
its CI now runs against a healthy base.

### The fourth casualty, for the record

While triaging this PR's own red CI I found a break nobody had reported.
`8bcaf33b8` — the local merge of `fix/cave-cma81-remove-ios-terminal`, no PR —
dropped a single colon:

```
5771a5558  let local = globalDirectThreads(for: familiarId).map(...)   // parent, correct
8bcaf33b8  let local = globalDirectThreads(for familiarId).map(...)    // does not compile
```

`main` did not compile for iOS from 01:06 CDT until #4769 landed. It stayed
invisible because the iOS job only runs when a PR touches an iOS-selecting path,
and none had — this PR tripped it only because it edits `package.json`. That is
precisely the class of thing this PR exists to surface: a break that CI *did*
catch on the push and that nothing reported.

Closes cave-8gc.


CompleteDotTech:
Merged `main` in and resolved the one conflict (`045f25268`) so this could land — your branch was green at 12/12 but had gone `CONFLICTING` against `main`, and the collision was partly mine to clean up.

The conflict was in `scripts/run-tests.mjs`: your two `main-health` entries and `worktree-lifecycle-retirement.test.mjs` sat in a SUITES region that `main` has since rewritten. Resolved by keeping your two entries and **dropping the `worktree-lifecycle-retirement.test.mjs` line** — `main` relocated that entry to line 91, so re-adding it here would have listed it twice in the same suite.

Verified after resolving: `pnpm check:tests-wired` → all 1772 wired, `node scripts/check-conflict-markers.mjs` clean, `pnpm lint` clean.

Your worktree at `.worktrees/cave-8gc-main-health-guard` was clean with nothing unpushed and had been idle a couple of hours, and I did the merge on a scratch branch rather than in it, so nothing of yours was touched. It is now behind `origin` by the merge commit — `git pull` there before you continue.

CompleteDotTech:
Correction to my previous comment — my first resolution introduced a defect, now fixed in `4431273af`.

`045f25268` rewrote **the entire `scripts/run-tests.mjs` with CRLF line endings**. I resolved that conflict with a Python text-mode write on Windows, which silently translated every `\n` to `\r\n` across all 4,481 lines. `main` has the file in LF.

The effect was worse than cosmetic: the next merge from `main` conflicted on the *whole file* rather than two lines, because no line matched. That is what made this PR look like it had become badly divergent when the real difference was still only your two `main-health` entries — `git diff --ignore-cr-at-eol` reported exactly `2 insertions, 2 deletions`.

Fixed by taking `main`'s copy of the file and re-inserting the two entries with a binary-safe write. Verified:

- `file scripts/run-tests.mjs` → no "with CRLF line terminators"
- `git diff origin/main -- scripts/run-tests.mjs` → exactly `+2` lines, the two `main-health` entries
- `pnpm check:tests-wired` → all 1774 wired
- `scripts/check-conflict-markers.mjs` → clean

Sorry for the noise. Worth knowing generally, since several sessions here script conflict resolutions on Windows: `open(path, "w")` in Python will do this to any file it rewrites. `open(path, "wb")` or `newline=""` avoids it.

Auto-merge is disabled on this repository, so this still needs someone to merge it once `Frontend build` reports — that job takes ~25 minutes because of the Playwright step, and `main` has been moving faster than that.

PR #4793, opened 2026-08-21
Fix main's last e2e failure (no duplicate send), and forbid duplicate suite wiring (#4790)

### LINKED ISSUE #4790: check-tests-wired accepts duplicate suite entries, so two branches wiring the same test merge into a silent double-run
`scripts/check-tests-wired.mjs` answers one question — is every test file on disk referenced by *some* suite? It builds a `Set` of referenced paths (line 53) and reports what is missing from it. Nothing anywhere asks whether a path is referenced **more than once**, and `scripts/run-tests.mjs` does not dedupe the arrays before running them.

That gap is reachable today rather than theoretical.

`main` at `b5f40e54d` has eight orphaned streaming-chat test files. Two open branches independently wire all eight into `SUITES.app`:

- `fix/cave-20wrn-restore-streaming-merge` (#4784) — at offsets ~305 and ~590
- `feat/cave-gcb0i-chat-signed-packages` (#4776) — at offsets ~372 and ~442

Because the insertions are far apart, git merges them **cleanly**. `git merge-tree` on the two branch tips produces a `run-tests.mjs` in which every one of the eight appears twice:

```
streaming-chat-wiring: 2
streaming-turn-response: 4      # 2x SUITES.app + 2x VITEST_TESTS
chat-assistant-output: 2
chat-result-markers: 2
chat-tool-verification: 2
streaming-markdown-blocks: 2
streaming-turn-view-model: 2
use-stick-to-bottom: 2
```

The result is not a conflict and not a test failure — it is eight test files quietly running twice in the `app` suite, on a runner that already takes minutes. `VITEST_TESTS` is a `Set`, so that half dedupes at runtime and the duplicate is invisible even in the source scan.

The same shape recurs whenever two sessions repair the same unwired-test gap, which is a common way for this repository's suites to drift back into agreement.

## Suggested fix

Have `check-tests-wired.mjs` fail on a path listed more than once across the suite arrays, the way it already fails on a path listed zero times. A gate that enforces "at least once" but not "at most once" is only half a wiring check, and the missing half is the one that produces no visible symptom.

Worth considering alongside it: whether `run-tests.mjs` should dedupe defensively at run time, or whether that would mask exactly the drift the check should surface.

Found while reviewing #4782 and #4784.

CompleteDotTech:
Two repairs to the test plumbing that let `main` go red without anyone seeing why.

## 1. task-work-fit's "duplicate send" — there is no duplicate send

`[desktop] › tests/task-work-fit.spec.ts › opening the code rail does not re-send
the task's first prompt` was `main`'s single remaining e2e failure. **The task's
first prompt is dispatched exactly once. The spec's instrument was wrong, not the
code.**

The send log counted *every* POST to `/api/chat/send` — but that endpoint is not
the task handoff's private channel. The shell generates the daily narrative
through it too (`workspace.tsx` → `generateDailyNarrative`). Captured log:

```
#1 +0ms     runId=?        sessionId=null      prompt="Write a short narrative of my day (Aug 21)..."
#2 +1463ms  runId=d0bc2d5b sessionId=s-bridge  prompt="Add the Coven stateless spoke protocol."
```

Send #1 carries no `runId` and no session — unrelated background traffic that
landed inside the window. Send #2 is the handoff, once.

Two things had disguised this:

- **It fails at the premise**, on the first assertion, *before* the code rail is
  ever opened — so the remount latch the test is named for was never implicated.
  The bare `Expected: 1, Received: 2` said nothing about which prompt.
- **The regression window was an artifact.** e2e genuinely passed at `b4c737dc1`
  and genuinely failed at `1257258ce`, whose entire diff is marketplace logos
  plus 55 lines of new specs — nothing there can re-send a prompt; it just added
  enough load to shift when the narrative call lands. Every "green" run in
  between had the e2e step **skipped** by path-aware CI, so reading the window
  from run conclusions was misleading. That is also why it reproduced
  deterministically on CI and never on a warm local run.

**The fix does not weaken the assertion.** It selects the sends carrying the
task's own prompt — a genuine re-send of the first prompt necessarily carries it
— so it is strictly more precise than the bare count: false positives drop, every
true positive stays. The prompt is now one shared constant, so the mock and the
assertions cannot drift. The full log (unrelated sends included) rides in both
failure messages.

Verified both directions: passes 3/3 as written, and still fails — naming both
`runId`s — when the handoff latch is deliberately disabled so the rail remounts
really do re-send.

## 2. `check-tests-wired` now enforces "exactly once" (#4790)

The guard built a `Set` of referenced paths and reported what was missing. A test
wired **zero** times failed loudly; a test wired **twice** passed silently and
ran twice — `run-tests.mjs` flat-maps the suite arrays with no dedupe.

Reachable, not theoretical: two branches independently wiring the same orphaned
test insert it at different offsets, git merges both cleanly, and the double-run
lands with no conflict and no symptom beyond a slower suite. Merging
`origin/main` into #4784 reproduces it exactly — eight streaming-chat tests wired
twice over.

- Added the missing half, printing the offending suite/index.
- Removed the **11** paths already duplicated across the suite arrays. Nine were
  within `app`; two (`worktree-lifecycle-retirement.test.mjs`,
  `session-security.test.ts`) sat in both `app` and `api`, which CI runs in the
  same pass. All 11 still run — once. Keyed globally for that reason.

Verified both directions: `✓ all 1757 test files wired into CI` on the deduped
tree, and a clean failure on an injected duplicate.

Note for anyone tracking #4790: `streaming-turn-response.test.tsx` is *not* one of
the duplicates. Its second occurrence is in `VITEST_TESTS`, a runner-selection
`Set`, not a run list — it does not double-run.

Closes #4790.


CompleteDotTech:
Two pieces of evidence worth recording, both gathered after the PR body was written.

**1. Corroboration that there is no production duplicate send.**

#4784's CI run [32468966372](https://github.com/OpenCoven/coven-cave/actions/runs/32468966372)
came back **green with the e2e step executed** — `task-work-fit` ran and passed — on
head `4d43d30cb`, which carries **identical chat code to `main`** and does **not**
contain this PR's spec fix.

A genuine duplicate-send defect in the handoff would have failed there too. It did
not. That is independent of the captured send log and points the same way: the
production path dispatches the task's first prompt exactly once, and the spec was
measuring unrelated traffic.

It also sharpens the failure's character. It is **intermittent across runs**, not
deterministic — when the daily-narrative call lands inside the window it fails both
retry attempts (which is why a single run looks deterministic), and when it doesn't,
the spec passes. That is precisely the profile of a false positive keyed to timing,
and it is why it tracked suite load rather than any commit.

**2. The same structural weakness exists in at least one other spec.**

`tests/prompt-enhance.spec.ts:298` asserts `expect(sends).toHaveLength(1)` over every
POST to `/api/chat/send`. A loaded local run caught it holding two — drafts
`"Ship the thing"` and `"fix login bug"`, both `origin: "enhance"`.

The proximate cause there looks different (two distinct enhance calls, apparently
leaking across test steps, rather than background traffic), so I have **not** changed
it here — that is a separate diagnosis and does not belong in this PR. But the shape
is the same and worth naming: *an exact-count assertion over an endpoint the shell
also uses is measuring more than its own surface.* Note that spec's own next
assertion already discriminates on `origin: "enhance"`, so the discriminator it needs
is right there.

I attempted to file this as a bead and was blocked — see the PR thread on tooling.

Issue #4841, opened 2026-08-22
Proxy pre-authorizes 13 client-v1 paths that have no handler yet


CompleteDotTech:
Found independently by two reviews of PR #4840 (head `b36e16fc9`), which lands the Chat v1 Phase 1 pairing authority.

`CLIENT_V1_AUTHENTICATED_PATHS` in `src/proxy-helpers.ts:175-188` lists thirteen paths — `/api/client/v1/familiars`, `/projects`, `/conversations`, `/conversations/search`, `/messages/send`, `/attachments`, `/commands`, `/tasks/handoff`, `/runs/:id/{stream,stop,retry}`, `/attention/:id/respond`, `/github/actions`.

**None of them has a `route.ts`.** `src/proxy.ts:315` computes `clientV1Ingress` from that list, `src/proxy.ts:316-323` then skips `mobileAccessGate` entirely, and `src/proxy.ts:466-471` returns `nextWithMobileAccessMarker(req, false)` *before* the sidecar-token block at `src/proxy.ts:473`.

### Failure scenario

The day a Phase 2 handler lands at `src/app/api/client/v1/conversations/route.ts`, it is reachable from any loopback process with **no sidecar token**. The route's own `requireScope` call becomes the only remaining layer — and if its author omits it, there is no second one. The Phase 0 allow-list in `api-contracts.test.ts` catches the route's *existence* but not a missing bearer check.

Compounding it: `requireScope` (`src/lib/server/client-v1/auth.ts:46`) and `consumeAuthenticated` / `consumeInvalidBearer` (`src/lib/server/client-v1/rate-limit.ts:130-135`) currently have **zero non-test call sites**, so bearer authentication attempts will also be unthrottled when those routes land.

### Not currently exploitable

Verified: `find src/app/api/client -name route.ts` returns only health, pairing×3, and admin×4. There is no live exposure today. This is phase1a's design carried through the merge unchanged — but it is new to `main`.

### Fix

Either move the bearer guard into the proxy, or add each path to `CLIENT_V1_AUTHENTICATED_PATHS` as its handler lands rather than ahead of it.

**This must be resolved before the first authenticated client-v1 route merges — which is #4834.**

CompleteDotTech:
**Fixed.** PR #4847 squash-merged 2026-08-22T08:29:26Z as `57c8cded`.

### The hole is measurably closed

[...]

Issue #4843, opened 2026-08-22
client-v1 admin routes are not bound to direct loopback


CompleteDotTech:
Found by the security review of PR #4840 (head `b36e16fc9`).

`/api/client/v1/admin/*` is in neither `CLIENT_V1_PUBLIC_PATHS` nor `CLIENT_V1_AUTHENTICATED_PATHS`, so `clientV1IngressKind` returns `null` and the admin family skips the `if (!trustedLocalPeer || remoteIngress)` gate at `src/proxy.ts:467`.

It is protected by `COVEN_CAVE_AUTH_TOKEN` (fails closed at `src/lib/server/client-v1/admin-auth.ts:59-60` with 503 when unset) plus, for mutations, an origin/referer check.

### Failure scenario

A remote Tailscale-Serve caller holding the sidecar token and sending no `Origin`/`Referer` — a native client rather than a browser — could **read** the credential list and pending pairing requests. Mutations are still refused, because `hasValidMutationSource` rejects a `ts.net` origin.

Severity is LOW today: the sidecar token is per-launch and never leaves the machine.

### Why fix it anyway

The pairing-approval console is the human-consent surface for the entire authority. It should require the same direct-loopback stamp the pairing routes require, rather than being reachable by anything holding the sidecar token.

Note `requireClientV1Admin` deliberately does not use the loopback stamp today (see its comment at `admin-auth.ts:54-56`) — so this is a design decision to revisit, not an oversight to patch blindly.

PR #4849, opened 2026-08-22
Meter the pairing poll route against the exchange brute-force budget (cave-11w87)

### LINKED ISSUE #4846: Pairing secret comparison on GET /pairing/requests/[id] is unmetered, bypassing the exchange brute-force budget
Found by adversarial verification of the rate-limit repair on PR #4840.

PR #4840 adds a per-pairing brute-force budget on the exchange route: `consumePairingExchangeFailure(id)`, 10 failures per pairing id per 60s window, charged only on `secret_mismatch`. On that route the bound is exact.

**System-wide it is not a bound at all**, because a sibling route performs the identical comparison with no limiter.

### The oracle

`src/app/api/client/v1/pairing/requests/[id]/route.ts:34` calls `pairingStore.lookup(id, secret)`.
`src/lib/server/client-v1/pairing-store.ts:344` performs the **byte-identical** `hashesEqual` against the same `secretHash` that `consumeForExchange` uses at `pairing-store.ts:281`, returning `secret_mismatch` → 401 versus `found` / `consumed`.

That handler references **no rate limiter**.

### Attack

Knowing a pairing id:

1. Send unlimited `GET /api/client/v1/pairing/requests/<id>` with `x-coven-cave-pairing-secret: <guess>`. A 401 means wrong, 200/409 means right. Zero budget consumed.
2. Once the secret is known, send one `POST .../exchange`. `peekPairingExchangeFailure` sees an empty bucket and allows it.

### Severity: low, but the claim is what matters

Not practically exploitable — the pairing secret is 43 base64url characters (256 bits, `contract.ts:420`), ingress is loopback-only, and the TTL is 5 minutes. Guessing is infeasible regardless.

This is also **not a regression**: the GET route is untouched by #4840, and the previous `consumePairingCreate` never covered it either.

It is filed because #4840 introduces a budget whose stated purpose is bounding brute force, and that purpose is only met on one of the two routes that compare the secret. The comment on the exchange route was narrowed in #4840 to say so explicitly rather than overclaim.

### Fix

Peek and charge the same per-id bucket in the GET handler, so both comparison sites share one budget.

### Related

`requireScope` and `consumeAuthenticated` / `consumeInvalidBearer` still have zero non-test call sites — see #4841, which must be resolved before the first authenticated client-v1 route lands.

CompleteDotTech:
Closes #4846.

`GET /api/client/v1/pairing/requests/[id]` calls `pairingStore.lookup(id, secret)`, which runs the **byte-identical** `hashesEqual` against the same `secretHash` that `consumeForExchange` compares, and answers 401 on a mismatch versus 200/409 on a hit. That handler referenced no rate limiter.

So the per-pairing brute-force budget #4840 added on the exchange route covered one of the two comparison sites, and therefore bounded neither: a guess could be tested unmetered through this route and the exchange spent once, with `peekPairingExchangeFailure` still seeing an empty bucket.

**Severity is low, as the issue says.** The secret is 43 base64url characters (256 bits), ingress is loopback-only, and the TTL is five minutes — guessing is infeasible regardless. This is also not a regression: the GET route was untouched by #4840 and was never covered. It is fixed because #4840 introduces a budget whose stated purpose is bounding brute force, and that purpose was met on only one of the two routes that compare the secret. This closes a gap in a claim, not a live exploit.

## What changed

The poll handler now peeks the same per-id bucket before comparing and charges it only on `secret_mismatch`, mirroring the exchange route exactly.

**One shared bucket, not one per route.** This is the load-bearing choice. Two buckets would meter each route and bound neither — an attacker gets ten guesses per route, and a full exchange budget still in reserve for the moment the poll oracle gives up the secret. `M5` in the matrix below is exactly that mutation, and two tests catch it.

**Peek before the comparison, charge after.** A limit charged after the fact would bound nothing; `M2` (peek removed, mismatch still charged) is caught.

**Only a wrong secret is charged.** A legitimate client polls this route while it waits on an administrator decision, so charging a correct secret would rate limit the one caller the route exists for. `M4` (charge every lookup) is caught by the 150-poll test, which asserts not merely that the holder is unthrottled but that the bucket is left completely unspent.

**`not_found` is not charged**, for the reason the exchange route already gives: the store answers `not_found` only when no record and no terminal record carries the id, so there is no secret to guess and charging would mint one bucket per guessed id — the map churn `peek` deliberately refuses to cause. `M7` is caught.

**An exhausted budget answers 429 `rate_limited` with `retry-after`**, the same response the exchange route gives, rather than a quieter 401 that would keep answering the attacker's question. `M6` is caught. While the budget is spent the correct secret is refused here too — the same deliberate per-pairing lockout the exchange route already documents, and it cannot be triggered by the holder's own polling.

**Other call sites:** `pairingStore.lookup` has exactly one non-test call site, this one, so nothing else needs the same treatment. `poll` and `inspect` also touch secret material but have no route behind them.

Two comment-only edits ride along: the exchange route's "Known gap, deliberately not closed here" paragraph is now false and says the opposite, and the rate limiter's JSDoc now records that one bucket covers every route comparing a pairing secret.

## Failing test first

Written before the fix and confirmed against the unfixed handler (`M1` in the matrix — the whole fix reverted, i.e. the state of `main`):

```
✔ poll exposes only id, status, and expiry across the complete lifecycle
✔ poll accepts pairing secrets only from the reviewed header
✖ poll meters wrong pairing secrets against the exchange's own budget
✖ poll refuses once exchange failures alone have spent the shared budget
✔ polling with the correct secret is never rate limited
ℹ pass 3   ℹ fail 2
```

The third test passes unfixed by design — it is the guard that the fix does not over-charge, and it is killed by `M4`/`M7` instead.

## Mutation matrix

Every assertion added was mutation-checked. The fix was committed first, each mutation applied to the committed file, and the file restored with `git checkout HEAD --` after each run (`git status --porcelain` empty afterwards).

| mutation | result | killed by |
| --- | --- | --- |
| M1 whole fix removed (state of `main`) | KILLED | meters-wrong-secrets; refuses-after-exchange-failures |
| M2 peek removed, mismatch still charged | KILLED | meters-wrong-secrets; refuses-after-exchange-failures |
| M3 mismatch no longer charged, peek kept | KILLED | meters-wrong-secrets |
| M4 charge every lookup, not only a mismatch | KILLED | correct-secret-never-rate-limited |
| M5 private bucket (`poll:${id}`) instead of the shared one | KILLED | meters-wrong-secrets; refuses-after-exchange-failures |
| M6 exhausted budget answers 401 instead of 429 | KILLED | meters-wrong-secrets; refuses-after-exchange-failures |
| M7 `not_found` charged too | KILLED | correct-secret-never-rate-limited |

No surviving mutants; every new test kills at least one. No existing assertion was weakened or deleted.

## Gates

`pnpm typecheck`, `pnpm lint`, `pnpm check:tests-wired` (`✓ all 1792 test files wired into CI`), and `pnpm build` all pass. Test files were run with the exact argv `scripts/run-tests.mjs` reports via `nodeArgsFor`. Green: the pairing GET route (5/5), the exchange route (5/5), `rate-limit.test.ts` (10/10), `pairing-store.test.ts` (8/8), and the pairing-create route (3/3).

## Deliberately left undone

The limiter methods are still named `consumePairingExchangeFailure` / `peekPairingExchangeFailure` and the constant `CLIENT_V1_PAIRING_EXCHANGE_FAILURE_LIMIT`, although the budget now covers both routes. Renaming would be more honest but would rewrite ~50 lines across `rate-limit.ts` and three test files, and #4841 is in flight over the same module — the issue names it as a blocker for the first authenticated client-v1 route. The JSDoc says plainly that the name records where the budget was introduced, not who may spend it. Worth a follow-up once #4841 lands.

### DIFF
--- a/src/app/api/client/v1/pairing/requests/[id]/exchange/route.ts
+++ b/src/app/api/client/v1/pairing/requests/[id]/exchange/route.ts
@@ -50,14 +50,11 @@ export function createPairingExchangePostHandler(runtime: ClientV1Runtime) {
     // blocking another's exchange. Read the budget before comparing secrets,
     // or a limit charged after the fact would bound nothing.
     //
-    // Known gap, deliberately not closed here: the bound is per-route, not
-    // system-wide. GET /api/client/v1/pairing/requests/[id] calls
-    // pairingStore.lookup(id, secret), which runs the byte-identical
-    // hashesEqual against the same secretHash and distinguishes
-    // `secret_mismatch` (401) from `found`/`consumed` — while referencing no
-    // rate limiter at all. An attacker holding a pairing id can therefore
-    // guess the secret unmetered through GET and spend a single exchange call
-    // once it has one. Metering that oracle is tracked separately.
+    // The bound is system-wide, not per-route: GET
+    // /api/client/v1/pairing/requests/[id] compares the same secretHash
+    // through pairingStore.lookup and charges THIS bucket for its own
+    // mismatches. Keep it that way — a second bucket for that route would
+    // meter it while leaving the pair of them unbounded.
     const limit = runtime.rateLimiter.peekPairingExchangeFailure(id);
     if (!limit.allowed) return clientV1RateLimitResponse(limit);
 
--- a/src/app/api/client/v1/pairing/requests/[id]/route.test.ts
+++ b/src/app/api/client/v1/pairing/requests/[id]/route.test.ts
@@ -274,6 +274,37 @@ test("polling with the correct secret is never rate limited", async () => {
       runtime.rateLimiter.peekPairingExchangeFailure(unknownId).remaining,
       CLIENT_V1_PAIRING_EXCHANGE_FAILURE_LIMIT,
     );
+
+    // The already-exchanged branch presents a CORRECT secret too, so it has to
+    // be as free as the 200 above. A client that polls once more after its
+    // exchange succeeds — a retry loop that raced its own success, or one
+    // confirming the outcome — still holds the right secret, and charging it
+    // here would spend the shared budget on the legitimate holder and lock the
+    // pairing's own exchange out for the rest of the window. Asserted directly
+    // because every other guard in this file exercises `found` or `not_found`:
+    // without this, charging the 409 path breaks no test.
+    const exchange = createPairingExchangePostHandler(runtime);
+    const spent = runtime.pairingStore.create({
+      ...pairingInput,
+      installationId: "chat-install-spent",
+    });
+    assert.equal(runtime.pairingStore.decide(spent.id, "approved", now), true);
+    const exchanged = await exchange(
+      exchangeRequest(spent.id, spent.secret),
+      context(spent.id),
+    );
+    assert.equal(exchanged.status, 200);
+
+    const replayed = await poll(request(spent.id, spent.secret), context(spent.id));
+    assert.equal(replayed.status, 409);
+    assert.equal(
+      ((await replayed.json()) as { error: { code: string } }).error.code,
+      "conflict",
+    );
+    assert.equal(
+      runtime.rateLimiter.peekPairingExchangeFailure(spent.id).remaining,
+      CLIENT_V1_PAIRING_EXCHANGE_FAILURE_LIMIT,
+    );
   } finally {
     assert.equal(resolve(root).startsWith(scratchPrefix), true);
     await rm(root, { recursive: true, force: true });
--- a/src/app/api/client/v1/pairing/requests/[id]/route.test.ts
+++ b/src/app/api/client/v1/pairing/requests/[id]/route.test.ts
@@ -4,8 +4,11 @@ import { resolve } from "node:path";
 import test from "node:test";
 
 import { PAIRING_TTL_MS } from "@/lib/server/client-v1/pairing-store.ts";
+import { CLIENT_V1_PAIRING_EXCHANGE_FAILURE_LIMIT } from "@/lib/server/client-v1/rate-limit.ts";
 import { createClientV1Runtime } from "@/lib/server/client-v1/runtime.ts";
+import { LOCAL_PEER_HEADER } from "@/proxy-helpers.ts";
 
+import { createPairingExchangePostHandler } from "./exchange/route.ts";
 import { createPairingRequestGetHandler } from "./route.ts";
 
 const scratchPrefix = resolve(process.cwd(), ".scratch-client-v1-poll-");
@@ -26,6 +29,23 @@ function request(id: string, secret?: string, querySecret?: string): Request {
   return new Request(url, {
     headers: secret ? { [secretHeader]: secret } : undefined,
   });
+}
+
+function exchangeRequest(id: string, secret: string): Request {
+  return new Request(
+    `http://127.0.0.1:3020/api/client/v1/pairing/requests/${id}/exchange`,
+    {
+      method: "POST",
+      headers: { [LOCAL_PEER_HEADER]: "loopback-secret", [secretHeader]: secret },
+    },
+  );
+}
+
+// A well-formed guess, not a malformed one: a secret failing the contract
+// shape is refused before the store ever compares it, so only this kind of
+// attempt reaches the comparison the budget exists to bound.
+function nearMiss(secret: string): string {
+  return `${secret.slice(0, -1)}${secret.endsWith("A") ? "B" : "A"}`;
 }
 
 test("poll exposes only id, status, and expiry across the complete lifecycle", async () => {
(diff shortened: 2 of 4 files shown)


CompleteDotTech:
## Independent adversarial review

Reviewed at `1992b4f20`, verified independently rather than by re-reading the author's matrix. The fix is correct and I could not break it. Two notes, one of which is a correction to the PR body.

**Verified from the source, not the claim.** `lookup` (`pairing-store.ts:341-348`) and `consumeForExchange` (`pairing-store.ts:278-284`) are byte-identical in their comparison, both resolving `record?.secretHash ?? terminal?.record.secretHash ?? DECOY_SECRET_HASH` and both dispatching `secret_mismatch` vs `not_found` on `record || terminal`. `hashesEqual` is `timingSafeEqual` over fixed-length 32-byte digests with a decoy for the miss path, so there is no timing side channel to fall back on. The bucket really is shared: both routes call `peek`/`consume` with the bare `id` against the single `pairing-exchange-failure` category, limit 10, window 60s. Call-site claims hold — `pairingStore.lookup` has exactly one non-test caller (this route), and `poll` / `inspect` have only test callers (the `inspect()` hits in `onboarding-core-tools.ts` are an unrelated `dependencies.inspect()`).

**Peek-before-compare is load-bearing, and the availability cost is unavoidable.** The obvious "kinder" variant — compare first, return 429 only on a mismatch — preserves the holder's poll during a lockout but hands the attacker an *unbounded* 429-vs-200 oracle, which is strictly worse than what is being fixed. I mutation-tested exactly that variant (`MX3`) and `poll refuses once exchange failures alone have spent the shared budget` kills it. So the per-pairing lockout is inherent to closing the oracle, not a design slip, and that test is the one pinning it down.

### One mutation survived, now closed

Charging the `consumed` (409) branch broke **no** test. The budget's stated invariant is that only a *wrong* secret is ever charged, and the guards covered `found` (150 polls leave the bucket whole) and `not_found` — but nothing exercised `consumed`, which also presents a *correct* secret. A charge there would spend the shared budget on the legitimate holder and lock that pairing's own exchange out for the rest of the window.

Closed in `01a2112b9` with a direct assertion. My 11-mutation set is now 11/11 killed, including the 7 the author reported plus `MX3` above, a global-bucket-key mutation (caught by the bystander assertion), and a `retry-after: 0` mutation.

### Correction to the severity paragraph

> ingress is loopback-only

This does not hold for **this** route, and it is worth recording because it is cited as a mitigation. `GET /api/client/v1/pairing/requests/[id]` is the only one of the three pairing routes with **no** `isTrustedLoopback` check — the create POST (`requests/route.ts:20`) and the exchange POST (`exchange/route.ts:28`) both have one. And `proxy.ts:316` skips `mobileAccessGate` for every client-v1 path (`clientV1Ingress ? null : await mobileAccessGate(...)`), with this path listed in `CLIENT_V1_PUBLIC_PATHS`. So with mobile access armed, a tailnet peer reaches this handler with no credential at all.

Severity is still **low** and the conclusion is unchanged: an attacker needs an existing pairing id, and the two routes that mint one are loopback-gated. But the oracle this PR closes was remotely reachable, which argues *for* the fix rather than against it. The missing loopback check is pre-existing (from #4840), out of scope here, and wants its own issue.

Gates on `01a2112b9`: `typecheck`, `lint`, `check:tests-wired` (1792 wired), `build` all pass; pairing GET 6 assertions/5 tests, exchange 5/5, `rate-limit` 10/10, `pairing-store` 8/8, pairing-create 3/3.

Agreed on leaving the `…PairingExchange…` naming to a follow-up behind #4841.

CompleteDotTech:
## Correction to my previous comment — the severity paragraph was right, I was wrong

Retracting the "ingress is loopback-only does not hold for this route" section above. I stopped reading `proxy.ts` one branch too early.

`proxy.ts:466-470` gates **every** client-v1 path, public and authenticated alike, on direct loopback:

```js
if (clientV1Ingress) {
  if (!trustedLocalPeer || remoteIngress) {
    return jsonError(403, "forbidden peer: client v1 requires direct loopback");
  }
  return nextWithMobileAccessMarker(req, false);
}
```

The `mobileAccessGate` skip I cited at `proxy.ts:316` is not a hole — it is there because this stricter gate runs later and supersedes it. `GET /api/client/v1/pairing/requests/[id]` is in `CLIENT_V1_PUBLIC_PATHS`, so it takes that branch and a tailnet peer gets a 403, not the handler. **The PR body's "ingress is loopback-only" is accurate, and the severity framing stands as written.** This is also exactly the gate #4843 is about — admin routes are the family that escapes it (`clientV1IngressKind` returns `null` for them), which is why that issue exists and this route is not in it.

What is left is much smaller and is *not* a reachability gap: the create POST and exchange POST each carry an in-handler `isTrustedLoopback` check as defence-in-depth on top of the proxy gate, and the pairing GET does not. That is an inconsistency in belt-and-braces, not an exposure, and it does not belong to this PR.

Everything else in my review is unaffected: the shared-bucket verification, the `MX3` peek-ordering result, and the `consumed`-branch mutation closed in `01a2112b9` all stand. My apologies for the noise.

PR #4867, opened 2026-08-22
Hydrate attached X sources into a mission run and remove them afterward (cave-v3ajh)

### LINKED ISSUE #4816: Add X API research ingestion and familiar publishing
# Add X API research ingestion and familiar publishing

Implement a local-first X API integration for Cave. Research Desk must be able to retrieve current X post data, save normalized source snapshots, and attach them to familiar-scoped research work. The appropriate familiar workspace room must support drafting and explicitly approved publishing to X without leaking credentials or silently retrying external writes. Scope includes secure account connection, cost-aware bounded reads, durable provenance, room-specific UX, and end-to-end tests.

## Bead metadata

- **Bead ID:** `cave-8i8q5`

- **Status:** open

- **Priority:** P1

- **Surface:** Missing

- **Authoritative owner:** GitHub Copilot CLI

- **Labels:** `comms`, `integrations`, `research`, `security`, `x-api`

- **Parent:** None

- **Dependencies:** None

## Acceptance criteria

An X account/app can be connected without exposing credentials to the browser or familiar subprocesses; Research Desk can look up X post URLs and run bounded recent searches, save normalized durable snapshots, and attach saved posts to a research mission; Comms Operations can draft an X post and publish only after explicit human confirmation, then record the returned post ID and canonical URL; data and actions are correctly familiar-scoped; auth expiry, rate limits, billing exhaustion, malformed responses, deleted posts, and ambiguous write failures are handled fail-closed; focused tests, API contracts, lint, typecheck, design gates, and native-room verification pass.

## Source of truth

Beads is authoritative. This draft card is a GitHub Teamwork visibility mirror and should be reconciled from the Bead rather than edited as an independent task.

CompleteDotTech:
Implements `cave-v3ajh` — **criterion 4 of #4816**, the only criterion the acceptance audit found materially unimplemented.

Before this, attaching an X post to a research mission recorded a mission id on the X-side saved-source record and nothing else. `runtime/x` appeared nowhere in `src/`, the research runner imported nothing from `x-sources`, and it held **zero** `finally` blocks. A user attached a post, saw *"X source attached to the mission"*, and the run never saw it.

### The durable ledger stays identity-only

`ResearchSourceRef` gains `provider` / `externalId` / `availability`, and `parseResearchSource` accepts them as strictly as it already accepts `status` — an unknown provider or a bogus availability refuses the whole source rather than being silently dropped.

`xSourceLedgerRef` is the single writer of an X entry in the mission ledger, and **every field it emits is a function of the SAVED source**. The post body never reaches it, and the author handle in the title is read back out of the stored canonical URL rather than off the fetched post — so no author display data enters a durable record either, which is the other half of the design's exclusion.

Attaching now writes that ref onto the mission as well, under the **same per-mission action lock** the runner holds for its own read-modify-write of `mission.json`.

### Hydration is just-in-time and familiar-scoped

`hydrateMissionXSources` runs immediately before `startFlow`, on all three launch paths (create, continue/refine, retry).

- It reads **only** sources saved under the mission's own familiar, so a source claiming a foreign mission id can never reach it.
- It requires that familiar's X research capability before resolving any content, and a mission with nothing attached touches neither X config nor the network.
- A live cache entry costs no upstream call; a cold one is fetched once and cached.
- Normalized text lands in `<workspace>/runtime/x/x-post-<id>.md` — deliberately a different directory from the durable `source-files/`, so removal can be a whole-directory `rm` that cannot take a permanent file with it.
- The iteration prompt names each hydrated file **and** each attached source that is not there.

### Removal is structural, not best-effort

Hydrated text may exist only while a run is live, so removal hangs off the runner's **single mission-write choke point**: every persisted transition out of an active status drops `runtime/x`.

Settling happens through at least eight distinct paths in `research-mission-runner.ts` — normal reconciliation, orphan recovery, session-owner mismatch, a failed flow run, cancel, finish, archive, a refused launch — and every one of them ends in that one write. A per-path `try/finally` would have to be re-added correctly to each of them and to each one added later; this cannot be forgotten.

Two further mechanisms cover a process that dies mid-iteration:

1. **every hydration purges before it writes**, so a crashed run's residue can never be presented to the next run as freshly rehydrated; and
2. an **application-startup sweep** (`instrumentation.ts`) removes residue older than 24 hours — the same window as the post cache the text is derived from, so hydrated content never outlives the cache entry it came from. Age-gated, so a run owned by another live process is never robbed.

### Fail-closed, per failure class

| Failure | Behaviour |
|---|---|
| Deleted post (404) | A durable fact: recorded with `markXPostAvailability`, cached body purged, file omitted, and named to the iteration as `UNAVAILABLE` with an explicit instruction not to cite, infer or invent its content. **The run proceeds**, visibly short of that source — no amount of retrying brings it back. |
| Rate limit (429) | Launch refused. |
| Billing exhaustion (402) | Launch refused. |
| Unusable authorization (401 through one refresh) | Launch refused. |
| No connected X account | Launch refused, before any request. |
| Malformed upstream reply | Launch refused. |
| X research capability off for the familiar | Launch refused. |

"Refused" means: `startFlow` is never called, no session is spent, the mission settles as `failed` with an actionable `lastError`, and Retry re-hydrates. The user explicitly attached that source; running without it would misrepresent the mission's inputs.

## Tests

`src/lib/server/research-mission-x-hydration.test.ts` (21 tests) drives the **real** store, the **real** familiar-scoped X ledger and cache, and the **real** runner over a temp cave home, with `fetch` stubbed and no X credentials. `sources-route-behavior.test.ts` gains the attach-ledger case.

These assert what is on disk and what the iteration is told — not route source text. That distinction is the root cause the audit named: *"a regex confirms the calls a file makes and can never see the call it forgot."*

The launch test captures the runtime directory **inside** the `startFlow` stub, so it fails if hydration runs too late to matter rather than merely running.

### Mutation matrix — 11 of 12 mutations killed

| Mutation | Result |
|---|---|
| Hydration never called at launch | **6 tests fail** |
| Removal choke point deleted | **2 tests fail** (both removal tests) |
| Removal made unconditional | **3 tests fail** (incl. the running-mission negative control) |
| Hydration failure swallowed, run starts anyway | **1 fails** |
| Pre-write purge removed | **1 fails** |
| Sweep ignores the 24h age gate | **1 fails** |
| Deleted post no longer recorded durably | **1 fails** |
| Prompt no longer names attachments | **2 fail** |
| Attach route no longer writes the ledger | **1 fails** |
| Schema drops the identity fields again | **1 fails** |
| Hydration stops being familiar-scoped | **1 fails** |
| Explicit cleanup in `hydrateForLaunch`'s catch removed | **survives** — see below |

The whole feature was also reverted wholesale and the new tests re-run against the base commit: they fail there, as expected.

**The one surviving mutation is honest redundancy, not an untested path.** Removing the explicit `dropXRuntime` in `hydrateForLaunch`'s catch changes nothing observable, because the choke point already removes the runtime on the failed-launch save that immediately follows. It is kept as defence in depth for the case where `persistLaunchResult` throws before saving, and this PR says so rather than quietly leaving an unkilled mutant.

## What I deliberately left

**The agent-writable workspace `sources.json` is not pre-seeded at launch.** This is the one sentence of the design I am not literally meeting: *"An X source uses `sourceType: \"x-post\"` and stores its canonical URL, post ID, user note, and availability in `sources.json`."*

What IS met is the functional requirement behind it — the run sees the source, and the durable ledger records it. The `x-post` entry is written to `mission.json`'s `sources`, which is what Research Desk displays, what `renderSourceLedgerMarkdown` publishes, and what `createResearchMissionWorkspace` seeds `sources.json` from in the first place; and the iteration prompt names each hydrated file with its canonical URL and post ID and tells the agent to record it in `sources.json`.

What is NOT done is Cave itself writing into that agent-owned file just before the agent starts. That would mean rewriting a tracked `source-ledger` artifact on every X-attached launch, with a real risk of clobbering agent-written entries that have not been merged back, and it deserves its own review rather than riding along here. **I have left it out deliberately rather than land it untested**, so this PR does not say `Closes #4816` — whoever reads this should decide whether the sentence-level reading needs that mirror before the issue closes. If it does, it is a small follow-up on top of this.

**No live X call was made and none was attempted.** The spec's own release gate — lookup/search/refresh smoke evidence against the real X API — still needs an OpenCoven test account and API credits, and remains open independently of this PR.

Refs #4816 (criterion 4).

## Gates

`pnpm typecheck`, `pnpm lint`, `pnpm check:tests-wired` (1808 files wired) and `pnpm build` are all clean.

Suites kept green: all `x-*` unit and route suites, all four `research-mission-runner-*` suites, `research-mission-lifecycle`, and `research-x-sources.test.tsx` (47/47 — including its assertion that attach does **not** call `/api/research/missions` from the client; the ledger write is server-side).

Two suites have pre-existing Windows-only failures, verified unchanged against the base commit `f76d45445` on this machine:

- `src/lib/server/x-sources.test.ts` — 24 pass / 8 fail before **and** after (EPERM on symlink creation, cross-process lock timeouts).
- `src/lib/server/research-mission-store.test.ts` — 12 pass / 2 fail before **and** after (EPERM on symlink creation, in the test's own setup).


CompleteDotTech:
CI note, so the run history reads correctly.

The first head (`e271e93e6`) failed the required `Frontend build` twice on the e2e step, both times on `tests/right-chat-panel.spec.ts:184` (pixel-5) — `cave-m1mgi`, plus `tests/mobile/command-center-pages.spec.ts:57` on iphone-13 with the byte-identical 663.484375 vs 603.703125 numbers `cave-vkb2d` already records.

That was not this change. This branch was cut from `f76d45445`, which predates #4865 — the `cave-m1mgi` fix, whose entire diff is `tests/right-chat-panel.spec.ts`. So the old, racy version of that spec was still being run here. Nothing in this PR's diff is client-reachable UI; the same flake had previously struck #4851, whose whole diff was `src/proxy-helpers.ts` and an auth test.

`origin/main` is merged in as `821b609b7` rather than the test being touched. Every gate was re-run on the merged tree — typecheck, lint, `check:tests-wired` (1813 files), build, and the full affected-suite sweep — and CI run [32609618372](https://github.com/OpenCoven/coven-cave/actions/runs/32609618372) is green on that head.

Pre-existing Windows-only local failures are unchanged across the merge: `x-sources.test.ts` 25/7 (7–8 is its documented baseline on this machine) and `research-mission-store.test.ts` 12/2, both EPERM on symlink creation in the tests' own setup, both identical at `f76d45445`.

CompleteDotTech:
## Independent adversarial review — verified at `821b609b7`, fixes at `46ab5a581`

Nothing below is taken from the PR body. Every verdict was re-derived from the
code, and every coverage claim was re-tested by mutation rather than read.

**Verdict: the feature is sound and the architecture holds.** Removal really
does funnel through one write, the ledger really is identity-only, and the
fail-closed split is right. But **three mutations survived the suite as
merged**, one of them a real defect, and reading found two more. All five are
fixed in `46ab5a581` on this branch.

### Where the removal funnel was verified, not assumed

`deps.saveMission` has exactly one call site in `research-mission-runner.ts`
(the local `saveMission` wrapper), and `saveResearchMission` has exactly one
consumer in `src/` (that dep). The new `updateResearchMissionSources` writes
`mission.json` outside the wrapper but never touches `status`, so it cannot
strand runtime text. The choke point is therefore real and total for status
transitions. Confirmed by grep, not by the comment.

### Findings

**1. REAL — `parseResearchSource` was materially looser than `status`, and
untested.** The PR's headline schema claim is *"accepts them as strictly as it
already accepts `status`"*. `status` tests membership on the raw value;
provider/availability tested `String(value)`. `String(["x"]) === "x"`, so
`provider: ["x"]` and `availability: ["deleted"]` — JSON an agent can write
into `sources.json` — were admitted and stored behind a type that promises a
string literal. Deleting **both** validation branches outright left every suite
in the repository green. Fixed and pinned.

**2. REAL — the modal settle path had no coverage.** Removal reduces to one
predicate over mission status. Adding `checkpoint` and `paused` to the active
set — one line, and the cheapest imaginable regression — kept all 21 tests
passing. The three settles the suite reaches are `failed`, `cancelled` and a
refused launch; a normal iteration completion into `checkpoint` is the *most
common* outcome and was not among them. Fixed with a direct assertion over the
whole nine-status set plus a runner test that drives a completed iteration
through `reconcileCompletedRun`.

**3. REAL — author display data did reach a durable record.** The PR states
*"no author display data reaches a durable record either"*. True of the mission
ledger; **false of the second surface**. `xAttachmentContext` rendered
`@${file.authorUsername}` from `post.author.username` — read off the *fetched*
post — and `startFlowSession` persists the whole `FlowDoc` verbatim as
`recordFlowRun({ flowSnapshot })` in `flow-runs.json`, which outlives the run.
Normally the handle is already in the durable canonical URL, so nothing new
leaks; the exception is the handle-free `/i/web/status/<id>` form, where the
ledger deliberately omits the handle and the prompt snapshot would still have
carried it. A mutation writing `publisher: post.author.name` into the ledger
also survived the suite — only the *body* leak was covered. Both fixed: the
prompt handle now comes from the saved canonical URL exactly as
`xSourceLedgerRef` derives it, and both surfaces are asserted.

**4. REAL — an agent's `sources.json` could strip the X identity permanently.**
`mergeFileSources(stored, file)` returns `[...file, ...stored.filter(not
matched)]`, so a file entry matching by URL displaces the stored ref
*wholesale*. The iteration prompt tells the agent to record each attached post
in `sources.json` by canonical URL, so this is the expected case, not a corner:
the agent's entry replaces the identity-only ref and `provider`/`externalId`/
`availability` are gone. The next hydration restores them by URL — but a
**completed** mission never hydrates again, and `publishFinalArtifacts` renders
the clobbered list. That directly undercuts the PR's own closability argument
(*"the `x-post` entry is written to `mission.json`'s `sources`, which is what
Research Desk displays"*). Fixed by carrying external-provider identity forward
onto the entry that displaces it; every field the agent owns still wins.

**5. REAL (cosmetic) — the `runtime/` scaffolding cleanup was dead code.**
`fs.rm(dir, { recursive: false, force: true })` throws `ERR_FS_EISDIR` for
**any** directory, empty or not (verified on this Node), and the throw was
swallowed — so the comment *"Leave no empty scaffolding behind"* described
something that never happened. Now uses `rmdir`, which removes an empty
directory and refuses a populated one with `ENOTEMPTY` — literally what the
comment says.

**6. REAL (residual, not fixed here) — the crash sweep has exactly one
trigger.** It runs only in `instrumentation.register()`. The sibling X *cache*
sweep has a second trigger (`sources/route.ts` sweeps on every Research Desk
load); the runtime sweep has none. So residue from a process killed less than
24h before a restart is skipped by the age gate, and if the app then stays up,
nothing removes it until some later start that happens to fall past the window.
Reconciliation covers the common case — opening the mission fails the orphaned
run and the choke point fires — but that needs the user to look. Matches the
spec's "startup and launch sweeps" wording, so not a blocker; filed as
`cave-3zc94`.

**7. REJECTED — hydration inside the action lock.** Hydration performs an
upstream X read while the per-mission action lock is held. Reviewed and
accepted: the lock is per-mission, the read is bounded and cached, and moving
it outside would reopen the lost-update race the attach fix just closed.

**8. REJECTED — `researchMissionsRoot()` is a shared Cave project.**
`ensureResearchLandingAccess` grants the landing root to each research
familiar, so `runtime/x` is readable by any familiar holding that grant while a
run is live. This is pre-existing (durable `source-files/` already live there)
and is where the spec puts the directory. Noted, not charged to this PR.

**9. REJECTED — the disclosed surviving mutation.** Confirmed honest.
Removing the redundant `dropXRuntime` in `hydrateForLaunch`'s catch changes
nothing observable: the refused launch saves as `failed` immediately after and
the choke point removes the runtime. The reasoning in the PR body is slightly
imprecise (*"every hydration purges before it writes"* is not true of the
early-throw paths, which never reach the write), but the conclusion is right.

### Per-failure-class verdict

Agreed on all seven. `startFlow` is verifiably unreachable on the refusing
paths — in all three launch sites it is called only inside the `hydrated.ok`
branch, and the 429 test asserts `flows.length === 0`. Proceeding *visibly
short* on a deleted post is the spec's own instruction (*"omits the source
file"*), the prompt is honest about it (`UNAVAILABLE … Do not cite, infer, or
invent its content`), and it is recorded durably rather than living in React
state. Refusing on transient classes is a defensible reading of *"continues
only when the mission can do so honestly"*.

### My mutation matrix (designed independently, before reading the PR's)

| Mutation | Before | After |
|---|---|---|
| Removal skips `checkpoint`/`paused` | **SURVIVED** | killed (2) |
| provider/availability validation deleted | **SURVIVED** | killed (1) |
| `String()` coercion retained | **SURVIVED** | killed (1) |
| Author display name into the ledger | **SURVIVED** | killed (1) |
| Prompt handle read off the fetched post | **SURVIVED** | killed (1) |
| X identity carry-forward removed | **SURVIVED** | killed (1) |
| `mergeXSourceRefs` never merges | killed (3) | killed (3) |
| Sweep age gate ignored | killed (1) | killed (1) |
| Capability gate removed | killed (1) | killed (1) |
| Post body into the ledger | killed (2) | killed (2) |
| `dropXRuntime` in the catch removed | survives (disclosed) | survives (disclosed) |

### Baselines confirmed on this machine, not taken from the PR

`research-mission-store.test.ts` 12 pass / 2 fail — both `EPERM: symlink
'C:\etc\hosts'` in the test's own setup. `x-sources.test.ts` 25 pass / 7 fail
(the PR said 24/8; one is the timing-dependent cross-process lock) — six
symlink/EPERM, one cross-process lock timeout.
`research-link-materialization.test.ts` 5/1, also EPERM symlink. None of these
touch hydration.

### #4816 criterion 4

**Closable once this merges.** The criterion is *"attach saved posts to a
research mission"* plus the spec's hydrate-just-in-time / remove-after. All of
that is now met, and finding 4 was the one thing that could have silently
un-met the durable half after a completed run. The unmet sentence the PR
discloses — Cave pre-seeding the workspace `sources.json` — is a mirror of a
record that is now durable in `mission.json`, survives the agent's own writes,
is what Research Desk renders and what `renderSourceLedgerMarkdown` publishes,
and the prompt already instructs the agent to record it. I agree with leaving
it out; it is not load-bearing for the criterion. The spec's live-X release
gate remains open independently and needs credentials nobody here should use.

No live X call was made and no X credential was touched; `fetch` is stubbed
throughout.

PR #4881, opened 2026-08-23
fix(design): rescale --radius-panel with the corner-radius setting, and close the codemod's radius blind spot (#4350)

### LINKED ISSUE #4350: Border Radius Inconsistency
Image

there are inconsistencies with border radius across the application.

CompleteDotTech:
Closes #4350 (bead `cave-uvwhh`).

The issue is one sentence and a screenshot, so this started as an audit. The
short version: the codebase is **not** broadly inconsistent — there is a clear
3-step radius scale and the overwhelming majority of surfaces are on it. Two
specific, mechanical defects were making it *look* inconsistent, and both are
fixed here in ~40 lines. Everything else is deliberately left alone.

## Audit

**The scale** (`src/styles/globals/foundations.css`, documented in
`docs/coven-design-language.md`):

| token | value | rescaled by themes? | rescaled by the appearance setting? |
|---|---|---|---|
| `--radius-control` | 8px | yes | yes |
| `--radius-card` | 12px | yes | yes |
| `--radius-panel` | 16px | yes | **NO** |
| `--radius-pill` | 999px | n/a | yes |
| `--radius-sm/md/lg/xl` | `calc()` off `--radius` | via `--radius` | via `--radius` |

Usage is healthy: 176 `rounded-[var(--radius-control)]`, 21 card, 11 pill, 6
panel in TSX; 1938 `border-radius` declarations across 132 stylesheets. Verified
from the built CSS that Tailwind's `rounded-md`/`lg`/`xl` compile to
`var(--radius-md)` etc. and resolve to *this* repo's `calc()` values, not
Tailwind's defaults — so those 300+ usages do scale correctly. `rounded-full`
(165 uses) is a fixed full pill by design and is not touched.

### Finding 1 — `--radius-panel` never rescaled (the user-visible bug)

Cave ships a **Corner radius** setting (Sharp / Default / Round). It overrode
`--radius`, `--radius-control`, `--radius-card` and `--radius-pill` — but not
`--radius-panel`, in *both* apply paths (`src/lib/appearance-corner-radius.ts`
and the pre-paint `public/scripts/theme-init.js`). `--radius-panel` is a peer
step of the documented scale, `themes.css` rescales it in all four themes that
retune radii, and ~70 surfaces consume it: modals, sheets, the dashboard, the
chat reader, the GitHub detail pane.

So on "Sharp", controls dropped to 2px and cards to 4px while every panel
stayed pinned at 16px — **panels rendered 4x rounder than the cards nested
inside them.** Measured in a real browser, before vs after:

```
Corner radius = "Sharp", computed border-radius

  BEFORE  control   2px | card   4px | panel  16px   panel/card=4.0x
  AFTER   control   2px | card   4px | panel   6px   panel/card=1.5x
```

Only the Default level looked right, which is why this survived. New values
continue each level's own even step and keep `control < card < panel`:

```
sharp    2 - 4 - 6     (step 2)
default  8 - 12 - 16   (:root, unchanged)
round   12 - 16 - 20   (step 4)
```

### Finding 2 — the "zero tolerance" codemod has a blind spot for radii

`scripts/codemods/tokenize-css.mjs` matches declarations with a line-anchored
`DECL_RE`, so it only ever inspects the **first** declaration on a line. This
tree writes a lot of compressed `.sel { a:1; b:2; }` CSS, and every later
declaration on such a line passed the gate untouched.

That was hiding **31 on-scale radius literals** that should have been tokens
(21x `8px`, 11x `999px`, 2x `12px` — 22 on fully-compressed lines, 9 more that
simply were not first on their line). Each is frozen at a literal while its
neighbours rescale with the theme and the corner-radius setting — the same
class of defect as Finding 1, scattered across 14 files. All 31 are now
tokenized; each takes its token from the codemod's own `RADIUS_TOKENS` table
and is byte-identical at the default theme and level.

## Gate

New **tier 1b** in `src/lib/design-token-drift.test.ts` replays every
`border-radius` declaration as a canonical one-per-line probe and asserts the
codemod would not rewrite it — so multi-value radii are covered on the same
terms, and position on the line stops mattering.

Mutation evidence (reintroduced one `border-radius: 999px` on a compressed
line):

```
--- codemod --check (the EXISTING gate) ---
[tokenize-css] checked — 0 file(s) with drift        <- passes the mutation
--- tier 1b (the NEW gate) ---
'src/styles/globals/surface-reporting.css:87  border-radius: 999px'   exit 1
```

Also mutation-checked: dropping `--radius-panel` from the runtime apply path
fails `appearance-corner-radius.test.ts`; skewing the boot script's panel value
fails `theme-script.test.ts` with a message naming both files. All restored.

Scoped deliberately to `border-radius`. `font-size` and spacing have their own
hidden counts behind the same blind spot; widening `DECL_RE` would rewrite
hundreds of unrelated values and belongs in its own PR.

## Deliberately left alone

- **~232 off-scale radii** already banked in the `offScaleRadiusPx` ratchet —
  the documented "short solid mark" exception family (1-4px dots, ticks,
  sparkline bars where a scale step reads as a circle). Not drift.
- **7 inconsistent `var(--radius-panel, 12px|8px|14px)` fallbacks.** Dead code
  (the token is always defined) and zero runtime effect; churning them would
  be noise.
- **`rounded-full`** — a fixed full pill that does not square at "Sharp" while
  `--radius-pill` does. Correct: the deliberate squaring is reserved for the
  signature composer pill, not avatars and status dots.
- **`offScaleSpacingPx` is 1606 vs a 1607 baseline** on this branch. That -1 is
  pre-existing on `main` (spacing props are untouched here), so banking it
  belongs to whoever earned it.

## Verification

`pnpm typecheck`, `pnpm lint` (incl. the design codemod check),
`pnpm check:tests-wired` (1814 wired), `pnpm build` — all green post-rebase.
36 targeted suites pass, run with the exact `nodeArgsFor` argv: the appearance
and preferences set, the CSS contract set (`globals.css`, `css-module-order`,
`themed-chrome-tokens`, `theme-contrast-audit`, `shell-inset-layout`), and the
surface suites for all 14 edited stylesheets.

### DIFF
--- a/public/scripts/theme-init.js
+++ b/public/scripts/theme-init.js
@@ -294,9 +294,12 @@
     if (weight !== "normal" && WEIGHT[weight]) html.style.setProperty("--cave-reading-weight", WEIGHT[weight]);
     if (hyphens === "on") html.style.setProperty("--cave-reading-hyphens", "auto");
 
+    // [base, control, card, panel, pill] — mirrors CORNER_RADIUS_VALUES in
+    // src/lib/appearance-corner-radius.ts. Keep both in sync; the parity is
+    // asserted by src/components/theme-script.test.ts.
     var RADII = {
-      sharp: ["0.125rem","2px","4px","4px"],
-      round: ["0.875rem","12px","16px","999px"]
+      sharp: ["0.125rem","2px","4px","6px","4px"],
+      round: ["0.875rem","12px","16px","20px","999px"]
     };
     var radiusLevel = String(stored("cave:corner-radius", appearance.cornerRadius || "default"));
     var radii = RADII[radiusLevel];
@@ -304,7 +307,8 @@
       html.style.setProperty("--radius", radii[0]);
       html.style.setProperty("--radius-control", radii[1]);
       html.style.setProperty("--radius-card", radii[2]);
-      html.style.setProperty("--radius-pill", radii[3]);
+      html.style.setProperty("--radius-panel", radii[3]);
+      html.style.setProperty("--radius-pill", radii[4]);
     }
 
     var backdropRaw = initialized ? backdrop : stored("cave:backdrop:v1", backdrop);
--- a/src/components/theme-script.test.ts
+++ b/src/components/theme-script.test.ts
@@ -139,7 +139,7 @@ for (const property of [
   "--font-serif", "--font-sans", "--font-mono",
   "--cave-reading-leading", "--cave-reading-tracking", "--cave-reading-align",
   "--cave-reading-width", "--cave-reading-weight", "--cave-reading-hyphens",
-  "--radius", "--radius-control", "--radius-card", "--radius-pill",
+  "--radius", "--radius-control", "--radius-card", "--radius-panel", "--radius-pill",
 ]) {
   assert.equal(
     defaultBoot.inline.has(property),
@@ -168,7 +168,7 @@ for (const property of [
   "--font-sans", "--font-mono",
   "--cave-reading-leading", "--cave-reading-tracking", "--cave-reading-align",
   "--cave-reading-width", "--cave-reading-weight", "--cave-reading-hyphens",
-  "--radius", "--radius-control", "--radius-card", "--radius-pill",
+  "--radius", "--radius-control", "--radius-card", "--radius-panel", "--radius-pill",
 ]) {
   assert.equal(
     explicitBoot.inline.has(property),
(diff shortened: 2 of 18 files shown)


CompleteDotTech:
### CI status: `Frontend build` is red, and the evidence says it is not this PR

Every other check passes — lint, typecheck, app tests, API tests, mobile tests,
protocol conformance, test wiring, bundle, PR checks. Only the e2e job fails.

**It fails on `main` too, whenever it actually runs.** `main`'s recent green
runs are green because path-aware CI gave them **no `Frontend build` job at
all** (verified: runs `32614523491`, `32614418004`, `32613852390` have no such
job). The two recent `main` commits that did run it both failed:

- `main` @ `7ac95ff8f` — run `32612350233`, job `97128093841`: FAILED on
  `chat-task-chip-nav.spec.ts:99/119` and `composer-runtime-chip.spec.ts:120/143`.
- `main` @ `bf992c2ec` — run `32612268777`: FAILED on the same job.

That first list is **exactly** the set this PR's first run failed on.

**Three runs here, three disjoint spec sets** — the signature of an
environmental flake, not a regression:

| run | failing specs |
|---|---|
| 1 | `chat-task-chip-nav:99/119`, `composer-runtime-chip:120/143` |
| 2 | `settings-about-narrow:5/29`, `task-work-fit:143/280/282/288`, `workspace-half-split:20/58/68/69`, `composer-runtime-chip:120/143` |
| 3 | `grimoire-autosave:101/129/148/149/167/168/183/184`, `fit-half-width:59` |

All three show `Error: aborted` / `ETIMEDOUT` from the web server, run 2 with
repeated `[cave-home-reconciliation] lock { phase: 'failed', result: 'timeout',
errorCode: 'ETIMEDOUT' }`, and each ends in `The operation was canceled` after
the job runs long (run 3: 33m47s).

**Why this change cannot cause it:** every changed CSS line in this PR is a
`border-radius` value and nothing else —

```
$ git diff origin/main...HEAD -- '*.css' | grep "^[+-][^+-]" | grep -v "border-radius"
(no output)
```

`border-radius` cannot affect layout geometry, element presence, or server
lock contention, and the failing specs assert on navigation, autosave, split
widths and viewport fit. The only other runtime change is `theme-init.js`,
whose new lines execute solely when corner radius is `sharp`/`round`; e2e runs
at `default`, where that branch is skipped entirely.

Not re-running further, and explicitly not merging. Flagging for a maintainer:
the e2e job looks broken independently of this PR and probably wants its own
issue.

PR #4888, opened 2026-08-23
Stop a legacy conversation leaving the keyless tail mid-walk (cave-wbxcu)

CompleteDotTech:
`GET /api/client/v1/conversations` pages on `createdAt` descending (cave-fhjlu,
#4863). A conversation record with no `createdAt` sorts as the empty-string
sentinel — one contiguous, id-ordered block at the tail of the walk. That block
was documented as stable, and it was not.

## The defect

**Trigger 1 — the write path stamped it.** `buildConversation`
(`POST`/`PUT /api/chat/conversation/:id`) composed the field as
`body.createdAt || existing?.createdAt || now`. A transcript older than the
field therefore acquired one — `now` — on its next turn, and jumped from the
tail of the ordering to the **head**, above every open cursor, where the rest of
that walk never served it. cave-fhjlu again, on exactly the rows the sentinel
exists to protect. The stamp was also a wrong fact: the conversation was not
created today. The same expression let any client body rewrite a *stored*
`createdAt`, which moves a row just as freely, and on any record rather than
only a legacy one.

**Trigger 2 — the row Cave substitutes for a file it cannot read.**
`fallbackConversationSummary` carried no `createdAt` at all, so a transcript
flickering between readable and unreadable moved a row into the tail block and
back out of it under an open cursor, every time it flickered. Unlike trigger 1
this needs no writer and does not self-heal.

## Reproduced first, over the whole walk

Five assertions were added to `scripts/client-v1-conformance.mjs` and run
against the **unfixed** build before anything was changed. Three failed, on the
wire:

| Assertion | Unfixed build |
|---|---|
| `reads.conversations-keyless-row-written-mid-walk-is-still-served` | the row was stamped `2026-08-23T03:30:29.860Z` mid-walk; the whole walk served 7 of 8 rows and **lost `conversation-keyless-written`** |
| `reads.conversations-recovering-row-keeps-its-position-mid-walk` | a file that became readable again mid-walk rose above the open cursor and **was never served** |
| `reads.conversations-unreadable-row-keeps-its-position-mid-walk` | a file that became unreadable mid-walk fell to the tail, reordering the walk |

The write is driven through `POST /api/chat/conversation/:id` **over the
socket** — the route the defect lived in. A case that simulated the stamp by
rewriting the transcript file would have kept passing after the write path was
fixed, and kept failing after it was broken again.

One of the five was vacuous on its first draft and the reproduction run is what
caught it: the recovering row's `createdAt` was the *oldest* in the ledger, so
it rose to a position the walk had not reached yet, was served normally, and
passed against the broken build. It now carries the newest `createdAt` in the
ledger, so recovering puts it above the open cursor. That detail is written into
the harness comment and the runbook.

## The fix, and why not the alternatives

`createdAt` is decided once, when the record is created, and is read-only after
that.

- `existing === null` → stamp `now`, or honour a body-supplied value. That is a
  genuine creation, and an import or offline replay knows its real creation time
  better than the clock does.
- `existing !== null` → carry the stored value through **exactly**, absence
  included. A body cannot rewrite it and cannot add one.
- `ConversationFile.createdAt` becomes optional, which is what the files on disk
  have always said. `tsc --noEmit` is clean with no other change: the only
  consumers are the summary projection and `session-list-merge.ts`, which
  already wrote `conv.createdAt ?? conv.updatedAt` for exactly this case.
- The fallback row carries over the `createdAt` the last readable scan saw. The
  field is immutable, so the last value read is still the true one — a recovered
  fact, not a guess. Only a value the transcript really carried is remembered,
  and it is forgotten when the file is deleted or when a later readable scan
  finds the field gone.

**Not** a derived value (the earliest turn's `createdAt`, the file's `ctime`).
A derivation is only safe if it is invariant under every write that can happen
while the record is still keyless, and none is: deleting the first turn moves an
earliest-turn key, and an atomic replace moves a `ctime` key. Trading a
guarantee for a prettier position is the wrong side of this trade — the sentinel
tail is already served, already reachable through a cursor, and already
published.

**Not** a startup backfill (option (b) on the bead). A backfill is a mass write
that moves every keyless row at once — a skip for any walk open across it — it
writes a fabricated creation time permanently into the records that are the
*reason* the field is optional, and it cannot help trigger 2 at all, because the
contents of an unreadable file are exactly what cannot be reached.

## Records already on disk

No migration, and none needed. A record with a `createdAt` keeps it. A record
without one keeps none — it holds the tail position it already holds instead of
acquiring a wrong value the first time anything writes to it. The desktop
sessions list already displayed `createdAt ?? updatedAt` for these rows and is
unchanged.

The one behaviour that is genuinely narrowed: a legacy transcript no longer
*acquires* a creation time by being written to. It never had a true one, and the
value it used to acquire was false.

## What is still not covered, stated plainly

A transcript this process has **never** read successfully has nothing to carry
forward, so it sits in the tail block until a scan can read it, and then takes
its real position. There is nothing to derive a key from for a file whose
contents cannot be reached, and inventing one is the move this PR removes. It is
one stable position for as long as the file stays unreadable, and it is
documented in `docs/api/client-v1.md`.

## Verification

- **Conformance, end to end.** After: `104 passed, 0 failed, 0 skipped` with
  `--include-ttl`, recorded at `d64ab964` in
  `docs/client-v1-conformance-results/2026-08-23-v0.3.9-win32-cave-wbxcu.json`
  (taken from a clean tree after the code commit, as the runbook requires).
  Before: `99 passed, 3 failed`.
- **Mutation matrix.** Nine unit mutations and five build-and-run mutations,
  every new assertion caught by at least one. Two build-and-run mutations were
  first written in a form that failed `pnpm build` (`noUnusedParameters`), which
  prints nothing and reads exactly like a miss; both were rewritten to mutate a
  behaviour while leaving every symbol used, and the trap is written into the
  runbook. One unit mutation — remembering `conv.createdAt ?? conv.updatedAt` —
  **was not caught** by the first draft of the store tests, and two cases were
  added for it.
- **Gates.** `pnpm typecheck`, `pnpm lint`, `pnpm check:tests-wired` and
  `pnpm build` all exit 0. Suites green: `cave-conversations.test.ts`,
  `reads.test.ts` (19), `pagination.test.ts` (17), the five client-v1 route
  suites, `client-v1-doc-contract.test.mjs` (7),
  `client-v1-conformance.test.mjs` (53), and the chat conversation route suites
  (`[id]` 22, `[id]/turns/[turnId]` 14, `conversation/route`).

`docs/api/client-v1.md` no longer claims the tail block can be left, and the
`clientV1ConversationPageKey` / `conversationSortKey` comments describe what the
code now does rather than what it used to.

Closes cave-wbxcu. With this landed, cave-fhjlu has no surviving skip path.

### DIFF
--- a/docs/api/client-v1.md
+++ b/docs/api/client-v1.md
@@ -593,13 +593,15 @@ both or skip both:
 | `/conversations` | `createdAt` descending, then `id` descending. A conversation with no `createdAt` sorts **last**, ordered among its peers by `id` descending. |
 | `/conversations/:id/messages` | Transcript order, oldest first. **Not** a keyset — see that route. |
 
-**Every sort key here is immutable, with one narrow exception named in the third
-bullet.** Nothing rewrites a project's `createdAt` or a conversation's, a
-familiar's id is its identity, and saving a conversation stamps only its
-`updatedAt` — so the position a cursor names is a position the ordering still
-has when you resume. A walk that starts now serves the rows that existed when it
-started, in the order it started with, however much the Cave is written to
-underneath it. Three things are worth planning for:
+**Every sort key here is immutable.** Nothing rewrites a project's `createdAt`
+or a conversation's, a familiar's id is its identity, and saving a conversation
+stamps only its `updatedAt` — so the position a cursor names is a position the
+ordering still has when you resume. A conversation's `createdAt` is decided when
+the record is created and is read-only after that: a client body cannot rewrite
+it, and a transcript that has none is never given one. A walk that starts now
+serves the rows that existed when it started, in the order it started with,
+however much the Cave is written to underneath it. Three things are worth
+planning for:
 
 - **A conversation created while you are paging is not served by that walk.**
   Its `createdAt` is *now*, which sorts above every cursor you already hold.
@@ -609,15 +611,15 @@ underneath it. Three things are worth planning for:
 - **A conversation deleted while you are paging simply stops appearing.** The
   cursor names a position in the ordering rather than an index, so the walk
   continues at the next surviving record.
-- **A conversation that has no `createdAt` at all can leave the tail block.**
-  Such a record is served last (see below), and it stays there for as long as it
-  has no `createdAt` — but the moment it acquires one it sorts above every
-  cursor you hold, so the walk in progress will not serve it. This is the one
-  row this ordering can still miss. It is rare (a transcript older than the
-  field, or one Cave could not read on the last scan) and it settles once the
-  record has a `createdAt`, but a walk cannot tell that it happened. If you must
-  not miss a conversation, re-read from the top and dedupe by `id`; that costs
-  nothing and covers this case.
+- **A conversation that has no `createdAt` at all stays in the tail block.**
+  Such a record is served last (see below), and it stays there: nothing in Cave
+  gives a stored record a `createdAt` it did not already have. It used to —
+  writing a turn to a transcript older than the field stamped it with *now*,
+  which moved the row from the tail to the head, past every open cursor, and the
+  rest of that walk never served it. Fixed on 2026-08-23; if you are talking to
+  an older Cave, that row can still be missed. Re-reading from the top and
+  deduping by `id` costs nothing and covers it, as it covers every other reason
+  a ledger might have moved.
 
 ⚠️ **The ordering of `/conversations` changed.** It was `updatedAt` descending —
 the order the Cave's own sessions list shows — until 2026-08-22. `updatedAt`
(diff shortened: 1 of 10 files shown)


CompleteDotTech:
CI status on head `5652c281d`, run [32618164650](https://github.com/OpenCoven/coven-cave/actions/runs/32618164650).

**Everything except the required `Frontend build` is green:** lint, typecheck, test wiring, protocol conformance, mobile tests, API tests, app tests, Frontend bundle, Select validation.

`Frontend build` failed twice, and the two attempts have different signatures:

| Attempt | Job | Playwright | Ending |
|---|---|---|---|
| 1 | [97142767572](https://github.com/OpenCoven/coven-cave/actions/runs/32618164650/job/97142767572) | 3 spec failures | `The operation was canceled`, no summary count |
| 2 (rerun) | [97144719161](https://github.com/OpenCoven/coven-cave/actions/runs/32618164650/job/97144719161) | `Running 310 tests using 2 workers`, **zero** failures | `The operation was canceled`, no summary count |

Attempt 2 is the `cave-zotrr` infrastructure kill: same `Running 310 tests` line and same zero-failure cancellation measured on unrelated PRs #4887 and #4881. Neither attempt printed an `N failed` summary, because neither run was allowed to finish.

Attempt 1's three failures are not from this change:

- `tests/chat-task-chip-nav.spec.ts:99` (board card inspector dialog), `tests/composer-runtime-chip.spec.ts:120`, `tests/image-carousel-fill.spec.ts:83`. None touches a conversation's `createdAt`, the client-v1 conversations page key, or the conversation store's summary projection.
- Run locally on this branch, `--project=desktop`: **7 passed, 1 flaky** (`image-carousel-fill`, passed on retry). The other two passed outright.
- That job's log is saturated with `[WebServer] Error: aborted` / `uncaughtException: Error: aborted` and a `[cave-home-reconciliation] lock { phase: 'waiting' }` storm, and every failure is `element(s) not found` on a dialog or carousel that renders after a fetch — the shape a degraded web server produces.
- They did not recur on the rerun.

Not pursuing further per the standing instruction: a genuine failure names specs and prints a failure count, and the rerun did neither. Nothing was weakened or skipped to get here.

PR #4889, opened 2026-08-23
fix(settings): harden Client access reconciliation

BunsDev:
## Summary
- bound and coalesce Client access ledger refreshes without starving slow or hung local admin reads
- preserve atomic confirmed snapshots and reconcile terminal pairing conflicts without phantom actions
- disambiguate accessible actions for duplicate installations and guard operator docs against contract drift

## Validation
- pnpm lint
- pnpm typecheck
- pnpm check:tests-wired
- node scripts/export-client-v1-contract.mjs --check
- pnpm test:api
- pnpm test:app
- pnpm test:mobile
- pnpm test:conformance
- pnpm build

Bead: cave-9pifu.4


BunsDev:
Review follow-up is pushed in `6e0b70bb474c365c55e056ccc18e1a42d77e7ee2`.

Addressed:
- memoized the pending-request and credential identity resolvers by their backing collections;
- made the hung-response fixture fail fast when called without an AbortSignal, with a regression test;
- removed the stray assertion whitespace;
- reconciled the branch with current `main`.

Local verification:
- `pnpm lint`
- `pnpm typecheck`
- `pnpm check:tests-wired` (1,838 files)
- client contract export check
- API, app, mobile, and conformance suites
- production build and bundle budgets
- `git diff --check`

The PR body references Bead `cave-9pifu.4`, but that identifier is not present in the current Beads store, so I am not fabricating a tracker close.

PR #4890, opened 2026-08-23
fix: serialize local cave home store locks

BunsDev:
## Summary
- serialize same-process Cave store requests before cross-process lock acquisition
- start each filesystem-lock timeout only after its local queue turn begins
- fail fast on active nested entry and normalize permission lock ordering
- keep Frontend build as the protected aggregator while distributing default Playwright coverage across four independently retryable runners
- isolate flag-enabled agentic journeys, mock runtime inventory in the picker test, and warm the board inspector

## Why
Concurrent Next.js requests could all spend the same 15-second acquisition deadline behind one healthy local store operation, producing ETIMEDOUT responses across unrelated Playwright surfaces. The local queue removes that filesystem stampede while preserving cross-process serialization and stale-lock recovery.

Two exact-head attempts and an independent control run then proved a second failure mode: the monolithic hosted E2E runner was cancelled after early cold-surface failures, including the same runtime-picker and task-inspector assertions. Sharding bounds each runner and makes failed lanes independently retryable without changing the required branch-protection context.

## Verification
- pnpm protocol:tweet-thread:validator:check
- pnpm test: 1,285 app test files passed after the queue and lock-order fix
- all seven cave-home migration suites
- project-permissions and project-grant-audit suites, including non-override repair
- CI workflow and path contract tests
- shard listing: all 310 tests covered; only the four dependency tests repeat per shard
- focused Playwright under CI settings: 7/7 passed for warmup, task-chip navigation, and runtime picker
- pnpm check:tests-wired: all 1,814 test files wired
- pnpm typecheck
- pnpm lint
- git diff --check
- independent review: no findings

Bead: cave-0wzeo
- CI follow-up: fixed the Settings hash hydration race exposed by exact-head sharded CI; warmed Settings/About and kept task navigation assertions bounded by the ordinary 30-second surface budget. CI-mode dependency chain 6/6, settings shell regression, typecheck, lint, workflow contract, and diff check passed; independent review found no findings.


CompleteDotTech:
## Reconciliation with #4914 — this PR is the one to keep, and #4914's lock hunk is withdrawn

#4914 and this PR arrived at the same starvation independently. I measured both
locally rather than guessing, and the result is not close: **this PR is a strict
superset**, so I have withdrawn the store-lock change from #4914 (head is now
`a0b5c45ef`; `git diff main...HEAD --stat` there is two test files and nothing
else). Posting the evidence here rather than only in my own PR, since it is the
argument for landing this one.

### First, a correction I owe this PR

An earlier concern was raised that the wait here is unbounded and would hang —
specifically reading `await previous.catch(() => {})` as the caller's wait.
**That was a misread, and I am withdrawing it.** The stored continuation is
`const tail = previous.catch(() => {}).then(() => turn)`, but the caller's actual
wait is the `Promise.race` immediately below, against a `setTimeout` that rejects
with `ETIMEDOUT` after `RECONCILIATION_STORE_QUEUE_TIMEOUT_MS`. The wait is
bounded, and the timed-out waiter releases its turn — which
`same-process-store-queue-timeout` in this PR already asserts. Correcting that on
the record matters more than leaving it quiet.

### The three-way measurement

Nested `withCaveHomeReconciledStore` on the same store:

| branch | nested-call outcome |
| --- | --- |
| `main` | `ETIMEDOUT` @ 15063 ms |
| #4914 (my lock hunk) | `ETIMEDOUT` @ 15072 ms |
| **#4890 (this PR)** | **`EDEADLK` @ 29 ms** |

Both PRs fix the starvation equally well — 256 concurrent readers go from 185/256
rejected on `main` to 0 rejected on either. That is not the differentiator. The
reentrancy handling is: this PR turns a 15-second silent stall into an immediate,
named, diagnosable error, and #4914 does not.

### The decider: `/api/project-grants` is broken on `main` today

`repairOrphanProjectPermissions()`, reached from
`src/app/api/project-grants/route.ts:98` (`{"repairOrphans": true}`), nests the
global reconciliation lock inside itself: `withProjectRegistryLock` wraps
`withProjectsStore` (which takes the lock) and the inner
`withProjectPermissionsStore` takes the *same* lock again. The second acquisition
can never succeed while the first is held.

| branch | `repairOrphanProjectPermissions()` |
| --- | --- |
| `main` | `ETIMEDOUT` @ 15025 ms |
| #4914 | still broken |
| **this PR** | **resolves @ 24 ms** |

There is no input that avoids it. Hoisting both stores into one
`withCaveHomeReconciledStores` and swapping in `withProjectRegistryMutex` is what
fixes it, and only this PR does that.

I have filed **cave-yn7t5** for the live endpoint defect so it is tracked whether
or not this PR lands — not as a suggestion that anything here should change.

### The inverted lock order in `updateMobileWriteAccess`

Worth calling out because it is easy to read past in the diff. In
`src/lib/project-permissions.ts`, `updateMobileWriteAccess` took the write mutex
*outside* the reconciliation lock — the opposite order from every other caller,
which is the classic setup for a lock-order inversion the moment anything else
holds the reconciliation lock and wants the mutex. This PR flips it so the mutex
is acquired inside the reconciled-store operation, matching the rest. It is
reached from `src/app/api/mobile-permissions/route.ts:81`.

### Test parity

- This PR **passes #4914's test file verbatim.**
- #4914's implementation **fails this PR's test file** at
  `same-process-store-reentrant` — it stalls where this PR raises `EDEADLK`.

### The one thing this PR does not have — offered, take it or leave it

The five tests added here cover queueing, reentrancy, detached handoff,
async-context lease expiry and queue-turn timeout. What none of them cover is the
**starvation property itself**: that N concurrent readers all complete, never
overlap, and none of them reaches the filesystem retry loop. That is the property
the change exists to establish, and it is the one thing a future refactor could
quietly lose while every existing assertion still passes.

#4914 has a 59-line guard for exactly that, and **it passes here unmodified** — I
ran it against this branch. Since #4914's implementation is being dropped, the
test is homeless, so it is offered here. It appends to the same file this PR
already edits (`src/lib/server/cave-home-migration-locks-takeover.test.ts`), just
before the closing `console.log`, and needs `withCaveHomeReconciledStore` (already
imported at the top of that file) and `mkdir` from `node:fs/promises`.

Entirely your call whether it earns its place.

```ts
  // Concurrent store reads are the hot path: every Cave-owned read takes the
  // cross-process lock, and a Next server under an e2e run has dozens in
  // flight. Before the in-process gate each of those ran the rename protocol
  // itself, so every loser burned ~7 filesystem operations per 50ms retry —
  // 64 concurrent `loadConfig()` calls measured 6.6s on an idle machine, and
  // 256 left 204 of them rejected with ETIMEDOUT (cave-18kpz). The gate must
  // keep the exclusion while making a waiter cost nothing, so assert both:
  // the operations never overlap, and no contender ever reaches the retry
  // loop that the "waiting"/"failed" diagnostics report.
  {
    globalThis.__caveHomeMigration = undefined;
    const { cave } = await home("store-gate-concurrency");
    await mkdir(cave, { recursive: true });
    const diagnostics: Array<Record<string, unknown>> = [];
    const warn = console.warn;
    const info = console.info;
    const capture = (...args: unknown[]) => {
      if (String(args[0] ?? "").includes("[cave-home-reconciliation] lock")) {
        diagnostics.push(args[1] as Record<string, unknown>);
      }
    };
    console.warn = capture;
    console.info = capture;
    let active = 0;
    let peak = 0;
    let completed = 0;
    try {
      const readers = Array.from({ length: 32 }, () =>
        withCaveHomeReconciledStore("cave-config.json", async () => {
          active += 1;
          peak = Math.max(peak, active);
          // Yield so an overlapping operation would be observed rather than
          // hidden by a synchronous body.
          await new Promise((resolve) => setImmediate(resolve));
          active -= 1;
          completed += 1;
          return completed;
        }),
      );
      const settled = await Promise.allSettled(readers);
      assert.equal(
        settled.filter((entry) => entry.status === "rejected").length,
        0,
        "every queued store read completes instead of timing out against the lock",
      );
    } finally {
      console.warn = warn;
      console.info = info;
    }
    assert.equal(peak, 1, "the in-process gate still serializes store operations");
    assert.equal(completed, 32);
    assert.deepEqual(
      diagnostics.filter((event) => event?.phase === "waiting" || event?.phase === "failed"),
      [],
      "no store reader enters the filesystem retry loop behind another reader",
    );
    globalThis.__caveHomeMigration = undefined;
  }
```

### One caveat on CI, so nobody reads the checks as a verdict either way

None of the above came from CI, and none of it can right now. `Validate
end-to-end behavior` executed in 38 of the last 100 runs: 14/14 passed before
2026-08-23T01:52Z, 0/24 after 02:07Z, on every branch including `main`.
Throughput fell roughly 8.4s to 28s per test-slot (~3.3x), so the suite exceeds
the cap and the runner kills the job with `exit 143`. Runner image, provisioner
and runner version are identical across that boundary, and heads on the same
lockfile sit on both sides of it. A red check on this PR is not evidence against
it, and a green `Frontend build` on `main` is not evidence either — there the job
skips steps 7-15.

Everything above was measured locally, on this branch at `ff936ef`.

PR #4914, opened 2026-08-23
Make the runtime-chip spec hermetic and warm the `board` chunk (cave-18kpz)

CompleteDotTech:
Two Playwright fixes for `tests/`. Both were measured on their own and stand on
their own; neither of them fixes the red required check, and this PR no longer
claims to.

## What is here

**1. `tests/composer-runtime-chip.spec.ts` mocks `/api/runtime-models/<runtime>`.**

`claude` is a dynamic-inventory runtime: selecting it empties the Model group and
refills it only once `/api/runtime-models/claude` answers (see
`useRuntimeModelInventory`). Every other endpoint this spec touches is mocked via
`page.route`, and the file header claims the whole surface is — but that route was
never intercepted, so the "Claude Sonnet 5" assertion depended on a live server
round-trip that compiles the route on first hit under `next dev` and spawns
`claude --version` behind it.

That assertion is also the only one in the file still on the default 5s expect
budget; its neighbours carry 10-45s. So it is the first thing in the file to break
whenever latency inflates. Reproduced locally and deterministically by delaying
that one route by 6s: the spec fails at exactly that line, on the attempt and on
the retry.

Mocked from the product's own catalog (`catalogForRuntime` and the
`runtimeModelInventory*` helpers), matching the pattern
`tests/chat-code-workflow.spec.ts` already uses, so the seed cannot drift from the
real response shape. No timeout is raised and no assertion is relaxed.

**2. `tests/warmup.setup.ts` warms the `board` chunk.**

`BoardInspector` is a plain static import of `board-view`, which is a
`next/dynamic` chunk, so the Card inspector dialog ships inside the `board` chunk
and the first spec to open it under `next dev` pays the whole Turbopack compile.
Exactly two specs need it (`chat-task-chip-nav`, `task-work-fit`) and whichever
runs first pays. Warmed here once, serially, like the other three chunks: on a
cold `.next` the warm-up step goes 20.1s -> 29.2s, which is that compile being
paid once instead of inside a 10s assertion.

Wired at `playwright.config.ts:134` (already present). No timeout is raised and no
spec is skipped.

## What was withdrawn, and why

An earlier revision of this PR also changed the cave-home store lock in
`src/lib/server/cave-home-reconciliation.ts`. **That change is withdrawn in favour
of #4890**, and this branch now contains no store-lock code at all —
`git diff main...HEAD --stat` is two test files.

#4890 is the better fix, and the decider is not the starvation both of them
address. Measured locally, a nested `withCaveHomeReconciledStore` on the same
store:

| branch | nested-call outcome |
| --- | --- |
| `main` | `ETIMEDOUT` @ 15063 ms |
| this PR's withdrawn hunk | `ETIMEDOUT` @ 15072 ms |
| #4890 | `EDEADLK` @ 29 ms |

`repairOrphanProjectPermissions()` — reached from
`src/app/api/project-grants/route.ts:98` — nests the global reconciliation lock
and so deadlocks unconditionally today. On `main` it fails at `ETIMEDOUT` @
15025 ms; with this PR's withdrawn hunk it was still broken; on #4890 it resolves
in 24 ms. That endpoint is broken on `main` right now and only #4890 fixes it.
#4890 also passes this PR's test file verbatim, while this PR's implementation
**fails** #4890's test file at its reentrancy assertion — it is a strict superset,
not a competing approach.

The one thing #4890 lacks is a guard on the starvation property itself. That
59-line concurrency test has been offered to #4890 directly (it passes there
unmodified) rather than carried here behind an implementation that is being
dropped.

The live `/api/project-grants` deadlock is tracked separately so it does not
depend on #4890 landing.

## What this PR does *not* do

It does not fix the red required check. The `Validate end-to-end behavior` step
executed in 38 of the last 100 runs: 14/14 passed before 2026-08-23T01:52Z, 0/24
after 02:07Z, on every branch including `main`. Throughput fell ~8.4s -> ~28s per
test-slot (~3.3x), the suite exceeds the cap, and the runner kills the job
(`exit 143`). Runner image, provisioner and runner version are identical across
that boundary; heads on the same lockfile sit on both sides of it. That is an
infrastructure incident, not something a spec change can close.

## Verification

- `pnpm typecheck` — clean
- `pnpm lint` — clean (0 files with drift, eslint `--max-warnings=0`)
- `pnpm check:tests-wired` — 1814 test files wired
- `pnpm build` — clean
- `tests/composer-runtime-chip.spec.ts` — passes locally

cave-18kpz

### DIFF
--- a/tests/composer-runtime-chip.spec.ts
+++ b/tests/composer-runtime-chip.spec.ts
@@ -1,4 +1,11 @@
 import { expect, test, type Page } from "@playwright/test";
+import {
+  catalogForRuntime,
+  runtimeModelInventoryAvailability,
+  runtimeModelInventoryFreshness,
+  runtimeModelInventoryRefreshState,
+  runtimeModelInventoryScope,
+} from "../src/lib/runtime-models";
 
 // Verifies the composer runtime picker (cave-yq5l / cave-v25g / cave-bfwk,
 // split-chip grammar since cave-g21f): the chat composer footer's model chip
@@ -46,6 +53,43 @@ async function seed(page: Page): Promise<Mutable> {
     state.familiarsServed += 1;
     return route.fulfill({
       json: { ok: true, familiars: [{ ...FAMILIAR_BASE, harness: state.harness }] },
+    });
+  });
+  // Claude, Copilot, OpenClaw, Grok and Hermes are DYNAMIC-inventory runtimes:
+  // selecting one empties the Model group and refills it only when
+  // `/api/runtime-models/<runtime>` answers (see useRuntimeModelInventory).
+  // Leaving that route unmocked made the "Claude Sonnet 5" assertion below
+  // depend on a live server round-trip — one that compiles the route on first
+  // hit under `next dev` and spawns `claude --version` behind it — inside the
+  // only default (5s) expect budget in this file. That is why this test was
+  // the first thing to fail whenever CI got slow (cave-18kpz). Reproduced by
+  // delaying this route 6s locally: it fails at exactly the assertion below,
+  // on both the attempt and the retry, with the CI message verbatim.
+  //
+  // Mocked from the product's own catalog so the seed cannot drift, matching
+  // the shape a desktop without the runtime CLI installed really returns.
+  await page.route(/\/api\/runtime-models\/([^/?]+)(?:\?.*)?$/, (route) => {
+    expect(route.request().method()).toBe("GET");
+    const url = new URL(route.request().url());
+    const runtime = decodeURIComponent(url.pathname.split("/").pop() ?? "");
+    const catalog = catalogForRuntime(runtime);
+    if (!catalog) {
+      return route.fulfill({ status: 404, json: { ok: false, error: "runtime not found" } });
+    }
+    const provenance = "fallback" as const;
+    return route.fulfill({
+      json: {
+        ok: true,
+        runtime,
+        models: catalog.models,
+        provenance,
+        freshness: runtimeModelInventoryFreshness(provenance),
+        refreshState: runtimeModelInventoryRefreshState(provenance),
+        availability: runtimeModelInventoryAvailability(provenance),
+        defaultOwner: catalog.defaultOwner,
+        allowCustom: catalog.allowCustom,
+        scope: runtimeModelInventoryScope(runtime, url.searchParams.get("familiarId")),
+      },
     });
   });
   await page.route("**/api/sessions/list**", (route) =>
--- a/tests/warmup.setup.ts
+++ b/tests/warmup.setup.ts
@@ -180,12 +180,20 @@ test("warm the code-split surface chunks", async ({ page }) => {
   await reopen.click();
   await expect(page.locator(".workspace-rail")).toBeVisible({ timeout: CHUNK_TIMEOUT });
 
-  // `board` — the last heavy surface this file did not warm, and the one two
-  // specs rotate on (cave-18kpz). `BoardInspector` is a static import of
-  // `board-view`, so the "Card inspector" dialog ships inside the same lazy
-  // chunk: chat-task-chip-nav clicks the task chip and gives that dialog 10s,
-  // against a cold compile the header above measured at 28.3s. Deep-link to
-  // the card so the drawer itself renders, not just the empty board shell.
+  // `board` — the last heavy surface this file did not warm. `BoardInspector`
+  // is a static import of `board-view`, so the "Card inspector" dialog ships
+  // inside the same lazy chunk, and the two specs that need it
+  // (chat-task-chip-nav, task-work-fit) were paying its compile inside their
+  // own budgets — chat-task-chip-nav allows the dialog 10s. Measured on a cold
+  // `.next`: adding this step moves the warmup from 20.1s to 29.2s, which is
+  // that compile being paid here instead of there. Deep-link to the card so
+  // the drawer itself renders, not just the empty board shell.
+  //
+  // ⚠️ This closes a real gap but did NOT fix chat-task-chip-nav on CI
+  // (cave-18kpz): run 32621714561 still failed it with the chunk warmed. That
+  // spec is failing because hosted runners are ~3.3x slower than they were
+  // before 2026-08-23T02:00Z, not because of this chunk — see the bead. Do not
+  // read this step as that spec's fix.
   await page.goto(`/?mode=board#card-${WARMUP_CARD.id}`);
   await page.waitForSelector(".board-shell", { timeout: CHUNK_TIMEOUT });
   await expect(page.getByRole("dialog", { name: "Card inspector" }))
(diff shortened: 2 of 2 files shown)


CompleteDotTech:
Update after CI run `32620280467` on head `0f241c3f4` — that run validated the first two fixes and exposed the third, now pushed as `731607f36`.

**What run 32620280467 proved** (job `97148044769`, step 8 `Install JavaScript dependencies: success`, step 13 `Install Playwright browsers: success`, step 14 `Validate end-to-end behavior` executed, `Running 310 tests using 2 workers`):

- `[cave-home-reconciliation] lock` diagnostic lines in the job log: **0**. Previously hundreds per run. The in-process gate removes the contention outright on CI, not just on my laptop.
- `composer-runtime-chip.spec.ts:120` **did not fail**. It had been in *every* observed failing run, on `main` and three unrelated branches, always at `Timeout: 5000ms`. Gone.
- PR #4890 is the control: it carries the lock fix and not the spec fix, and its run (job `97145165710`) also logged **0** lock diagnostics — and still failed on `composer-runtime-chip.spec.ts:120` at `Timeout: 5000ms`. So the two fixes are independent, and the spec defect is real rather than a starvation symptom.

**What remained**, and what `731607f36` fixes: `chat-task-chip-nav.spec.ts:99`, failing on attempt *and* retry waiting 10s for `dialog[name="Card inspector"]`. `BoardInspector` is a static import of `board-view`, which is a `next/dynamic` chunk — so that dialog ships inside the `board` chunk, and `warmup.setup.ts` warmed `command-palette`, `shortcuts-sheet` and `workspace-rail` but never `board`. Exactly two specs need it (`chat-task-chip-nav`, `task-work-fit`), whichever runs first pays a compile the warmup file's own header measured at 28.3s cold, and both are in the rotating cast. That non-determinism — *which* spec happens to open a cold chunk first — is a large part of why the failing set differed on every attempt.

Warming it costs 20.1s -> 29.2s in the warmup step on a cold `.next`, paid once, serially. Verified cold locally: `chat-task-chip-nav` 2.5s, `task-work-fit` all six cases green, `composer-runtime-chip` green. No timeout raised, no spec skipped.

**Still unresolved and not repo-side.** Run 32620280467 was killed at 05:46:16 by `The runner has received a shutdown signal` after **seven minutes with no test output at all** (last progress 05:39:15), \~13 minutes into a 60-minute cap, with `conclusion: failure` rather than `cancelled`. #4890's run died the same way at 05:25:47. Earlier, two unrelated PRs died 27s apart on consecutively-numbered hosted runners in Azure `centralus`. Zero lock diagnostics in those windows, so this is not the contention this PR fixes — a stall that also freezes the Playwright process is host-level. The mitigations worth taking are in the PR description (shard `desktop`, give the e2e step its own `timeout-minutes`, surface the shutdown explicitly).

CompleteDotTech:
## Root cause established, and it is not this repository

I was wrong to frame the runner teardown as possibly downstream of the lock contention. A controlled measurement across the last 100 CI runs settles it.

The `Validate end-to-end behavior` step actually executed in 38 of them:

| window | runs with e2e executed | passed |
| --- | --- | --- |
| before `2026-08-23T01:52Z` | 14 | **14** |
| after `2026-08-23T02:07Z` | 24 | **0** |

The last healthy run (`32611100288`, job `97124948778`, started 01:52:07Z) finished the whole suite: **289 passed, 1 flaky, 20 skipped, in 21.6 minutes**. The one flaky was `mobile/command-center-pages.spec.ts` (the known `cave-vkb2d`). Both `composer-runtime-chip.spec.ts:120` and `chat-task-chip-nav.spec.ts:99` **passed** in it.

Since 02:07Z, 24 consecutive runs across `main` and every branch have been torn down. None finished.

**The boundary is not code.** `next@16.2.12` heads sit on both sides of it: `6c46fdb5a` passed the full suite at 01:52, and two runs on the same lockfile failed at 02:12 (`97127097837`) and 03:59 (`97138542146`) — each with exactly `chat-task-chip-nav.spec.ts:99` at 10000ms and `composer-runtime-chip.spec.ts:120` at 5000ms, then a shutdown signal. So the Next 16.3.1 bump at 01:52Z is a coincidence of timing, not the cause.

Runner image (`20260816.277.1`), provisioner (`20260729.566`) and runner (`2.336.0`) are byte-identical across the boundary; the Azure region varies (`westus2` on both sides). What changed is throughput: **~8.4 s per test-slot healthy, ~28 s now — about 3.3x slower.** At that rate the 21.6-minute suite needs \~70 minutes, past the 60-minute job cap, and in practice the runner is terminated first (kills observed at 12.75, 14.5, 17.5 and 21.5 minutes, `conclusion: failure`, `exit code 143`).

## This also corrects my reading of the lock

The green control run had **418** `[cave-home-reconciliation] lock` diagnostics, **69** `Error: aborted`, and **25** login-shell probe lines — and passed in 21.6 minutes. Those signals are normal, not the outage. The contention is still a real inefficiency worth removing (measured: 204 of 256 concurrent store reads rejected with `ETIMEDOUT`; this branch takes CI from 418 diagnostics to **0**), and removing it buys headroom against exactly this kind of slowdown — but it was never the trigger, and I should not have implied it might be.

## What still stands

- **`composer-runtime-chip.spec.ts:120`** was a genuine defect: an unmocked live round-trip inside the only default-5s budget in the file. Reproduced deterministically by delaying that route 6 s. Fixed, and absent from both CI runs on this branch — it had been in *every* failing run before.
- **The store-lock gate** is a real robustness win: 418 → 0 diagnostics on CI, independently reproduced by #4890.
- **The `board` warm-up** closes a real gap (+9 s of previously unpaid compile) but, per run `32621714561`, does **not** fix `chat-task-chip-nav`. The code comment now says so explicitly rather than claiming a fix it did not make.

## What I did not do, and why

I did not shard the e2e job in this PR. Sharding is the right durable answer — it takes a healthy run from 21.6 min to ~6 min per shard and limits the blast radius of one lost runner — but at 4 shards the *degraded* runtime is still ~18 min, inside the window where runners are currently being killed, so it would not green the check today. Restructuring the single required context on that guess, during an incident where **no** run completes and therefore nothing can be validated, is not a trade I would make unattended. It should land when a run can be measured again.

**This PR cannot be made green while the incident lasts, and neither can anything else.** Re-run it once full-suite runs start completing again.

CompleteDotTech:
Closing as superseded by #4890, which merged as `aaa814345`.

Both of this PR's remaining changes are already on `main`:

- the `/api/runtime-models/<runtime>` route mock in `tests/composer-runtime-chip.spec.ts`
- the `board`-chunk warm-up in `tests/warmup.setup.ts`

and #4890 additionally sharded the e2e suite four ways, which is what actually cleared the red required check (`Frontend build` now aggregates in ~18s behind `Frontend E2E (1..4/4)` at 6–13 min each).

Verified before closing: `git diff main..HEAD` on this branch is **+102 / −440**, including 79 lines removed from `src/lib/server/cave-home-reconciliation.ts`. Merging it would revert #4890's store-lock serialization and its CI sharding. There is nothing left here to land.

For the record, this PR's lock fix was withdrawn earlier in favour of #4890 after a three-way measurement showed #4890 fails a nested store call fast (`EDEADLK` @ 29 ms) where this branch reproduced `main`'s 15 s `ETIMEDOUT`, and uniquely fixed the unconditional `/api/project-grants` deadlock (`cave-yn7t5`).

PR #4919, opened 2026-08-23
fix(desktop): align offline cache validation with native limits

BunsDev:
## Summary
- measure serialized cache payloads in UTF-8 bytes before invoking the native layer
- reject the full Unicode control-character category in cache scope and entry names
- add regression coverage for non-ASCII control input

## Why
The TypeScript guard must mirror the native Rust limits so valid cache availability is not misreported when only an input is invalid.

## Verification
- resolved both review findings
- exact-head required CI dispatched for the current branch

Bead: `cave-f1k8n`


BunsDev:
Maintainer reconciliation on exact head `e9c1554e90`:

- The UTF-8 payload-length finding is already present in current `main`; the remaining effective PR delta is the Unicode control-category validation plus its regression case.
- The native `char::is_control` boundary is mirrored with `/\p{Cc}/u`, so non-ASCII controls such as U+0085 are rejected before invoking Tauri.
- The separate dirty `fix/cave-f1k8n-review-validation` preservation worktree was not modified.

Delivery evidence: GitHub reports clean mergeability; all 17 exact-head hosted checks are terminal green/skipped as intended; the head signature is verified.

PR #4929, opened 2026-08-23
fix(cave-4x1ag): define --destructive and --shadow-popover on the base palette

CompleteDotTech:
Fixes `cave-4x1ag` (P1).

## The defect

`--destructive` is defined **only** inside six `[data-theme=…]` blocks — ghosty, claymorphism, claude, codex, pastel-dreams, snow — and referenced **19 times with no fallback** across the four Research surfaces. The default Coven palette has no `[data-theme]` block at all, so there, and on tide / ember / slate / contrast / solstice, the custom property resolves to nothing, every declaration naming it is dropped, and the element silently inherits. A **failed** research step was indistinguishable from a healthy one on half the palettes.

`--shadow-popover` had the same shape at smaller scale: defined in claymorphism, pastel-dreams and snow only, with **five** bare references (`cave-chat/session-chrome.css:273`, `familiars-memory.css:366`, `research-reader.css:329`, `settings-familiars.css:344`, `settings-profile.css:957` — the bead counted 2), so those popovers shipped with no shadow on 9 of 12 palettes.

## Reproduced against rendering, not source

`pnpm build`, then a headless Chromium loading the real built CSS chunks and reading `getComputedStyle` over all 12 palettes × 2 modes. Built CSS before the fix held 12 `--destructive` definitions and 4 `--shadow-popover` definitions, none at `:root`.

**Before** — the failed step's node, on the default palette:

| theme | mode | `--destructive` | computed `color` | `border-color` | `background` |
|---|---|---|---|---|---|
| coven | dark | *(UNDEFINED)* | `lab(98.26 0 0)` | `lab(98.26 0 0)` | `rgba(0, 0, 0, 0)` |
| coven | light | *(UNDEFINED)* | `lab(5.20692 0.815533 -2.…)` | same | `rgba(0, 0, 0, 0)` |
| contrast | dark | *(UNDEFINED)* | `rgb(255, 255, 255)` | `rgb(255, 255, 255)` | `rgba(0, 0, 0, 0)` |

The computed `color` equals `document.body`'s in every one of those rows — the declaration is gone, the ink is inherited — and `border-color` falls back to `currentColor`. 12 of 24 combos for `--destructive`; 18 of 24 had `box-shadow: none`.

**After** — all 24 combos painted, on both tokens.

## What changed, and why the definition rather than the references

`src/styles/globals/foundations.css`, `:root` only:

```css
--destructive: var(--color-danger);
--destructive-foreground: var(--color-danger-foreground);
--shadow-popover: 0 8px 24px color-mix(in oklch, var(--shadow-color) 40%, transparent);
```

**Definition, not references.** Rewriting the 19 sites to `--color-danger` would have erased six deliberate palette identities — ghosty and claude take destructive to grayscale, claude's light mode to `#141413`, pastel-dreams to a pale rose — and **none** of those palettes overrides `--color-danger`, so the retune would simply have been lost. It would also have left the token still theme-scoped-only for the seven other stylesheets that name it, so the next reference re-introduces the bug.

**The six theme-scoped definitions stay.** A `:root` base plus per-palette overrides is the established shape in this file, and it is precisely how the three prior instances of this exact bug were fixed: `--shadow-color` (cave-8ppoa), `--bg-subtle` (cave-fdt5), `--bg-sunken` / `--danger-*` (cave-309m). Each is a `:root` value with per-palette retunes on top.

**The values are argued from the design language, not picked for looking red.** `docs/coven-design-language.md` §2 names `--color-danger` (red 24) the single solid error token and states *"never hardcode a state color"*; §3's tint recipe derives fill and border from that one token — which is literally the shape the Research surfaces use (`color-mix(in oklch, var(--destructive) 20%, transparent)`). `--destructive` is the shadcn-compatibility name for the same role, and every *other* name in that vocabulary (`--background`, `--card`, `--popover`, `--primary`, `--ring`) already has a `:root` value; `--destructive` was the one hole. Aliasing rather than duplicating a red also inherits the light-mode retune for free — `--color-danger` is already re-tuned under `:root[data-mode="light"]` — which is exactly the shape of `--danger-text`, `--brand` and `--bg-base` above it.

`--shadow-popover` takes its geometry from the shipped palette values (three of the four use an 8px drop; snow's dark `#00000066` is 40% ink, the closest anchor) and its ink from `--shadow-color`, so it tracks every palette and the light-mode retune the way every other `color-mix` consumer of that token does. Elevation here is carried primarily by the lightness ladder (`--bg-elevated`) per the design language; this only reinforces it.

**`--shadow-popover` is handled here rather than split out** because it is the same defect in the same file with the same fix and the same gate interaction — same bead, same commit.

## Cross-palette evidence — the six existing palettes are byte-identical

Rebuilt, re-ran the same harness, diffed the two tables row by row.

`--destructive` — **identical** before/after (these must not move):

```
claude dark #dc2626        claude light #141413
claymorphism dark #ef4444  claymorphism light #dc2626
codex dark #dc2626         codex light #b91c1c
ghosty dark #4d4d4d        ghosty light #999
pastel-dreams dark #fca5a5 pastel-dreams light #fca5a5
snow dark #991b1b          snow light #b91c1c
```

Changed: exactly `coven`, `tide`, `ember`, `slate`, `contrast`, `solstice` × dark/light — the 12 rows that were unpainted.

`--shadow-popover` — **identical** before/after: claymorphism dark+light `2px 2px 10px 4px #95959d2e`, pastel-dreams dark+light `0 8px 16px -4px #00000014`, snow dark `0 8px 20px 0 #0006`, snow light `0 4px 16px 0 #03152d0a`. Changed: the other 18 rows, `none` → `oklch(0 0 none / 0.4) 0px 8px 24px 0px` (dark) / `oklch(0.449999 0.00998794 none / 0.4) 0px 8px 24px 0px` (light).

`theme-contrast-audit` goes **515 → 527 pairs, 0 failures** — exactly +12, the `--destructive-foreground` on `--destructive` pair now resolving on the six palettes that previously skipped it. No palette needed retuning to clear AA.

## Test added

`src/lib/theme-contrast-audit.test.ts` skips a pair when a token does not resolve, commented *"token not used by this theme"* — it cannot tell that apart from *"token defined nowhere"*. That is why a fully green audit coexisted with this bug for as long as it did. A new block asserts `--destructive`, `--destructive-foreground` and `--shadow-popover` resolve on **every** palette × mode.

Mutation matrix (control → mutate → assert failure → restore → control):

| # | Mutation | Result |
|---|---|---|
| M0 | control, fixed tree | **pass**, exit 0 |
| M1 | delete `--destructive` from `:root` | **fail**, exit 1 — names exactly `coven, tide, ember, slate, contrast, solstice` × dark/light (12 entries) |
| M2 | delete `--destructive-foreground` from `:root` | **fail**, exit 1 — same 12 |
| M3 | delete `--shadow-popover` from `:root` | **fail**, exit 1 — 18 entries: the 12 above plus `ghosty`, `claude`, `codex` × dark/light |
| M4 | restore the whole pre-fix `foundations.css` | **fail**, exit 1 — all three tokens named |
| — | restore | **pass**, exit 0, `git diff` back to the intended two files |

`design-token-drift` is unchanged by this PR: `space=1606` both before and after, `hex=0`, codemod still a no-op over 132 CSS files. (It prints an unrelated pre-existing `1606 < baseline 1607` progress nudge on `main` too; not banked here, since it is not this PR's progress to claim.)

## Interaction with #4927 (`check:tokens:defined`) — not merged as of this writing

#4927 is **OPEN** at `84d1501`, so this PR is based on `main` (`7052192`) without it.

I ran #4927's scanner unmodified against this tree, both with and without the fix:

```
WITHOUT fix:  110 undefined, 28 theme-scoped-only, 19 runtime-constructed
WITH    fix:  110 undefined,  0 theme-scoped-only, 19 runtime-constructed
```

and with the fix it prints, exit 0:

```
[token-refs] token defined only inside a [data-theme]/[data-mode] block: --destructive is gone
  (banked at 19) — remove it from the bank in scripts/design-system/token-reference-scan.mjs.
[token-refs] token defined only inside a [data-theme]/[data-mode] block: --shadow-popover is gone
  (banked at 9) — remove it from the bank in scripts/design-system/token-reference-scan.mjs.
```

**So for whoever merges second:** both banked counts go **down to zero**; nothing goes up, and nothing is raised to make a gate pass. `assessTokenReferences` classifies "is gone" as *progress*, not a *failure*, so `check:tokens:defined` still exits 0 either way — it does not block, it asks. The follow-up in whichever PR lands second is to delete both entries from `BANKED_THEME_SCOPED_ONLY`, leaving that map empty. If #4927 merges first I am happy to do that removal here on a rebase; if this merges first, #4927 should ship the map empty.

`BANKED_UNDEFINED_NO_FALLBACK`, `BANKED_UNDEFINED_WITH_FALLBACK` and `BANKED_DYNAMIC_SITES` are untouched — the undefined count stays at 110 and the dynamic-site count at 19.

## Gates

```
pnpm typecheck        exit 0   tsc --noEmit
pnpm lint             exit 0   tokenize-tsx-design: 0 file(s) with drift; eslint --max-warnings=0
pnpm check:tests-wired exit 0  ✓ all 1814 test files wired into CI
pnpm build            exit 0   ✓ Compiled successfully; bundle-budget within budget
                               (root CSS 614 KB / 640 KB; first-load CSS 918 KB / 940 KB)
```

Also green, each run via `nodeArgsFor`: `design-token-drift`, `theme-contrast-audit`, `tokenize-tsx-design`, `theme-token-hex`, `settings-appearance`, and `research-tab-desk` / `-library` / `-prompt` / `-resources` / `research-desk-view`.


CompleteDotTech:
Note on the head: `main` was merged into this branch from outside this session (`e2afd09` — a `Merge branch 'main' into fix/cave-4x1ag-destructive-token`, bringing in #4921 and #4887). I fast-forwarded rather than force-pushing, and re-verified rather than assuming:

- `git diff 6f0a4be8 HEAD -- src/styles/globals/foundations.css src/lib/theme-contrast-audit.test.ts` is **empty** — the merge did not touch either file.
- The PR's net diff against `merge-base(origin/main, HEAD)` is still exactly those two files, **63 insertions, 0 deletions**.
- All four gates re-run green on the merged head: `pnpm typecheck`, `pnpm lint`, `pnpm check:tests-wired` (1814 wired), `pnpm build` (root CSS 615 KB / 640 KB, first-load CSS 919 KB / 940 KB).
- `theme-contrast-audit` still 527 pairs / 0 failures; the browser harness re-run against the *merged head's* built chunks reports **0 unpainted rows and 0 `box-shadow: none` rows** across all 12 palettes × 2 modes.

One number in the PR body shifted, and not because of this change: `design-token-drift` now reads `space=1603 radius=230` (was `1606`/`232`) — that movement arrived with the merged commits, not from here, so I left the baselines alone rather than banking someone else's progress.

PR #4940, opened 2026-08-23
feat(code): Work scheduler from the Cody Code Reading v2 handoff (cave-7c329)

### LINKED ISSUE #4900: Build the Work scheduler surface from Cody Code Reading v2 (lanes, queue, gates, history)
# Build the Work scheduler surface from Cody Code Reading v2 (lanes, queue, gates, history)

Third of three parts of the (Started) Coven Cave Code handoff. Part 1 (the
Coding Room rebuild) landed in PR #4423 / cave-0rcku.

FRAME: the `isWork` half of Cody Code Reading v2.dc.html. Nothing in src matches
it today — grep for "LANE CAPACITY" / "PSYCHE" returns nothing.

SHAPE: a lane-capacity strip (one tile per familiar: initial, name, state dot +
word, queued count, load bar, note) over a two-pane body — a rank-ordered queue
table (#, BEAD, TASK, FAMILIAR, STATE, ⋯) with drag reorder, a reassign menu and
a row action menu, beside a rail of Gates | History | Bead with a reserved,
explicitly-not-connected "PSYCHE" slot at the bottom.

THE HARD PART IS HONESTY, NOT LAYOUT. What the frame shows vs what exists:

  REAL TODAY
  - queue rows            GET /api/beads?mode=ready  (bd ready --json)
  - reassign              POST /api/beads {action:"claim", assignee}  (cave-p63a)
  - lanes                 familiars + which have a running session
  - bead detail           GET /api/beads?mode=show&id=
  - "why it sits here"    the repo's own orchestration-ready-task contract
                          (docs/orchestration-ready-tasks.md): a blocked task
                          MUST carry unresolved dependencies, one named primary
                          blocker and one imperative next step. That IS the
                          frame's gate card, already specified and enforced.
  - scheduler-state menu  derive it (live / stale poll / daemon offline /
                          partial project load), never let the user pick a
                          state the system is not in

  NOT REAL — needs a decision before any of it renders
  - hand-ranked queue     bd has priority (P0–P3), not a rank. A drag that
                          vanishes on reload is a lie. Either map drag onto a
                          new `--priority` action on /api/beads (small, honest,
                          4 bands not N positions) or drop the drag and sort by
                          priority.
  - gate APPROVAL         no backend at all. The honest equivalent is to route
                          to the blocker, not to render an Approve button.
  - lane load %           no source. Render share-of-queue and label it, or omit.
  - audit history + undo  keep a session-local log of actions taken HERE, and
                          offer undo only for what is genuinely reversible
                          (reassign). Never an undo that cannot fire.
  - PSYCHE                ports verbatim — the frame itself marks it "not
                          connected · nothing is wired to it".

ROUTE WORK: GET /api/beads supports only prime|show|ready; a blocked-bead mode
is needed for the gates rail. POST supports create|claim|comment|close.

## Bead metadata

- **Bead ID:** `cave-7c329`

- **Status:** open

- **Priority:** P2

- **Surface:** Missing

- **Authoritative owner:** 68980965+BunsDev@users.noreply.github.com

- **Labels:** None

- **Parent:** None

- **Dependencies:** None

- **Unresolved blockers:** None

## Source of truth

Beads is authoritative. This draft card is a GitHub Teamwork visibility mirror and should be reconciled from the Bead rather than edited as an independent task.

CompleteDotTech:
Implements the `isWork` half of `Cody Code Reading v2.dc.html` (issue #4900 /
`cave-7c329`), the third part of the Coven Cave Code handoff. The room half
landed as #4423 / `cave-0rcku`.

**It is a deliberate subset.** Five of the frame's elements are not here, each
because rendering it would assert something the Cave cannot back. Every cut is
argued below with the evidence that forced it.

## Baseline — what I verified before changing anything

| Issue's claim | Verdict | Evidence |
| --- | --- | --- |
| grep for `LANE CAPACITY` / `PSYCHE` returns nothing | **True**, with one hit | `LANE CAPACITY`: 0 hits repo-wide. `PSYCHE`: 1 hit, in `docs/design-handoff/IMPLEMENTATION-STATUS.md` — the ledger row describing this frame, not an implementation. |
| `GET /api/beads` supports only `prime`/`show`/`ready` | **True** | The `switch (mode)` in `src/app/api/beads/route.ts` had exactly those three cases and a `400 unsupported mode` default. |
| `POST` supports `create`/`claim`/`comment`/`close` | **True** | Same file; `create` is handled before the id requirement, the other three in the action switch. |
| reassign is `POST {action:"claim", assignee}` | **True** | It runs `bd update <id> --assignee <x> --status in_progress --json`. That trailing `--status` turns out to matter — see the undo decision. |
| `docs/orchestration-ready-tasks.md` requires the blocked triple | **True, but about a different store** | The doc is real and says exactly that. It governs **Cave Board tasks** — implemented in `src/lib/task-orchestration.ts` / `src/lib/cave-board-types.ts` / `/api/board`, with `deriveReadiness`, `primaryBlockerId` and a structured `nextStep`. Beads have **none** of it. |
| `bd` has priority `P0–P3` | **False** | `bd priority --help`: "Priority levels: 0 Critical, 1 High, 2 Medium, 3 Low, 4 Backlog" — **five** integer bands, and `bd update --priority` accepts `0-4`. |
| "Nothing in src matches it today" | **True of this frame; not true of the problem** | `src/components/familiar-work-queue-view.tsx` + `src/lib/beads-work-queue.ts` already ship a queue surface on the Board's Queue tab: lanes, per-familiar rollups, priority filters, reassign, bead detail, degradation banner. It is not this frame, but it is close enough that a second queue needed justifying — see below. |

Two findings changed the design.

**1. The gate card's triple is not available on beads.** The issue says the
orchestration-ready-task contract "IS the frame's gate card, already specified
and enforced", and proposes a blocked-bead mode to feed it. Those two halves do
not meet. That contract is enforced on Board cards; `bd blocked --json` returns
`blocked_by: string[]` and `blocked_by_count`, and nothing else of the triple —
no designated primary blocker, no structured next step, and not even the
blockers' titles. Verified against this repository's own tracker: 26 blocked
beads, union of keys `acceptance_criteria, assignee, blocked_by,
blocked_by_count, created_at, created_by, description, design, external_ref, id,
issue_type, labels, metadata, notes, owner, priority, spec_id, started_at,
status, title, updated_at`.

So the gate card here shows what beads actually carry: every unresolved
dependency **named** (joined to its record), a primary blocker **marked as
derived** with its basis stated, and no next step invented.

**2. Reassign is not reversible, so it gets no undo** — reversing the issue's
own example. `{action:"claim", assignee}` runs
`bd update <id> --assignee <x> --status in_progress`. Undoing it needs the
previous assignee *and* the previous status restored; `/api/beads` has no action
that does that, and `bd update --help` has **no unassign flag**, so an
unassigned-to-familiar reassign could not be reversed at all. An Undo button
there would look identical to a working one and do nothing. It is not offered,
and the history row prints the reason.

## What landed

**Route** (`src/app/api/beads/route.ts`)

- `GET ?mode=blocked` runs `bd blocked --json`, then **one** `bd list --id a,b,c --json`
  join so each blocker id resolves to a title, status and priority. One join, not
  N `bd show`s; capped at 200 ids so a pathological graph cannot build an
  unbounded argv; a failed join degrades to `blockers: []` and the card says the
  names are unavailable rather than failing the whole read.
- `POST {action:"priority", id, priority}` runs `bd update <id> --priority <n> --json`,
  integer `0-4` only. Anything else is a `400` that never reaches `bd`.

**Model** (`src/lib/work-scheduler.ts`) — pure: queue order, lanes,
share-of-queue, gate cards, derived scheduler state, and the history/undo model.

**Surface** (`src/components/code-work-scheduler.tsx` and
`src/styles/globals/surface-work-scheduler.css`) — a **Work** tab in the Coding
Room beside Sessions: a familiar lane strip over a queue table, with a
Gates / History / Bead rail.

### Why a second queue surface is not drift

The Board's Queue tab is **PR triage**: its lanes are pull-request states
(checks failing, needs review, ready to merge) and it answers *what is in flight
and what does it need?* This is **scheduling**: its lanes are familiars, and it
answers *who is working, what is next for them, and what is stuck?* It is also
the only surface in the Cave that can see **blocked** beads at all — `/api/beads`
had no mode for them, so the blocked half of the tracker was invisible to every
UI. The distinction is written into the component's header comment so the next
session does not "reconcile" one into the other. If they should merge one day,
that is an owner call.

## The honesty decisions

| Frame element | Decision | Why |
| --- | --- | --- |
| **Hand-ranked queue, drag reorder** | **Cut.** Order is a pure function of the tracker (priority band, then oldest update, then id). Moving a row means picking a priority band from the row menu, which writes `bd update --priority`. | A dragged position vanishes on reload because `bd` stores no rank. Mapping drag onto bands was the issue's alternative; I dropped the *gesture* too, because a drag whose only legal drops are five buckets teaches a precision the data does not have. The table prints why: *"There is no hand-rank — moving a row means changing its priority, which the tracker stores."* Note the bands are **five**, not four: `bd` is 0–4. |
| **Gate APPROVAL** | **Cut.** The card's only action is `Open <blocker-id>`. | No approval backend exists anywhere in the Cave. Routing to the blocker is the honest equivalent — yes. |
| **Lane load %** | **Rendered as share, labelled.** `N queued · X% of queue`, plus a ten-cell meter. | No familiar declares a capacity, so a load percentage has no denominator. "Share of queue" has one and is stated in the text, not hidden in a tooltip. I chose a segmented meter over a bar with an inline width so the surface adds no `style={{}}` and the `inlineTsxStyles` ratchet stays at 280. |
| **Audit history + undo** | **History kept; undo only on priority.** | Priority undo replays a recorded previous band — an exact reversal, proven end to end against a real `bd` database (3, set 0, read 0, undo, read 3). Reassign gets a row and the reason it cannot be reversed. `schedulerLogEntryIsCoherent` pins the invariant: every entry offers a firing undo **xor** states why there is none. The rail also says the log is session-local and not a tracker audit trail. |
| **PSYCHE** | **Not rendered.** | The frame marks it "not connected · nothing is wired to it". A permanently dead slot teaches a reader to ignore part of the surface, and cannot even say what it would become — the label is honest about the wiring and silent about the meaning, which is the worst of both. The reservation is recorded in the design-handoff ledger and in the component header instead. This is the decision I most expect to be overruled; reinstating it is a small change if you want the space held visibly. |
| **Scheduler-state menu** | **Derived, never chosen.** `no-project`, `unavailable`, `stale`, `partial`, `live`. | A picker would let someone assert "live" over data that is not. Staleness outranks a partial read because stale means *everything* on screen may be wrong, while partial names the section that is missing and that section says so itself. |

One more, not in the issue's list: **a lane reports no state at all when the
session roster could not be read** — `presence: null`, rendered as "state
unknown", never "idle". Related: the Work tab fetches
`/api/sessions/list?collapseFamiliarWorkspace=1` itself rather than taking the
room's `context.runtimeState.sessions`, because that context is familiar-scoped
(`code-room.tsx` says so explicitly) and a lane strip built from it would report
every other familiar as idle.

## What I deliberately did NOT build

- **No drag affordance of any kind**, including a band-to-band drag.
- **No Approve control**, and no "request approval" placeholder.
- **No reassign undo**, and no `action:"assign"` route to enable one. Enabling it
  honestly needs a route that restores assignee *and* status, plus an answer for
  clearing an assignee (`bd update` has no flag for it). Say the word and it is a
  small follow-up; I did not want half the beads to show an Undo and the other
  half not, unpredictably.
- **No PSYCHE slot.**
- **No component source-scanner test.** The properties worth guarding here are
  behavioural, and a scan asserting that the word "PSYCHE" is absent, or that a
  reassign handler is defined, would prove nothing. The Playwright spec guards
  the same properties at the DOM instead.
- **Nothing removed or changed on the Board's Queue tab.**

## Mutation matrix

Every assertion added was broken, the failure observed, and the mutant reverted.

### Pure model — `src/lib/work-scheduler.test.ts` (25 tests)

| # | Property guarded | Mutation | Observed |
| --- | --- | --- | --- |
| L1 | queue order is derived from the tracker | drop the `.sort(queueSortKey)` | FAIL *queue order is derived from the beads alone…*, FAIL *a reorder survives a reload…* |
| L2 | lane share = queued/total, not a placeholder | `share()` returns a constant `0.5` | FAIL *every lane figure equals a value derived from the rendered rows*, FAIL *share-of-queue tracks the distribution rather than sitting at a placeholder* |
| L3 | an unread roster yields no state | ignore `sessionsKnown`, always compute presence | FAIL *a lane reports NO state when the session roster was not read* |
| L4 | a gate card names its dependencies | delete the blocker join (`byId.set`) | FAIL *a gate card names its unresolved dependencies from real blocker records*, FAIL *primary blocker is derived and says so…* |
| L5 | primary prefers a blocker that can be worked | `pool = blockers` (drop the actionable preference) | FAIL *a blocker that is itself blocked yields to one that can actually be worked* |
| L6 | never "live" over data it did not read | stop counting a failed queue source | FAIL *scheduler state never reports live over data it did not read* |
| L7 | undo replays the PREVIOUS band | undo carries `input.priority` instead of `previousPriority` | FAIL *an undo that is offered actually reverses the action when fired*, FAIL *a reorder survives a reload…* |
| L8 | reassign offers no undo | give reassign an undo | FAIL *reassign is recorded but never offers an undo, and says why*, FAIL *every log entry the model can produce offers a firing undo or states why not* |

One defect was caught while writing these rather than by mutation: a lone
blocker that was **itself blocked** reported `sole-blocker` and hid the deeper
chain. `none-actionable` now outranks `sole-blocker`.

### Route

`src/app/api/beads/route.test.ts` gained behavioural coverage: the joined
**title** must arrive; exactly one `bd list --id` call; no join when there are no
blocker ids; the exact `["update", id, "--priority", "0", "--json"]` argv; and
`-1`, `5`, `2.5`, `"0"`, `null`, `NaN` each rejected `400` with **zero** `bd`
invocations. **That file cannot run on this Windows machine** — see the
pre-existing failure below — so the mutations were run through a live harness
driving the real route against a real `bd` database in a scratch repo.

| # | Property guarded | Mutation | Observed |
| --- | --- | --- | --- |
| R1 | `mode=blocked` names its blockers | remove the join, return `blockers: []` | `blockers: []`; the assertion on the joined title fails (`actual: 0`) |
| R2 | priority is a validated integer 0–4 | forward `parsed.body.priority` unchecked | `priority -1` returns `502` from `bd` instead of a clean `400`; every rejection assertion fails. Note `"0"` as a *string* would have passed through silently. |

Live evidence, real `bd`, real Dolt database:

```
mode=blocked status: 200
data: [{"id":"proj-38v","blocked_by":["proj-a9l"]}]
blockers: [{"id":"proj-a9l","title":"Provision the signing key","status":"open","priority":1}]
priority now: 3
set 0 -> 200 read: 0
undo to 3 -> 200 read: 3
reject -1 -> 400 priority must be an integer 0-4
reject 5 -> 400 priority must be an integer 0-4
reject "0" -> 400 priority must be an integer 0-4
reject 2.5 -> 400 priority must be an integer 0-4
reject null -> 400 priority must be an integer 0-4
```

### Rendered surface — `tests/code-work-scheduler.spec.ts` (4 tests)

| # | Property guarded | Mutation | Observed |
| --- | --- | --- | --- |
| E1 | the lane figure on screen is the derived one | `percent = 50` | FAIL `toContainText("2 queued · 67% of queue")` |
| E2 | no reorder control exists | add `draggable` to each row | FAIL `toHaveCount(0)` — received 3 |
| E3 | a gate card offers no approval | add an Approve button | FAIL `toHaveCount(0)` — received 1 |
| E4 | the offered undo fires the exact inverse | undo sends `previous + 1` | FAIL deep-equality on the captured POST body |

## Pre-existing failure, proved on the base commit

`src/app/api/beads/route.test.ts` fails on this Windows machine **before any of
my changes**. Its fake `bd` is an extensionless Node script; `execFile("bd", …)`
cannot run it on Windows, falls through to the WSL path in
`src/lib/server/beads-cli.ts`, and hits the real WSL `bd` with no database:

```
AssertionError ... "error":"Command failed: wsl.exe --cd .../project-a -e /usr/bin/env ... bd ready --json
Error: no beads database found"
502 !== 200
```

Proved identical with the two touched files reverted to `f8751740e`: same
assertion, same `502 !== 200`, at the equivalent line (183 pristine, 197 with my
additions). It passes on Linux CI, where the fake harness works. I did not
weaken or skip it.

## Test wiring

`src/lib/work-scheduler.test.ts` is appended to `SUITES.app` **and** to
`ALIAS_LOADER` (its graph reaches `@/lib/presence`, which resolves `@/lib/types`).
Verified through the runner rather than by hand:

```
$ node -e "import('./scripts/run-tests.mjs').then(m=>console.log(m.nodeArgsFor('src/lib/work-scheduler.test.ts').join(' ')))"
--require ./scripts/css-source-contract-hook.cjs --experimental-strip-types --import ./scripts/test-alias-register.mjs src/lib/work-scheduler.test.ts
```

The route test was already wired; the Playwright spec is out of
`check-tests-wired`'s scope by design.

## Gates — verbatim

```
$ pnpm typecheck
> coven-cave@0.3.9 typecheck
> tsc --noEmit
EXIT=0

$ pnpm lint
> pnpm codemod:design:check && pnpm lint:source
> node scripts/codemods/tokenize-tsx-design.mjs --check
[tokenize-tsx-design] checked — 0 file(s) with drift
> eslint "src/lib/**/*.ts" "src/lib/**/*.tsx" "src/app/**/*.ts" "src/app/**/*.tsx" "src/components/**/*.ts" "src/components/**/*.tsx" --max-warnings=0
EXIT=0

$ pnpm check:tests-wired
> node scripts/check-tests-wired.mjs
✓ all 1815 test files wired into CI
EXIT=0

$ pnpm build
...
bundle-budget — root-layout CSS (every route pays this):
     614 KB  TOTAL root CSS  (budget:    640 KB)
    25.7 KB headroom  (4.0%)
bundle-budget — initial / route CSS:
     918 KB  TOTAL first-load CSS across 13 stylesheets  (budget:    940 KB)
    21.5 KB headroom  (2.3%)
✓ bundle-budget: within budget.
standalone-budget: 7709 files, 1857 directories, 434627289 bytes (limits: 12000 files, 503316480 bytes)
✓ standalone-budget: within budget and free of local build roots.
EXIT=0
```

Unit and e2e:

```
$ node --require ./scripts/css-source-contract-hook.cjs --experimental-strip-types --import ./scripts/test-alias-register.mjs src/lib/work-scheduler.test.ts
ℹ tests 25
ℹ pass 25
ℹ fail 0

$ node --experimental-strip-types --import ./scripts/test-alias-register.mjs src/lib/design-token-drift.test.ts
design-token-drift: ok (codemod no-op over 133 css files; ratchets font=138 space=1604 radius=230 hex=0 inline=280)

$ node --require ./scripts/css-source-contract-hook.cjs --experimental-strip-types src/lib/design-handoff-ledger.test.ts
design-handoff-ledger.test.ts OK (70 paths, 0 files, 24 commits)

$ pnpm exec playwright test --project=desktop tests/code-work-scheduler.spec.ts
  ✓ lane figures are derived from the queue, and an unread state is never guessed
  ✓ the queue has no reorder control, and its order is the tracker's
  ✓ a gate card names its blocker, states why one is primary, and offers no approval
  ✓ the undo that is offered fires the exact inverse; the one that is not says why
  8 passed (53.0s)
```

Also green and unchanged: `code-surface-mode`, `code-reading-wiring`,
`settings-github`, `workspace-chat-handoff`, `pending-code-navigation`,
`familiar-work-queue-view`, `beads-work-queue`, `code-surface`.

## Design system

Tokens only; no hex anywhere (`hexOutsideDefinitions` stays 0); no new inline
styles (`inlineTsxStyles` stays 280); the CSS codemod is a no-op over the new
sheet. Every `var(--token)` used is defined on the **bare `:root`** in
`foundations.css`, not only inside a `[data-theme]` block — checked
deliberately, given the `--destructive` / `--chart-*` hazard. Two sub-grid px
values (a 2px meter gap and a 2px primary-blocker rule) carry the sanctioned
`tokens-exempt` marker rather than being hidden as a second declaration on a
line. The lane dot reuses the existing presence vocabulary from
`src/lib/presence.ts`, and the word is always printed beside it, so colour is
never the only channel. The stylesheet is imported by the component, so it
code-splits with the dynamically-imported Work tab (its own 8.6 KB chunk).

## Adjacent bug found and filed

`cave-y2oxb` — a `PopoverSubmenu` inside an `OverflowMenu` cannot be opened by
click: the submenu's trigger carries `role="menuitem"`, and `OverflowMenu`'s
`onBodyClick` closes the whole menu on any click landing on one. No shipped
component pairs the two primitives, which is why it was never caught. This
surface uses flat labelled sections instead of working around it.

Closes #4900


CompleteDotTech:
## Post-merge-base sweep at `9c3f76cb5`

`main` was red all afternoon and `pnpm test:app` bails on the first failure, so
roughly two-thirds of the app suite never executed on this PR. This is the
targeted replacement: every test that references anything this branch touches,
enumerated mechanically and run individually.

**How the list was built.** Needles = each of the 11 touched paths, its
path-minus-extension, and its basename (basenames like `route` / `route.test`
excluded — they match most of the tree and drown the signal; the
parent-qualified form `beads/route.ts` is used instead), plus the symbols this
branch adds or changes (`CODE_TOP_TABS`, `parseCodeDeepLink`, `mode=blocked`,
`usePausablePoll`, …). Matched against all **1875** test files. That yields
**72**; I added **10** explicit extras that a needle scan structurally cannot
catch — tree scanners that enumerate a directory without naming any file
(`route-inventory`, `css-module-order`, `globals.css`, `dead-ui-removal`) and
the endpoints the new surface calls (`queue/readiness`, `sessions/list`,
`familiars`, `beads/overview`).

**Runner branch picked per file**, by parsing `VITEST_TESTS` out of
`scripts/run-tests.mjs` (it is not exported) — 4 files take
`vitest.mjs run`, the rest take `nodeArgsFor`. Using plain `node` on those four
would have manufactured failures.

**Result: 73 pass, 5 fail — none of the 5 is caused by this branch.**

| File | Cause | Proof |
| --- | --- | --- |
| `src/lib/pausable-poll-discipline.test.ts` | **Pre-existing, Windows-only.** `path.relative` yields `components\chat-view.tsx` while `RAW_INTERVAL_ALLOWLIST` keys use `/`, so all 15 allowlisted ticker files report as offenders. | Reproduced identically at merge base `f8751740e` with this branch reverted. Passes on Linux at this head: `ok (12 guarded interval(s), 16 allowlisted ticker/scoped file(s))`. Filed `cave-oc7dh`. The new component contains **no** `setInterval` and is not among the offenders. |
| `src/app/api/beads/route.test.ts` | Environmental: the fake `bd` is an extensionless script `execFile` cannot run on Windows. | Passes on Linux at this head (`beads route.test.ts: ok`). |
| `src/app/api/beads/overview/route.test.ts` | Environmental: `EPERM … symlink` (Windows symlink privilege). | Reproduced identically at merge base. |
| `src/app/api/familiars/route.test.ts` | Environmental: `EACCES listen … coven.sock` (no AF_UNIX bind). | Reproduced identically at merge base. |
| `src/lib/work-scheduler.test.ts` | **My sweep harness**, not the test: `plan.tsv` lacked a trailing newline, so the loop read the path as `work-scheduler.test.tsnode`. | Passes standalone: `pass 25 / fail 0`. |

Every merge-base reproduction was a precise revert of all 11 touched paths
(modified files checked out at `f8751740e`, new files moved aside) and the tree
was verified clean afterwards.

**Notable passes**, all of which the bailed suite never reached:
`src/app/api/api-contracts.test.ts` (the whole-suite gate), `scripts/ci-paths.test.mjs`,
`scripts/ui-consistency.test.mjs`, `src/app/route-inventory.test.ts`,
`src/app/css-module-order.test.ts`, `src/lib/design-token-drift.test.ts`,
`src/lib/design-handoff-ledger.test.ts`, and the four source-regex readers of
`code-view.tsx` (`code-reading-wiring`, `code-surface-mode`, `settings-github`,
`workspace-chat-handoff`).

**On the source-regex tests specifically** — the class that bit #4939. All four
pass, and they pass for the right reason rather than by luck: grepping them for
`githubTab`, `role="tablist"`, `CODE_GITHUB_TABS.map` and `aria-selected`
returns nothing, so **no test describes the region of `code-view.tsx` this
branch edited**. That region is guarded behaviourally instead, by
`tests/code-surface.spec.ts`, which passes. Nothing was restored to an old
spelling and no assertion was deleted.

**Playwright**: `code-surface.spec.ts`, `reader.spec.ts` and
`code-work-scheduler.spec.ts` — 26 passed. `tests/mobile/code-surface-drill-in.spec.ts`
— 6 passed.

**Silent-incompleteness check** (things that break nothing but would be wrong):
`CODE_TOP_TABS` is the single source for `isCodeTopTab` / `normalizeCodeTopTab` /
`parseCodeDeepLink`, there is no `Record<CodeTopTab, …>` and no surface-preference
spec enumerating code tabs, so `?ctab=work` deep-links correctly with nothing
else to update. The new stylesheet follows the established convention —
component-imported like `code-workbench.tsx` → `surface-code-room.css`, not
`@import`ed into `globals.css`, which is reserved for shell-level sheets.

One correction to this PR's own earlier wording: the measured `inlineTsxStyles`
count is **280** against a baseline of **281** — this branch adds zero inline
styles (`git diff f8751740e HEAD -- '*.tsx' | grep -c "^+.*style={{"` → `0`), and
the 1 of slack is pre-existing on `main`.

PR #4952, opened 2026-08-23
Run the iOS XCTests in CI instead of only compiling them (cave-ac372)

CompleteDotTech:
`iOS build` compiled the XCTest bundles and stopped. 685 XCTest methods under `CovenCaveTests/` had never executed once — the check certified that they *parse*, not that they *pass*.

**The first runs that executed them found 23 real failures and a 36-minute hang.** That section is the point of this PR, not a footnote.

## What changed

`.github/workflows/ci.yml`, `ios` job (`macos-15`, already path-gated to `apps/ios/**` and to `ci.yml` itself):

1. **Boot an iOS Simulator** — resolves the newest available iPhone on iOS 18+ and boots it.
2. **Compile the iOS test bundles** — unchanged in purpose, now targeting the concrete destination the test action uses, with a shared explicit `-derivedDataPath`.
3. **Run the iOS XCTest suite** — `xcodebuild test-without-building -only-testing:CovenCaveTests`, with per-test timeouts enabled, writing a result bundle.
4. **Verify the iOS XCTest suite executed** — `if: always()`.

`timeout-minutes` 30 → 45.

## Why step 4 exists, and why it is not a log grep

`xcodebuild` exits 0 when its selector matches nothing. A renamed target or a retired bundle would switch the whole gate off and still report success — the exact defect this bead is about, moved one layer up. A suite can also truncate and exit 0.

So `scripts/ios-xctest-summary.mjs` reads the **result bundle** and asserts on a property: the number of test cases that executed must be at least the number of XCTest methods that exist in the source tree. The floor is derived from the sources, so adding tests never needs a number bumped here, and removing the wiring that runs them fails loudly. It also fails on any failed case and on a non-`Passed` result, and refuses to certify anything if the source scan itself finds nothing.

This is not a stylistic preference. In the run below the console log's own summary line says **"Executed 544 tests, with 47 failures"** while the result bundle says **685 executed, 24 failed** — the log counts one partition and double-reports. A gate that grepped the log would have asserted against a number that is simply wrong.

Simulator selection is `scripts/ios-select-simulator.mjs` rather than a `jq` filter because runtime versions must be compared **numerically**: any lexicographic sort ranks `SimRuntime.iOS-9-0` above `SimRuntime.iOS-26-0`.

## What running them found

**Run [32666989516](https://github.com/OpenCoven/coven-cave/actions/runs/32666989516)** — first execution ever. 140 cases ran in ~3 seconds, then `AppModelProjectContextTests.testNewerStandaloneTasksLoadWinsOverOlderSuccessfulTaskSnapshot` logged `started.` and never returned. It wedged the job for **36 minutes** until the job timeout killed `xcodebuild`, which then never finalized the result bundle — so the run reported **no counts and no failure names**, a `cancelled` step indistinguishable from an infrastructure blip.

That was a hole in the gate, not just in the test. XCTest does not enforce its per-test allowance unless timeouts are enabled, which is why a documented 600s default did not stop a 36-minute hang. The step now passes `-test-timeouts-enabled YES` with a 60s default allowance. The verification script needed no change — it correctly refused to certify the unfinalized bundle rather than reporting success.

**Run [32669522433](https://github.com/OpenCoven/coven-cave/actions/runs/32669522433)** — with timeouts on, the whole suite completed in about **12 seconds**:

```
iOS XCTest execution report
  source test methods    685
  executed               685
  passed                 661
  failed                 24
  result                 Failed
```

- Executed matched the source floor **exactly, 685 of 685** — nothing truncated.
- The hang is now `Test exceeded execution time allowance of 1 minute`, **named**, and the suite continued.
- `testTheGateActuallyFails(): failed - cave-ac372 canary` appears in the failure list.

23 of those 24 are real, pre-existing failures, inventoried in **cave-vz17i**: 19 in `AppModelProjectContextTests` (mostly one `XCTUnwrap nil ChatThread` shape, plus the hang cave-dwy2a and cave-q9nwj), 2 in `FamiliarDashboardClientTests`, 1 in `VoiceSessionContractTests`, 1 in `CaveClientRetryTests`.

The two `FamiliarDashboardClientTests` failures are the #4946 Familiar-hub tests this bead was filed over, so that suspicion is confirmed rather than theoretical. One is a real product bug: a familiar id containing `/` is **double-encoded**, producing `/api/familiars/nova%252Ffamiliar/dashboard` where the contract wants `%2F`.

## Cost and placement

Test *execution* is not the cost — the full suite is ~12 seconds. The job's cost is `build-for-testing` (~11 min) and the simulator boot (~3 min), and `build-for-testing` was already being paid before this PR. So the marginal cost of actually running the tests is the boot plus seconds, and routine PR CI — already path-gated to iOS changes — is the right home. `release.yml` would move the signal away from the change that caused it.

`CovenCaveUITests` is deliberately excluded (`-only-testing:CovenCaveTests`): it drives `XCUIApplication`, minutes per case and the most flake-prone thing this pipeline could own. It keeps its compile gate; where it should run is cave-x2nhf.

## The gate had two silent escape hatches. Both are now closed

The verification script computed

```
executed = passed + failed + skipped + expectedFailures
```

and then gated only `failed`. So the cheapest way to make this very PR green was
to `XCTSkip` the 23 known failures: 661 passed + 23 skipped = 684 executed,
floor satisfied **exactly**, `result: "Passed"`, job green — silently, forever,
with no expiry and nothing anywhere recording that 23 assertions had stopped
being asked. `XCTExpectFailure` produces the identical shape through
`expectedFailures` and additionally leaves `result: "Passed"`.

That was worse than the bug this PR was filed to fix. A gate whose easiest
bypass is invisible is not a gate, and the previous test suite pinned the
behaviour **as intended**. Both counters are now hard failures.

The `-skip-testing:` quarantine this PR originally proposed is unsound for a
different reason and has been dropped: a selector-skipped test yields *no
result at all*, so it never reaches `skippedTests` — it lowers the executed
count instead, which the source-derived floor already rejects. It also means
nothing could ever detect that the underlying repair had landed.

## How the 23 known failures are handled: an expiring, named quarantine

`QUARANTINED_FAILURES` in `scripts/ios-xctest-summary.mjs`, modelled on
`IN_FLIGHT_REPAIRS` in `scripts/check-test-clock-consistency.mjs` (#4943).
It is a ratchet, not a waiver:

- **Every listed test still runs.** Nothing is skipped, nothing is
  `-skip-testing:`-ed out, the floor is untouched. All 684 execute; 661 must
  pass.
- **An unlisted failure is red.** New breakage fails immediately — which is the
  entire point: this turns "684 tests nobody has ever run" into "684 tests run,
  23 enumerated exceptions".
- **A listed test that starts *passing* is red** — "the repair has landed,
  delete the entry". Suppression cannot outlive what it suppresses, and a
  leftover entry can never mask the next failure of that test.
- **Every entry names a bead and a hard expiry.** Past that date the gate fails
  whether or not the test still fails.
- **An ambiguous key is refused.** The result bundle reports a bare,
  unqualified `testFoo()` with no class attached. This tree has 684
  declarations under 683 distinct names — `testRawValuesAreStable` is declared
  in both `DrawerDestinationOrderTests` and `FamiliarHubNavigationTests` — so
  the gate refuses any key that does not resolve to exactly one XCTest method
  in the source tree. The ambiguity is real, not hypothetical.
- **No source index, no deferral.** The registry fails closed rather than
  applying unvalidated keys.

The test step now carries `continue-on-error`, because a run whose only
failures are deferred must not fail the job. That hands the verify step sole
authority, so it is given the test step's `outcome` and cross-checks it:
xcodebuild succeeding while the bundle reports failures, xcodebuild failing
over a spotless bundle, or any other outcome (`skipped`, `cancelled`) all fail
the job. `scripts/ios-build-ci.test.mjs` pins all three couplings.

`scripts/ios-select-simulator.mjs` and `scripts/ios-xctest-summary.mjs` were
also missing from `IOS_PATH`, so editing the gate's own machinery ran every job
*except* the one that would have exercised it. They are in it now.

## Proof

- **The decision procedure is mutation-tested.** 16 mutations of
  `scripts/ios-xctest-summary.mjs` — including removing the skip guard and the
  expected-failure guard, letting a stale entry outlive its repair, accepting an
  ambiguous key, ignoring an expiry date, and dropping the xcodebuild
  cross-check — are each caught by `scripts/ios-xctest-summary.test.mjs`.
  **0 survivors.**
- **The workflow contract is mutation-tested.** 11 mutations of the `ios` job in
  `ci.yml` — dropping `continue-on-error`, dropping the step `id`, dropping
  `--xcodebuild-outcome`, softening the verify step, removing the simulator
  guard's `exit 1`, replacing the verify step with a no-op, a `generic/`
  destination, dropping `-only-testing`, dropping the per-test timeout, dropping
  `if: always()`, and pointing the floor at an empty directory — are each caught
  by `scripts/ios-build-ci.test.mjs`. **0 survivors.**
- **The shipped registry is validated on every runner**, without waiting on a
  15-minute macOS job: well-formed, unexpired, and unambiguous against the real
  Swift tree.
- **The tests demonstrably execute** — 684 of 684, where previously it was zero.
- **The job demonstrably fails on a failing test** — the canary was named in run
  32669522433, then removed.
- **The simulator guard demonstrably fails closed.** In run 32670074219 the
  `No available iPhone simulator…` line in the log is GitHub echoing the script
  body, not the guard firing: the run selected and booted `iPhone 16`
  (`SimRuntime.iOS-26-2`). The `exit 1` path is now pinned by a test.

Everything above was verified on Linux/Windows, since XCTest cannot run there.
The XCTest counts are read from the CI result bundle.

## Merge readiness — read this before merging

**Merge order matters.** #4958 (cave-vz17i) repairs 12 of the 23 entries.
Whichever of #4952 and #4958 lands *second* must first merge `main` and delete
the entries that have gone stale — otherwise the second merge turns `main` red
on the stale-entry rule. That rule is the mechanism working, not a defect, but
it is five seconds of work done in advance versus a red `main` discovered
afterwards. The stale-entry error names the exact entries to remove.

The residue after #4958 is 11: the hang (cave-dwy2a), the seven `openChat`
failures (cave-22b4y), and three others. Each deletes its own entry when fixed.


BunsDev:
The iOS-only v0.3.10 release is now waiting on this PR. `main` already contains #4958 at `e49d555335c0db6f5c5be4179dbc4bc59f3ba762`, while this head is 3 commits behind. Per the merge-order note above, please update the branch from current `main`, remove every quarantine entry that #4958 made stale, and rerun the iOS gate before merge; merging the current head would make `main` red. I will resume the release stamp from the exact merged head once this prerequisite lands.

PR #4958, opened 2026-08-23
Fix 12 of the 23 iOS XCTest failures the suite's first run exposed (cave-vz17i)

CompleteDotTech:
Fixes 12 of the 23 iOS XCTest failures that `cave-ac372` (#4952) exposed the
first time this suite was ever executed. Two are product defects in
`CaveClient.familiarDashboard(id:)`; ten are tests that could never have
passed. Each is fixed on the side that was actually wrong — nothing here
relaxes an assertion to match current behaviour.

## App defects

**The familiar id was percent-encoded twice.** `Self.encodedPathSegment(id)`
builds one escaped path segment, and `request(_:)` was then called without
`pathIsPercentEncoded`, so `appendingPathComponent` escaped the `%` again.
`nova/familiar` went out as `/api/familiars/nova%252Ffamiliar/dashboard`,
which a desktop reads as the literal id `nova%2Ffamiliar` — a dashboard read
for a familiar that does not exist. The avatar routes on this same client
already pass the flag; this read did not. `CaveClient`'s own comment on
`pathSegmentAllowed` warns about exactly this ("a second copy of this rule is
how `nova/familiar` ends up escaped one way here and another way there").

**A 200 whose envelope says `ok:false` was reported as a decode error.** The
`guard payload.ok` sat *after* the strict success decode, and a refusal body
carries none of the success fields — so the client threw
`.decoding("keyNotFound(version)")` and discarded a refusal the body stated
plainly. Beyond the wrong message, that mislabels a retryable outcome as an
unretryable one: `FamiliarDashboardError.isRetryable` is true for
`.unavailable` and false for `.decoding`, so the hub would blame its own
parser and stop. The refusal envelope is now read first; the `ok` guard stays
behind it as the invariant.

## Test defects

**`request.httpBody` is nil inside a `URLProtocol` stub** (2 tests). By the
time a protocol sees the request, `URLSession` has turned the body into
`httpBodyStream`. Both `CaveClientRetryTests.testUpdateTaskProjectPatchesProjectIdThroughBoardHelper`
and `VoiceSessionContractTests.testStartVoiceConversationUsesProjectRootContract`
unwrapped `httpBody` and failed while the client was sending exactly the body
they meant to check. New `URLRequest.bodyDataForTesting()` prefers `httpBody`
and falls back to draining the stream; the body assertions themselves are
unchanged.

**Six `requestOpen(_:)` tests handed the model a `ChatThread` it had never
seen.** `requestOpen` records an id-based `ProjectNavigationIntent` and
re-resolves the thread from `threads` — deliberately, so an intent that has to
wait for hydration reads current state rather than a pinned object. Every
production caller inserts the thread first (`openServerSession`,
`openChat(for:)`, `ChatView`'s composer paths). The tests now seed
`app.threads` the same way. That also lets
`testRequestOpenFailsExplicitlyWhenProjectMetadataIsInvalid` assert the
refusal it is named for instead of the generic "not on this device" one it was
actually getting.

**Two reconnect tests asserted `app.toast == nil` across a reconnection that
deliberately posts "Reconnected to Cave"** — `connectionState`'s `didSet`
announces it whenever the shell returns to `.connected` after a drop, and the
tests force exactly that by calling `configure(host:)` mid-flight. They now
assert the toast is *exactly* that notice. This is not weaker than `nil`:
`showToast` replaces `app.toast` outright, so any navigation-failure toast
raised afterwards would be sitting there instead — which is the thing the
assertion exists to catch.

## Verification

`main`'s `ci.yml` still only compiles the iOS tests, so this PR's own `iOS
build` job proves the suite compiles and nothing more. Execution evidence comes
from a scratch branch that merges #4952's runner into **this PR's exact head**
and dispatches `ci.yml` against it, read from the **result bundle** via
`scripts/ios-xctest-summary.mjs` (the console summary disagrees with the bundle
and is not used):

| | source methods | executed | passed | failed |
|---|---|---|---|---|
| before — #4952 head `a9aa1ed31`, run 32670074219 | 684 | 684 | 661 | **23** |
| after — this PR head `c8968cb63`, run 32681769643 | 684 | 684 | 673 | **11** |

Scratch branch `verify/cave-vz17i-head-c8968cb63` at `2eba29ea7`: it contains
`c8968cb63`, and `git diff c8968cb63 2eba29ea7 -- apps/ios` is empty — so the
measured iOS tree is byte-identical to this PR's.

The 11 that remain are exactly the ones listed under *Left for follow-up*; no
failure moved sideways and nothing new broke. Full numbers, the containment
proof, the named 11, and the mutation matrix — including one mutant that
failed to land and the targeted replacement that did — are in
[this comment](https://github.com/OpenCoven/coven-cave/pull/4958#issuecomment-5389975773).

⚠️ Runs 32674822084 (41 failures) and 32681866850 (13 failures) are
**mutation** runs on scratch branches — deliberately broken code, commit
subject "MUTATION RUN — do not merge". They are not regressions.

## Left for follow-up

- **cave-22b4y** — seven `openChat(for:)` failures. `resolveProjectNavigation`
  answers `.pending` until the `.projects` surface has succeeded in the
  current connection generation, and `openChat` treats pending as failure and
  returns nil. Which side is wrong needs a decision, not a patch; the bead
  carries the full trace, including why two sibling tests pass by accident of
  ordering.
- **cave-jiox8** — three connection-bootstrap failures
  (`testRefreshConnectionWaitsForFirstProjectContextLoadBeforeConnected`,
  `testPendingTaskIntentWaitsForRelocatedBootstrapBeforeHydrating`,
  `testSessionUnlinkRelinkKeepsTheLatestOptimisticState`).
- **cave-dwy2a** — the
  `testNewerStandaloneTasksLoadWinsOverOlderSuccessfulTaskSnapshot` hang, not
  touched here.

Closes nothing on its own: cave-vz17i stays open until the remaining 11 land.

Node gates (`typecheck`, `lint`, `check:tests-wired`, `build`, `test:mobile`)
are not applicable — this changes Swift only.


CompleteDotTech:
## Execution evidence

The `iOS build` job on this PR runs `xcodebuild … build-for-testing` — it
**compiles** the test bundles and never executes them, because `main`'s
`ci.yml` does not yet run the suite (that is #4952). So this PR's own green
proves compilation only. Execution numbers come from the scratch branch
`verify/cave-vz17i-ci-xctest`, which merges #4952's runner into this branch,
read from the result bundle via `scripts/ios-xctest-summary.mjs`.

| | source methods | executed | passed | failed |
|---|---|---|---|---|
| before (#4952 head, run 32670074219) | 684 | 684 | 661 | **23** |
| after (`3c836808d`, run 32674147914 / job 97279570007) | 684 | 684 | 673 | **11** |

23 → 11: the 12 fixes land, and nothing else in the suite regressed
(661 + 12 = 673 passing, executed count unchanged at 684).

### The 11 that remain

Seven `openChat(for:)` — **cave-22b4y**:
`testConcurrentTaskChatOpensReuseOneAuthoritativeThread`,
`testOpenChatForExistingDirectTaskThreadRepairsStaleSessionFamiliarBinding`,
`testOpenChatForExistingTaskGroupThreadPreservesRosterWhenOnlyOneParticipantHasABoundSession`,
`testOpenChatForExistingTaskGroupThreadPreservesRosterWhenSeveralParticipantsHaveBoundSessions`,
`testOpenChatForMissingTaskSessionOpensAuthoritativeSessionAfterRefresh`,
`testOpenChatForTaskSessionFetchesAuthoritativeSessionWhenCacheIsEmpty`,
`testOpenChatForTaskSessionUsesAuthoritativeSessionFamiliarOverTaskAndCaller`.

Three connection-bootstrap — **cave-jiox8**:
`testPendingTaskIntentWaitsForRelocatedBootstrapBeforeHydrating`,
`testRefreshConnectionWaitsForFirstProjectContextLoadBeforeConnected`,
`testSessionUnlinkRelinkKeepsTheLatestOptimisticState`.

One hang — **cave-dwy2a**:
`testNewerStandaloneTasksLoadWinsOverOlderSuccessfulTaskSnapshot`.

## Mutation matrix

Every assertion this PR changed was mutation-checked on `6cd9bd4a4`
("MUTATION RUN — do not merge", run 32674822084 / job 97281104994), which
breaks the guarded behaviour and confirms the test fails. 41 failures against
the clean 11, so 30 mutation-induced failures.

| # | mutation | target assertion | verdict |
|---|---|---|---|
| M1 | `startVoiceConversation`: `projectRoot` → `projectRoot + "/mutant"` | `testStartVoiceConversationUsesProjectRootContract` | **killed** — failed on the value, proving `bodyDataForTesting()` really reads the body |
| M2 | `updateTaskProject`: `["projectId":]` → `["project":]` | `testUpdateTaskProjectPatchesProjectIdThroughBoardHelper` | **killed** — same, on the key |
| M3 | `requestOpenContext(for:)`: `.failure(.invalidProjectMetadata)` → `.success(.unassigned)` | `testRequestOpenFailsExplicitlyWhenProjectMetadataIsInvalid`, `testValidatedOpenContextRejectsMalformedDotSegmentRoots` | **killed** (both) — proves the `app.threads` seeding reaches the *named* refusal |
| M4 | thread branch: drop the `threadsHydrated && surfaceSucceeded` gate → always `.failed(.threadUnavailable)` | `testReconnectPendingSessionHydrationSupersedesStaleGenerationWithoutRetryLoop` | **killed** — got the navigation-failure toast instead of `Reconnected to Cave` |
| M5 | task branch: drop the `.tasks` `surfaceSucceeded` gate → always `.failed(.taskUnavailable)` | `testReconnectPendingTaskHydrationSupersedesStaleGenerationWithoutRetryLoop` | **SURVIVED** — see below |
| M6 | `completeProjectNavigation`: `switchProject(to:)` → no-op | the six `requestOpen` tests | **killed** — all six failed on the alpha/beta context mismatch |

M4 is the one that answers the obvious objection to swapping `XCTAssertNil(app.toast)`
for `assertOnlyReconnectNoticeToast`: it is not a weakening. `showToast` assigns
`toast` outright, so a navigation failure raised after the reconnect replaces the
notice — and under M4 that is exactly what the assertion caught.

### Finding: M5 survived

`testReconnectPendingTaskHydrationSupersedesStaleGenerationWithoutRetryLoop`
passes both clean and mutated, so the task-side half of that assertion change is
**unproven**, not proven safe. Two candidate explanations, not distinguishable
from the batch run:

1. the mutation never reaches the path (it sits under `if navigationGeneration != nil`), making M5 inert here; or
2. `reportProjectNavigationFailure`'s dedup guard (`guard lastProjectNavigationFailure != state`) suppresses the second identical failure, so no toast replaces the notice — which would be a genuine gap.

Worth a single-mutation re-run to separate them. Filed rather than papered over.

CompleteDotTech:
## XCTest execution numbers for head `c8968cb63`

Read from the **result bundle** via `scripts/ios-xctest-summary.mjs`, not the
console log — the console summary is unreliable here (an earlier run printed
*"Executed 543 tests, with 11 failures"* against a bundle reporting 684/11,
because the console counts `XCTestCase` invocations while the bundle counts
source test methods).

| | source methods | executed | passed | failed |
|---|---|---|---|---|
| before — #4952 head `a9aa1ed31`, run [32670074219](https://github.com/OpenCoven/coven-cave/actions/runs/32670074219) job `97269491520` | 684 | 684 | 661 | **23** |
| after — head `c8968cb63`, run [32681769643](https://github.com/OpenCoven/coven-cave/actions/runs/32681769643) job `97299598779` | 684 | 684 | 673 | **11** |

684 rather than 685 because #4952's deliberate canary test was removed before
this baseline. `URLProtocolRequestBody.swift` adds a helper and no test
methods, so the denominator is unchanged on both rows — the 12 tests that moved
all moved from failed to passed.

### Containment

`ci.yml` on `main` still only *compiles* the iOS tests, so execution has to be
measured on a scratch branch that carries #4952's runner. That branch is
`verify/cave-vz17i-head-c8968cb63` at `2eba29ea7`, and it is this PR's head
plus #4952, nothing else:

```
$ git merge-base --is-ancestor c8968cb63 2eba29ea7 && echo CONTAINS
CONTAINS
$ git diff --stat c8968cb63 2eba29ea7 -- apps/ios
(empty)
```

So the measured `apps/ios` tree is byte-identical to this PR's, and the run's
head contains this PR's head. (The earlier run 32674822084, which reported 41
failures, is the **mutation** run described below — its head commit subject is
"MUTATION RUN — do not merge". It measures deliberately broken code and is not
a regression.)

### The 11 that remain

Unchanged from the before-run — no failure moved sideways and nothing new broke.
All three follow-up beads are filed:

**cave-22b4y** — seven `openChat(for:)` failures, all `XCTUnwrap failed: expected non-nil value of type "ChatThread"`:
`testConcurrentTaskChatOpensReuseOneAuthoritativeThread`,
`testOpenChatForExistingDirectTaskThreadRepairsStaleSessionFamiliarBinding`,
`testOpenChatForExistingTaskGroupThreadPreservesRosterWhenOnlyOneParticipantHasABoundSession`,
`testOpenChatForExistingTaskGroupThreadPreservesRosterWhenSeveralParticipantsHaveBoundSessions`,
`testOpenChatForMissingTaskSessionOpensAuthoritativeSessionAfterRefresh`,
`testOpenChatForTaskSessionFetchesAuthoritativeSessionWhenCacheIsEmpty`,
`testOpenChatForTaskSessionUsesAuthoritativeSessionFamiliarOverTaskAndCaller`

**cave-jiox8** — three connection-bootstrap failures:
`testPendingTaskIntentWaitsForRelocatedBootstrapBeforeHydrating` (`("0") is not equal to ("1")`),
`testRefreshConnectionWaitsForFirstProjectContextLoadBeforeConnected` (`("1") is not equal to ("0")`),
`testSessionUnlinkRelinkKeepsTheLatestOptimisticState` (`XCTAssertTrue failed`)

**cave-dwy2a** — `testNewerStandaloneTasksLoadWinsOverOlderSuccessfulTaskSnapshot`
(`Test exceeded execution time allowance of 1 minute`)

### Mutation matrix

Each assertion changed or added here was broken deliberately on a scratch
branch and confirmed to fail **by name**, then restored. Mutation run
[32674822084](https://github.com/OpenCoven/coven-cave/actions/runs/32674822084)
job `97281104994` (684 executed, 643 passed, **41 failed**). Its base carries
the same `apps/ios` tree as this head — `git diff 44c81a1d8 c8968cb63 --
apps/ios` is empty, so the main merge changed nothing under test.

| mutation | assertion it must break | observed failure |
|---|---|---|
| **A1** — `resolveProjectNavigation` returns `.failed(.threadUnavailable)` where it would stay `.pending`, so a warning toast replaces the reconnect notice | `assertOnlyReconnectNoticeToast` (session reconnect test) | `testReconnectPendingSessionHydration…`: `("This chat is not available on this device yet. Refresh Chats and try again.") is not equal to ("Reconnected to Cave")` ✅ |
| **B** — `updateTaskProject` sends `{"project": …}` instead of `{"projectId": …}` | the body assertion behind the new `bodyDataForTesting()` | `testUpdateTaskProjectPatchesProjectIdThroughBoardHelper`: `(["project": "alpha"]) is not equal to (["projectId": "alpha"])` ✅ |
| **C** — `startVoiceConversation` appends `/mutant` to `projectRoot` | the voice body assertion, same helper | `testStartVoiceConversationUsesProjectRootContract`: `projectRoot "/repos/cave/mutant"` vs `"/repos/cave"` ✅ |
| **D** — `completeProjectNavigation` skips `switchProject(to:)` | the `projectContext` assertions in the seeded `requestOpen` tests | `testRequestOpenSwitchesToThreadProjectBeforeOpening`, `…SwitchesToUnassignedForProjectlessThread`, `…CanonicalizesNestedWorktreeRoot…`, `…TreatsUnknownAndDeletedRootsAsUnassigned` all fail on the wrong `ProjectContext` ✅ |
| **E** — `requestOpenContext(for:)` degrades invalid project metadata to `.unassigned` | `testRequestOpenFailsExplicitlyWhenProjectMetadataIsInvalid` | `XCTAssertFalse failed` ✅ |

**B is the one that matters most**, because it is the check that the new
`bodyDataForTesting()` did not simply make the test pass: under the mutant it
reports the *actual mutated body*, so it is genuinely reading the request, not
short-circuiting.

One mutant did **not** land, and I am recording that rather than quietly
dropping it: the task-side twin of A1 left `testReconnectPendingTaskHydration…`
**passing**, so that mutation never reached the assertion and proved nothing
about it. Replaced with a targeted mutant — change the notice text itself from
`"Reconnected to Cave"` to `"Reconnected"`, which both reconnect tests must
catch. Run
[32681866850](https://github.com/OpenCoven/coven-cave/actions/runs/32681866850)
job `97300300218`, **684 executed / 671 passed / 13 failed** — the 11 known
plus exactly these two, and nothing else:

| mutation | assertion it must break | observed failure |
|---|---|---|
| **A3** — reconnect notice text becomes `"Reconnected"` | `assertOnlyReconnectNoticeToast` in **both** reconnect tests | `testReconnectPendingSessionHydration…` and `testReconnectPendingTaskHydration…`: `("Reconnected") is not equal to ("Reconnected to Cave")` ✅ |

A1 and A3 together are the full proof for that helper: A3 shows both call sites
actually read the toast, A1 shows the assertion still fails when a *navigation
failure* toast replaces the notice — which is the regression `XCTAssertNil` was
there to catch and which this replacement had to preserve.

### Gates

Swift-only change, so `pnpm typecheck` / `lint` / `check:tests-wired` / `build`
/ `test:mobile` are **not applicable** — and CI agrees: this PR's path filter
resolves `frontend=false`, which is why the frontend jobs show `skipping` and
no app-suite file count appears. `iOS build` on this PR (`build` +
`build-for-testing`) is green on `c8968cb63`. The scratch branch, which does
touch `scripts/`, ran the full frontend matrix — lint, typecheck, test wiring,
mobile tests, protocol conformance and bundle all passed there.

PR #4980, opened 2026-08-24
fix(desktop): correct the Cave doc and harden the contracts that pin it

CompleteDotTech:
Seventh review of `e553d0ba5` (the squash of #4953) found seven defects, all live on `main` because they were found after it merged. Six are fixed here; the seventh is deliberately left alone and explained below. Every one was re-verified against current `origin/main` before being touched.

Bead: `cave-sgvek`.

## The doc contradiction (P2)

`PortOccupant::Cave` asserted, unqualified, that "a packaged copy answers even though it holds a sidecar token". `Gated`, four lines below in the same enum, says every shipped copy answers 401 — as does the FMEA row in `docs/daemon-connectivity-reliability.md`. A reader of the enum alone could not tell which verdict a second packaged copy reaches today, and that choice is user-visible: `port_conflict_message` prints materially different text per verdict.

The mechanism `Cave` describes is correct; it was missing a temporal qualifier. `trustedLocalBrowserApi` landed in `735823320` (#4874) on 2026-08-22 and v0.3.9 was stamped 2026-08-21, so a copy built since #4874 does land in `Cave` while every packaged copy in the field lands in `Gated`. The earlier correction reached three sites and missed this one, one variant up.

## The contracts (P3 x3)

The two source-text contracts each claimed to be anchored the same way as the other. They were not — `port-contract` closed on `fatal_exit(`, `desktop-reachability` on `fatal_exit(&error)`. Both also claimed the per-arm anchor gave a window that "cannot span two arms at any width". Mutation testing refuted it:

| mutation | old `port-contract` | old `desktop-reachability` | now |
|---|---|---|---|
| rename one arm's `fatal_exit` argument | pass | **fail** (false alarm) | both pass |
| third arm, neither reaps nor releases | fail | **pass** (false negative) | both fail |
| `Failed`/`Cancelled` arm drops the reap | **pass** (false negative) | fail | both fail |
| hoist `release_all_claims()` above the reap | **pass** (false negative) | pass | port-contract fails |

A lazy `[\s\S]*?` runs straight into the next arm as soon as this one stops ending in the closing token. Both files now share one byte-identical tempered token that refuses to cross another `Err(SidecarStartError::`, so the anchor does the work the count assertion was silently doing.

`port-contract` also checked only that `release_all_claims()` appeared somewhere in the arm. Hoisting it above the reap kept the check green while dropping the claim with node still bound — leaving the next copy to probe the port, find a dying orphan, and tell the user to switch to it. It now pins the order.

`the_gated_and_stranger_refusals_prescribe_different_actions` asserted only that the two instructions differ; its own assertion message named the property it never checked. Rewriting `Gated` to prescribe quitting kept the two unequal by one trailing word. It now pins that `Gated` offers switching and `Stranger` never does.

## The Windows dialog (P3, P4)

The CRLF conversion shipped inside `#[cfg(target_os = "windows")]`, undisclosed in its commit message and untested. Routine PR CI compiles `src-tauri` on `ubuntu-latest` only, so nothing in that block is type-checked between merge and release tag — which is the window it shipped in. It moves to `windows_dialog_file_text`, compiled and tested everywhere, with a test pinning idempotence and that no bare LF survives. It now also converts a lone CR, which classic Notepad drops the same way. I first described that as closing a latent gap; a second review showed it is reachable. `SidecarStartError::failed` splices the bounded node output tail into the message verbatim, and node tooling emits `
`-terminated progress lines, so a failed bind after one arrives at the dialog carrying a bare CR. It is a live fix, and the test now feeds a spliced tail rather than only `already_running_message`, which provably cannot carry one.

`report_existing_gui_owner` cited `report_existing_port_owner` for why its body keeps the summary. That argument turns on `already_running_message` being shared with a headingless surface; this body is built inline and rendered nowhere else, so nothing pins it. The comment now says what actually carries over.

## Not fixed, on purpose

`e553d0ba5`'s message tallies the `Gated` correction at four sites where three were real, having folded in a stale-example fix from a different finding. It is immutable history, no artifact on `main` repeats it, and the surviving code comments are correct.

## One thing the contracts still cannot see

Both assertions read order in the source text, so wrapping a call in `if claimed_the_port { ... }` leaves the text ordered while the call may never run. Verified: that mutation passes both files, and the pre-hardening forms missed it identically. It is the ceiling of a source-text contract, not something this change introduced — but both comments now say so, because a comment claiming more than it delivers is the defect this branch exists to fix.

## Verification

- 266 Rust lib tests pass locally on Windows and 229 on CI (Linux compiles fewer `cfg`-gated tests); no new warnings on either. Both new tests are confirmed running and passing in the CI job log — `the_windows_dialog_file_is_wholly_crlf_whatever_the_message_arrives_as` and the tightened `the_gated_and_stranger_refusals_prescribe_different_actions`. That the CRLF test runs on Linux at all is the whole point of lifting the conversion out of the Windows-only block.
- Both contract tests pass under the runner's own argv (`scripts/run-tests.mjs` `nodeArgsFor`). They sit in the `api` suite, which aborts at file 11 on Windows for unrelated reasons, so they were run directly rather than through `pnpm test:api`.
- `pnpm lint` exits 0.
- Eleven mutants run against both files; every expectation held.

## Filed, not fixed here

`cave-jsdls` — three scope/lifetime defects an independent adversarial pass found in the same mechanism, all live: Windows holds the claim after a failed startup with nothing bound (the hole #4953 closed for macOS/Linux, still open on the platform that stays alive after failing); the owner record is never removed on any success path, so a refusal can name a stale or recycled PID; and the claim is keyed per-user while the port is machine-wide, so two logged-in accounts both win and the second still gets the raw `EADDRINUSE`.


BunsDev:
Maintainer review found two small cleanup blockers before merge:

- `CHANGELOG.md` names `cave-q1gb5`, which is not present in the current Beads store; this PR is canonicalized as assigned bead `cave-sgvek`. Please use the live bead ID.
- Both expanded source-contract comments end a sentence with `//` (`matched.//` and `claimed.//`). Please remove the stray slashes so the contract rationale reads cleanly.

The tempered-token/order assertions and cross-platform Windows helper otherwise match the stated defects, and the selected checks are green.

CompleteDotTech:
Both fixed in `843237f3b` (rebased onto your `main` merge, `5af405ffd`).

**Stray slashes — confirmed, my defect.** The appended ceiling paragraph never emitted its leading newline, so `port-contract.test.mjs:181` read `…they matched.//` and `desktop-reachability.test.mjs:210` read `…these same arms.//` (the second is `arms.`, not `claimed.` — the defect is real in both files either way). Both now break correctly. Worth noting these stayed green through every gate because it is all still one valid comment — a fair reminder that these contracts read source text, not meaning, which is the same ceiling the paragraph itself documents.

**Bead ID — our two stores disagree, and I could not verify yours from here.** In this checkout `cave-q1gb5` resolves and `cave-sgvek` does not:

```
$ bd show cave-q1gb5
◐ cave-q1gb5 [BUG] · Seventh-review follow-ups on the sidecar port claim … [P2 · IN_PROGRESS]

$ bd show cave-sgvek
Error fetching cave-sgvek: no issue found matching "cave-sgvek"
```

`cave-sgvek` appears in neither `bd list` nor `.beads/issues.jsonl`. The likely cause is that `cave-q1gb5` was created locally and never synced to `refs/dolt/data`, so it is invisible to you while yours is invisible to me — not that either ID is wrong.

Rather than write an ID neither side can currently resolve, the entry now cites the PR:

```
- Correct the port-claim occupant docs and harden the two contracts that pin them (#4980)
```

That resolves for both of us and matches what most entries in the file already use. Happy to switch it to `cave-sgvek` if you confirm that is the canonical ID — I just did not want to assert an identifier I cannot check, in a PR whose whole subject is claims that outran their evidence. I have left the local bead pointing at this PR so the trail survives either way.

Not running `bd dolt push` to reconcile the stores, since that is shared state and outside what I was asked to change here.

PR #5027, opened 2026-08-25
Put the crons on the calendar

BunsDev:
`calendar-view.tsx` contained no mention of a cron — not the word, not an rrule, not a projection. `Scheduling Spec.dc.html` files exactly that as a **P1**:

> Agenda with 2 items = 80% void; crons invisible; no link back to Rituals

and the frame's whole calendar idea is that active crons project onto it so the surface is never a void.

### What changed

Active crons now project onto whatever window the current view shows — a day for Day, a week for Week, a month for Month, a fortnight for Agenda — as quieter **dashed** rows, so a forecast never reads like something that already happened. A `ritual runs` toggle turns the projection off, and appears only when there is something to project: a dead toggle teaches a reader to ignore part of the toolbar.

### One definition of a schedule

The projection reads recurrences through `recurrenceFromRrule`, extracted from the plan preview shipped in #5026. The calendar therefore cannot draw a cadence the create dialog never promised — two readers with their own RRULE handling is precisely how those surfaces would quietly drift apart.

### Two refusals, both deliberate

- **A paused cron never projects.** It is not going to fire, and drawing its occurrences would put events on a calendar that will never happen.
- **An unreadable rule contributes nothing** (`FREQ=SECONDLY`, a weekly rule with no day) rather than being approximated to some nearby schedule the user did not ask for.

### The cap admits itself

Occurrences over a window are unbounded — an every-5-minutes cron across a month is ~8,600 of them. So the projection caps, and when it truncates the footer says *"showing the first N"*. A surface that silently drops the tail tells the reader the month is quieter than it is, which is the same class of lie as a status that cannot be known.

### The footer

Answers the two questions a projected calendar raises — how much of this is forecast, and where do I go to change it — and renders only when something projects, so an empty calendar gets no line about zero crons. Its `manage crons →` asks by event, mirroring the week ribbon's handoff in the opposite direction, because the calendar is a slot inside `AutomationsView` and does not own the tabs.

It reads the already-warm `schedules:automations` resource rather than adding a fetch that would race the one `AutomationsView` makes.

### Verification

10 tests in `calendar-cron-projection.test.ts` (wired into the app suite), covering the paused refusal, the unreadable-rule refusal, chronological ordering across crons, day bucketing, the truncation report, and inverted/zero-width windows. `pnpm lint`, `tsc --noEmit`, drift ratchets and the production bundle budget all clean.

Driven in a browser before merge — see the verification comment on this PR.

cave-9f3i3

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -744,6 +744,7 @@ export const SUITES = {
     "src/components/calendar-view-polish.test.ts",
     "src/components/calendar-agenda-redesign.test.ts",
     "src/lib/calendar-layout.test.ts",
+    "src/lib/calendar-cron-projection.test.ts",
     "src/components/calendar-overflow-pill.test.ts",
     "src/lib/canvas-layout.test.ts",
     "src/lib/canvas-gallery.test.ts",
--- a/src/components/automations-view.tsx
+++ b/src/components/automations-view.tsx
@@ -798,6 +798,16 @@ export function AutomationsView({ familiars, onNewReminder, onEdit, onOpenLink, 
     onRestore: selectTab,
   });
 
+  // The calendar's projection footer offers "manage crons →". The calendar is
+  // a slot inside this view and does not own the tabs, so it asks by event —
+  // the mirror of the week ribbon's sessionStorage handoff in the other
+  // direction.
+  useEffect(() => {
+    const openCrons = () => selectTab("crons");
+    window.addEventListener("cave:rituals:open-crons", openCrons);
+    return () => window.removeEventListener("cave:rituals:open-crons", openCrons);
+  });
+
   const openCalendarDay = (day: RitualDay) => {
     window.sessionStorage.setItem("cave:calendar:pending-open-date", day.key);
     selectTab("calendar");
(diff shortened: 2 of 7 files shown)


BunsDev:
## Browser verification

Driven against a production build. This is where the real defect turned up — the unit tests could not have caught it.

**The defect.** The calendar's default view is **Week**, and only `AgendaView` renders projected rows. So the footer read *"13 active crons project onto this calendar"* over a view showing **zero** of them, with a `ritual runs` toggle beside it that changed nothing visible. Live in three of the four views.

Fixed in `69a783c82` by scoping the projection, its footer and its toggle to the view that actually draws them.

**After the fix:**

| View | run rows | footer | toggle |
|---|---:|---|---|
| Week | 0 | hidden | hidden |
| Month | 0 | hidden | hidden |
| Agenda | **122** | shown | shown |
| Agenda, runs off | 0 | hidden | shown |
| Agenda, runs back on | **122** | shown | shown |

122 rows across 15 day groups, e.g. `01:00 Control UI Priority Audit · CRON`, `08:30 Coven Task Freshness Daily · CRON`.

`manage crons →` was also confirmed to land on the **Crons** tab. An earlier probe reported "Home", but that was my own measurement reading the wrong tablist — the workspace-level one rather than `[aria-label="Rituals sections"]` — not a bug in the handoff.

**Remaining:** Day, Week and Month renderers are filed as `cave-aa10e`. The guard deliberately stays narrow until each renderer exists — widening it early is exactly the defect above.

PR #5099, opened 2026-08-28
Unify the sidebar and dock the chat list inside the chat surface

BunsDev:
## What

The rail swapped its whole contents when Chat was active: `SidebarMinimal` was replaced by the session list, so the one piece of navigation that should be constant changed identity depending on where you were. Chat and Home were a toggle above it, which made "where am I" a two-part question.

Chat is a destination in one flat list now, and the session list moved into the chat surface as a **docked rail** beside the conversation — mounted at the surface rather than in `ChatList`, so it survives opening a thread, which is exactly when you want to switch threads. Collapsing it leaves a full-height spine carrying a sideways "Sessions" wordmark; the whole header row and the whole spine are the control.

Sections are **titled, not collapsible**. Two short groups did not earn a chevron each, and three interactive-looking headings above a list of genuinely interactive rows read as clutter. Marketplace and Memories get their own **Explore** heading instead of the unlabelled step that used to separate them.

## Also in here

- **Chat drops the `chat-contextual` nav policy.** It justified force-opening the rail and never writing the remembered preference while Chat swapped in a different nav. With one shared sidebar it would only re-open a rail the user deliberately collapsed, under a toggle still labelled "Collapse Chat sidebar".
- **A mobile route back to the list.** Below 1024px the rail *and* its spine are `display: none`, and the list no longer lives in the app sidebar — so it had become unreachable on a phone. It opens as a left-edge sheet, mirroring the code rail's own sheet on the opposite edge.
- **Settings and Dashboard** render through `AnalyticsPageShell`, which built a flat rail of its own — the two pages whose sidebar did not match the rest of the app. Same sections, same registry now.
- **Title-bar alignment.** Only the traffic lights' *leading* inset was ever accounted for, never their *top* inset. Scoped to `[data-tauri-titlebar]`, with a test asserting no other platform inherits it.
- **The wordmark is gone from the title strip.** It sat at flex-start and rendered underneath the traffic lights; the menu bar already names the app.

Search, "Show archived" and the Organize menu went with the old header row. The filtering machinery is still in place behind constants — whether to reintroduce a compact affordance is `cave-mwl1x`.

## Two bugs the new coverage forced out

- **The rail's collapsed state did not survive a reload.** The hydration gate was a `useRef` set inside the hydrate effect, and the persist effect runs *later in the same commit*, where `railOpen` is still the `true` default — so it wrote `true` over the stored `false`. Under StrictMode's mount → unmount → mount the second mount read back that `true`, so the clobber was not self-correcting. It is state now: both updates land in one commit and the first persist is a no-op.
- **Collapsing the sidebar leaves the aside mounted as `inert` + `aria-hidden`** rather than removing it, so the specs assert that contract instead of `toBeHidden()`.

## Unrelated defects fixed here

Both surfaced while validating, and both are one-line reads of a shared contract.

- **Research desk: "Couldn't load saved links." over an empty shelf.** `listCompatibleResearchLinks` validates every manifest on the **read** path, so ONE saved paper repeating an author name at index 220 threw and took **500 healthy links** down with it. `paper.authors` is an ordered credit list, not a set — collaborations repeat a name across affiliations, and arXiv/HF metadata simply contains repeats. Identity filters still reject duplicates.
- **Profile card: "Refresh failed: memory unavailable (local_access_required)".** Canonical memory is an enrichment, but its failure code landed in `errors`, and a non-empty `errors` makes the card **abort the refresh and keep stale data**. Off Cave's own host that code is the permanent, expected answer — so every refresh failed forever and discarded familiars, sessions and profile that had loaded fine. It travels as a notice now. The 403 itself is deliberate and unchanged.

## Verification

- `tsc --noEmit` clean; `pnpm lint` clean (design ratchets, token drift, test-clock checks included).
- Unit: 53 + 46 touched/consumer tests green.
- Both research-desk regression tests confirmed to **fail without the fix**.
- Playwright `chat-sidebar-nav` (rewritten for the new architecture): **15/15 green**.

Note for reviewers: an earlier run showed 8 desktop failures, all of them `next dev` compile contention at 8 workers — every one died waiting on `.shell-frame` to render at all. This spec is sensitive to cold-compile load.


BunsDev:
E2E alignment review update (after reconciling the Home/Chat decision):

The unified-sidebar + docked Sessions architecture is the desired direction. The Home/Chat segmented mode switch should remain retired; Chat is a destination and its conversation navigation lives in the inner Sessions rail.

One architectural blocker still remains on the current PR head before I consider this ready for final review:

1. **Desktop still has two Project/Familiar authorities.** `SidebarRailHeader` on this branch still imports/renders `SidebarScopeSelector` under `rail-header__scope--desktop`, while the persistent titlebar also renders `WorkspaceContextSwitcher`. That recreates exactly the duplicate desktop context ownership we were removing. The desired contract is: titlebar = single desktop Project/Familiar authority; `SidebarRailHeader` keeps the full stacked `WorkspaceContextSwitcher` only as the mobile fallback.
2. **Chat still exposes Project as a second mutable selector.** `ChatView` currently renders `ComposerContextChips` without `showProject={false}`, and `ChatSessionContextRow` still receives `projects`, `projectId`, `onProjectChange`, and `onAddProject`. Once shell context owns Project, Chat should display session/execution project context but should not expose another competing workspace selector. The preserved `fix/cave-i8kbm-chat-context-selector` branch has the intended contract: suppress the redundant composer project chip and make the session project context read-only/legible.

Please port those semantics onto this current branch rather than merging the old WIP branch wholesale. Keep the good parts already here: unified global nav, retired Home/Chat toggle, docked desktop Sessions rail + persistent spine, mobile Sessions sheet, the corrected rail persistence hydration, and the resolved dead ChatList API cleanup.

I also reviewed the two unrelated regression fixes bundled here: allowing duplicate paper authors is narrowly scoped to the ordered `paper.authors` list while preserving uniqueness for identity/filter arrays; canonical-memory unavailability is correctly demoted to a notice so an optional local-only enrichment cannot poison an otherwise successful profile refresh.

Once the two context-authority items above are corrected and exact-head CI is green, this is the shape I want to put in front of final human review.

PR #5131, opened 2026-08-28
feat(ios): reconcile server archived/pinned onto local thread flags (cave-sve2a)

CompleteDotTech:
## Bead
cave-sve2a — ios: reconcile server archived/pinned back onto local thread flags

## Scope
Read side of the archive/pin/delete flow (write side: PR #4271 server, #4278 iOS). A change made on another client now appears on iOS: on every applied session load, AppModel maps each thread's sessionIds to their SessionRows and folds archived/pinned back onto the ChatThread flags.

## Canonical status
Coordination designates this PR as the canonical read-side implementation. It replaces the read-side pieces that landed in #5136 (cave-ioswipe.6):  and  are removed in favor of  + . The project-context selection guard from #5136 (archived rows must not tip selection) is retained. #5136 keeps its server-persistence/group-delete scope.

## Decisions
- **Archive = ALL**: a thread is archived when all of its owned sessions are archived (any live session keeps it visible).
- **Pin = ANY**: a single pinned session pins the thread.
- **Local-intent-wins in flight**: a thread with a flag PATCH still in flight (threadFlagWrites) keeps its optimistic local value until the next load after the PATCH settles.
- **No server opinion left alone**: threads owning no sessions, or whose sessions have no row in this fetch, are untouched.
- **Fetch includes archived rows** (includeArchived: true) so the fold can see them; serverSessions stays the active-only view consumers already expect.

## Files
- apps/ios/CovenCave/CovenCave/Models/Models.swift — SessionRow decodes pinned
- apps/ios/CovenCave/CovenCave/State/AppModel.swift — reconcileServerFlagsOntoThreads, protocol includeArchived, applyLoadedSessions fold; removes 5136's sessionsIncludingArchived/reconcileThreadFlags
- apps/ios/CovenCave/CovenCave/Networking/CaveClient.swift — removes 5136's sessionsIncludingArchived()
- apps/ios/CovenCave/CovenCaveTests/* — protocol conformer signatures
- scripts/ios-server-flag-reconcile.test.mjs — source-text contract test (wired into run-tests.mjs; supersedes 5136's ios-thread-flag-reconciliation.test.mjs)

## Tests
- scripts/ios-server-flag-reconcile.test.mjs — ok
- ios-thread-server-persistence, ios-pin-threads, ios-archive-threads, ios-session-refetch, ios-auto-reconnect — all pass
- node scripts/check-tests-wired.mjs — ok
- pnpm exec tsc --noEmit — clean


CompleteDotTech:
Superseded: the read-side reconcile landed via #5136 (merged 9efa9f65e, archived-inclusive fetch + flag folding). Keeping one implementation.

PR #5146, opened 2026-08-28
Patrol surfaces stale worktree locks distinctly from live ones (cave-eg1ag)

CompleteDotTech:
Bead: cave-eg1ag

## What

The lifecycle patrol reported a locked unit purely by its cleanliness/retention
lane, so a lock whose stated risk still held and a lock contradicted by current
state read identically. PR #4720 already released the auto-lock hook's *own*
locks at source; the uncovered case is a lock the hook did not write
(e.g. `active cave-1c8zf PR completion` still held after the bead closed).

Those foreign locks must never be auto-released — the reason is a claim the
hook cannot evaluate — but the patrol should **name** them so a human can clear
them, instead of leaving an operator to diff lock reasons against merged PRs by
hand.

## Change

- `scripts/worktree-lifecycle-inventory.ts` — `parseWorktrees` now captures each
  worktree's `locked <reason>` field from `git worktree list --porcelain -z`,
  and the inventory threads it onto the observation as `lockReason`
  (`null` = not locked, `""` = locked with no reason recorded).
- `src/lib/worktree-lifecycle.ts` — `WorktreeLifecycleObservation.lockReason`
  (optional, legacy-safe); the human report renders `worktree locked: <reason>`
  directly under each locked unit's header, alongside its live verdict. The lock
  is never a classification input, so a stale lock still shows as
  `cleanup-ready` — which is exactly what makes the contradiction visible at a
  glance.

The retirement apply path's release/non-release boundary is unchanged: a
hook-written auto-lock is still released once its risk is disproven, and a
foreign lock is still left in place and reported.

## Tests

- `src/lib/worktree-lifecycle.test.ts` — renderer: clean/retained + foreign lock
  (named, reason shown, verdict intact), dirty + lock (named, reason shown),
  hook auto-lock, bare lock (`(no reason recorded)`), and unlocked units (no
  lock line).
- `scripts/worktree-lifecycle-patrol.test.mjs` — end-to-end: locks the clean
  `feat/old` and dirty `feat/live` fixtures with a foreign reason and asserts
  the JSON report carries `lockReason` and the human report names it, without
  ever changing the lane or releasing the lock.

## Verification

- `pnpm typecheck` — green
- `pnpm lint` — green
- `pnpm check:tests-wired` — green (1921 files)
- `pnpm test:app` — green (1359 test files passed)

### DIFF
--- a/scripts/worktree-lifecycle-inventory.ts
+++ b/scripts/worktree-lifecycle-inventory.ts
@@ -141,6 +141,11 @@ type WorktreeEntry = {
   ref: string | null;
   branch: string | null;
   refError: string | null;
+  /**
+   * The worktree's lock reason from `git worktree list --porcelain -z`.
+   * `null` = not locked; `""` = locked with no reason recorded.
+   */
+  lockReason: string | null;
 };
 
 type LocalBranchRef = {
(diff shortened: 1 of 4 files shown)


CompleteDotTech:
Reviewed while resolving the duplicate: this inventory-level approach is the right shape (lockReason rides the observation, never a lane input). One optional addition matching the cave-no5nr/cave-qamzg precedent of naming the remedy: when the unit's lane is retire-after-gate/cooldown AND the reason is not the hook's own auto-lock, append the exact remedy (confirm with owner, then 'git worktree unlock <path>') — otherwise an operator still has to remember the command. Non-blocking either way.

PR #5149, opened 2026-08-28
fix(worktrees): surface stale locks beside live patrol verdicts (cave-eg1ag)

CompleteDotTech:
## cave-eg1ag: Patrol should surface stale worktree locks distinctly from live ones

Part (a) of cave-a245b. PR #4720 already fixes the machine-generated case at
source — the auto-lock hook releases its own locks once the tree is clean and
the head is retained. What remains uncovered is a lock the hook did NOT write,
e.g. `active cave-1c8zf PR completion` on `fix/cave-1c8zf-compact-tool-rollup-pr`,
held after PR #4685 merged. Those are never auto-released (the reason is a claim
the tool cannot evaluate), but the patrol should NAME them so a human can clear
them — instead of diffing lock reasons against merged PRs by hand.

### What changed

- **`renderWorktreeLifecycleReport`** (src/lib/worktree-lifecycle.ts): optional
  `lockReasons` map + `isAutoLockReason` predicate. Every locked unit prints its
  lock reason beside its live verdict (`  lock: "…"`). When the unit's lane
  already proves clean + retained (`retire-after-gate` or `cooldown`), the lock's
  stated risk is contradicted by current state and gets a distinct callout:
  - foreign lock → `STALE FOREIGN LOCK — … confirm with its owner, then: git worktree unlock <path>`
  - hook's own lock → `stale auto-lock — … the hook re-evaluates and releases its own locks, or --apply resolves it`
  A lock never changes a lane; this is presentation over the same classification.
- **Patrol** (scripts/worktree-lifecycle-patrol.ts): reads the lock census once
  per run from `git worktree list --porcelain`, adds a `Locked worktrees (N)`
  section to the read-only report, and emits `worktreeLocks` (path/reason/isAutoLock)
  in `--json`. A failed census prints `Lock census unavailable: …` and does not
  fail the patrol.
- **Retirement script**: `readWorktreeLocks` / `isAutoLockReason` exported so the
  patrol shares the exact reason-prefix contract with the auto-lock hook
  (the existing `worktree-lifecycle-retirement.test.mjs` cross-file assertion
  continues to guard drift).

### Verification

- `src/lib/worktree-lifecycle.test.ts`: new renderer contract block — foreign
  stale lock callout with exact unlock command, auto-lock callout, live-unit lock
  with no stale marker, and no lock lines when the census is absent.
- `scripts/worktree-lifecycle-patrol.test.mjs`: pass (patrol suite).
- `pnpm typecheck`: clean.
- `git diff --check`: clean.

No lane inputs or retirement behavior changed; the apply path's foreign-lock
safety boundary (reported, never released) is untouched.

### DIFF
--- a/scripts/worktree-lifecycle-patrol.ts
+++ b/scripts/worktree-lifecycle-patrol.ts
@@ -23,6 +23,8 @@ import {
   createGitRetirementOperations,
   parseMaxRetire,
   retireLifecycleUnits,
+  readWorktreeLocks,
+  isAutoLockReason,
 } from "./worktree-lifecycle-retirement.ts";
 import {
   createMetadataRepairOperations,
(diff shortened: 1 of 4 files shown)


CompleteDotTech:
Superseded by #5146, which threads the lock reason through the inventory observation (every consumer sees it) instead of a patrol-side census. Same bead (cave-eg1ag), same evidence; keeping one implementation avoids a double report. Closing this one.

PR #5170, opened 2026-08-28
fix(passkey): verify WebAuthn attestation statements (cave-01v4u)

CompleteDotTech:
## What

`verifyRegistration()` recorded `attestationObject.fmt` but never checked the attestation statement, so a software authenticator asserting the UV flag without a biometric check was indistinguishable from a Secure Enclave one. This adds `verifyAttestationStatement(...)`, called from `verifyRegistration` AFTER the existing credential-key and UV checks.

## Format matrix (fail closed)

| fmt | disposition |
| --- | --- |
| `none` | accept, recorded `attestationVerified: "none-equivalent"` (nothing to verify) |
| `apple` (empty attStmt) | accept, recorded `"none-equivalent"` (Apple's anonymous platform attestation == none; no cert to walk) |
| `apple` (x5c) | verify `sig` over `authData \|\| clientDataHash` with the leaf key; walk x5c to a pinned root |
| `packed` (x5c) | same as `apple` x5c |
| `packed` (self, no x5c) | verify `sig` over `authData \|\| clientDataHash` with the **credential** key; recorded `"none-equivalent"` (proves nothing about the model) |
| `fido-u2f` | verify `0x00 \|\| rpIdHash \|\| clientDataHash \|\| credentialId \|\| point`, walk x5c to a pinned root, and require the x5c leaf key == credential key |
| `tpm` / `android-key` / `android-safetynet` | **refused** with the format named (no per-format verifier; fail closed rather than record an unverified statement as verified) |
| unknown | **refused** |

Pinned roots: **Apple WebAuthn Root CA (G1)** and **Apple App Attestation Root CA**, embedded as constants. Chains that do not reach a pinned root are refused.

`RegistrationResult` keeps `attestationFormat` as the RECORDED format and adds `attestationVerified: true | "none-equivalent"`, persisted additively in `passkey-store.ts` (`attestationVerified?`, so pre-existing credentials keep loading).

## Test evidence

`src/lib/server/webauthn-verify.test.ts` — in-test fixtures (throwaway CA + leaf cert generated once with OpenSSL, EC P-256; leaf private key stored as a JWK so it is not mistaken for a real credential):

- `none` / `apple`-empty → `"none-equivalent"`, never `true`
- packed x5c → pinned root: **accepted**
- packed x5c → unknown root: **refused** (`attestation`)
- packed self-attestation: verified against the credential key, `"none-equivalent"`
- packed self-attestation with wrong signature: **refused** (`signature`)
- fido-u2f matching credential cert: **accepted**
- fido-u2f mismatched credential key: **refused** (`attestation`)
- tpm: **refused** with the format named
- malformed attStmt (non-map / missing x5c): **refused** (`malformed`)

```
node --experimental-strip-types --test src/lib/server/webauthn-verify.test.ts   # 31/31 pass
node --experimental-strip-types --import ./scripts/test-alias-register.mjs --test src/lib/server/passkey-store.test.ts src/lib/server/passkey-ceremony.test.ts  # 34/34 pass
pnpm typecheck  # clean
git diff --check  # clean
```

## Live iOS flow

The iOS app requests `attestation: "none"` today and produces `apple`/`none`; both empty/statement-less paths still pass unchanged, so existing registrations are not broken.

### DIFF
--- a/src/lib/server/passkey-ceremony.ts
+++ b/src/lib/server/passkey-ceremony.ts
@@ -215,6 +215,7 @@ export async function completeRegistration(input: {
     signCount: result.signCount,
     aaguid: base64UrlEncode(result.aaguid),
     attestationFormat: result.attestationFormat,
+    attestationVerified: result.attestationVerified,
     label: (input.label ?? "").trim().slice(0, 64) || "Passkey",
     createdAt: now,
     lastUsedAt: null,
--- a/src/lib/server/passkey-store.ts
+++ b/src/lib/server/passkey-store.ts
@@ -33,8 +33,15 @@ export type StoredCredential = {
   algorithm: number;
   signCount: number;
   aaguid: string;
-  /** Recorded, NOT verified — see cave-01v4u. */
+  /** Recorded verbatim, even when it was not cryptographically checked. */
   attestationFormat: string;
+  /**
+   * Whether the attestation statement was verified at registration (cave-01v4u):
+   * `true` means the certificate chain reached a pinned root; "none-equivalent"
+   * means the statement was self/absent and proves nothing about the model.
+   * Absent on credentials stored before cave-01v4u landed — additive schema.
+   */
+  attestationVerified?: boolean | "none-equivalent";
   label: string;
   createdAt: number;
   lastUsedAt: number | null;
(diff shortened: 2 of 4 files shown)


CompleteDotTech:
Superseded by #5168, which additionally implements the Apple anonymous-attestation nonce extension (OID 1.2.840.113635.100.8.2) and moves the client to attestation:'direct', so it closes the real gap this version only documents. Same account, same bead — keeping one implementation.

PR #5177, opened 2026-08-29
test(e2e): calendar rove spec respects the documented stay-put edge (UTC Saturdays)

CompleteDotTech:
## Summary

The month-grid roving hook stays put when a row-step would leave the grid (documented in `src/lib/use-roving-tabindex.ts`: "A row-step off the top/bottom edge stays put rather than clamping"), but `calendar-month-grid.spec.ts` always asserted `start + 8` after ArrowDown.

When the anchor (today's cell) is in the grid's last row — e.g. any Saturday in a Sunday-start grid, which is what UTC CI hits — ArrowRight moves into the final row and ArrowDown correctly stays put, making the old assertion (an index that does not exist) fail deterministically. Reproduced locally with `TZ=UTC`, passes with `TZ` unset.

## Fix

The spec now asserts the documented behavior: `start + 8` when a next row exists, the stay-put `start + 1` at the bottom edge. This is currently blocking E2E shard 1 for every PR that runs CI on a UTC Saturday.

## Verification

`TZ=UTC pnpm exec playwright test tests/calendar-month-grid.spec.ts --project=desktop` — 5 passed.

### DIFF
--- a/tests/calendar-month-grid.spec.ts
+++ b/tests/calendar-month-grid.spec.ts
@@ -74,7 +74,13 @@ test.describe("calendar month grid keyboard model", () => {
     expect(await focusedCellIndex(page)).toBe(start + 1);
 
     await page.keyboard.press("ArrowDown");
-    expect(await focusedCellIndex(page)).toBe(start + 8);
+    // A ↓ row-step from the grid's last row would leave the grid entirely;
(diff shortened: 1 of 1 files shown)


CompleteDotTech:
The calendar rove spec fix is already on main (applied via the merge flow) — this branch is now redundant. Closing.

PR #5210, opened 2026-08-30
Broadcast one message to multiple selected chats

BunsDev:
Adds broadcast: multi-select chats in the sessions list and send one message
to all of them. Each target keeps its own session and reply — a fan-out, not
a shared room. Concurrency capped at 4, deliver-and-detach (no waiting for
replies), works across the docked rail, full chat list, and mobile sheet.

Verification:
- `pnpm typecheck` — clean
- `pnpm lint` — clean
- `src/lib/chat-broadcast.test.ts` — new unit tests for the broadcast core
- `src/app/api/api-contracts.test.ts` — updated for the new route

Bead: cave-g7yg6

### DIFF
--- a/scripts/run-tests.mjs
+++ b/scripts/run-tests.mjs
@@ -673,6 +673,7 @@ export const SUITES = {
     "src/components/chat-start-from-bands.test.ts",
     "src/components/chat-start-from-deadspace.test.ts",
     "src/components/chat-inline-composer.test.ts",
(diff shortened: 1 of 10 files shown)


BunsDev:
Closing: this branch's content is already merged on main via PR #5130 (identical commit, same 10 files). Discovered post-hoc that my initial diff check compared against a stale local main ref rather than origin/main. No action needed.

PR #5346, opened 2026-09-09
fix(conformance): accept case-sensitive canonical lookup refusal

BunsDev:
## Summary
- treat a validated 404 for `/conversations/BRANCHED/messages` as passing canonical-ID boundary evidence
- preserve the existing 200-path requirement that returned messages carry the canonical `branched` ID
- reject malformed 404 envelopes and all other statuses

## Root cause
The protected Linux runner uses a case-sensitive filesystem, so the mixed-case lookup correctly returns 404. The Cave harness recorded that safe refusal as `skip`, while the frozen Chat schema-v2 evidence validator requires every one of the 110 Cave assertions to be `pass`. The server behavior was correct; the evidence classification was not.

## Validation
- `node --test scripts/client-v1-conformance.test.mjs`

## Integration
This unblocks the Linux leg recorded on OpenCoven/sdk#38. The exact fix commit must remain reachable from `main` because Chat will pin it as Cave authority; merge with a merge commit rather than squash.

### DIFF
--- a/scripts/client-v1-conformance.mjs
+++ b/scripts/client-v1-conformance.mjs
@@ -437,6 +437,21 @@ export function checkEnvelope(body, expectation) {
   return failures;
 }
 
+export function canonicalConversationIdFailures(response) {
+  if (response.status === 404) {
+    return checkEnvelope(response.json, { kind: "error", code: "not_found" });
+  }
+  if (response.status !== 200) {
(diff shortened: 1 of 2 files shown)


BunsDev:
Superseded by #5347. The original branch was based on current `main` and accepted a validated filesystem-dependent 404. #5347 instead backports the stronger non-case-only alias probe onto frozen Cave `0.3.12`, so every platform must return and prove the transcript-owned canonical ID without skips.

PR #5347, opened 2026-09-09
fix(conformance): make v0.3.12 canonical ID probe portable

BunsDev:
## Summary
- backport the portable canonical conversation-ID probe onto the exact frozen Cave `0.3.12` authority commit
- use a non-case-only filename alias so Linux, macOS, and Windows exercise the same behavior
- require one bounded message carrying the transcript-owned canonical ID
- fail missing aliases, server/rate-limit responses, wrong IDs/messages, empty or oversized pages, and cleanup failures

## Root cause
The previous mixed-case probe was filesystem-dependent. Linux correctly returned 404 and the harness recorded a skip, but the frozen SDK registry requires all 110 Cave assertions to pass. This probe removes filesystem casing from the assertion rather than weakening evidence acceptance.

## Validation
- `node --test scripts/client-v1-conformance.test.mjs`
- prior isolated compatibility release build and full socket authority run: 110 passed, 0 failed, 0 skipped

## Integration
The branch is based on `bc310e9753783678014086ed6f7ef7d3fb797967` and preserves release `0.3.12`. The exact feature commit must remain reachable from `main` because Chat will pin it as Cave authority; merge with a merge commit rather than squash.


BunsDev:
Superseded by #5348. The backport commit `d20d83c46ba0c32433ce8dc6a358fb14b6bd0e45` is preserved there as an explicit merge parent, while the PR diff applies the equivalent portable probe cleanly to current `main`.

PR #5354, opened 2026-09-09
fix(chat): harden automatic conversation naming and topic-shift gating

BunsDev:
Generated conversation titles now follow the visible conversation path and preserve manual naming. Revision/text checks prevent delayed auto-renames from overwriting newer names, and unchanged auto titles leave both state and the sessions cache untouched.

Title normalization removes hidden prompt/reasoning blocks and credential material before truncation. It handles nested inline markup, quoted tag attributes, complete HTTP authorization values, and grapheme boundaries while preserving ordinary colon headings. Periodic renaming requires a material topic change; Client v1 reads project the same stored title.

Verification: focused naming/config/auto-rename/first-turn/secret-redaction tests, typecheck, lint, and test-wiring checks pass. Fresh code review found no remaining actionable findings. All 1,429 app and 472 API test files pass. App coverage combines the passing prefix with a Bash rerun from the lifecycle fixture after /bin/sh emitted unsupported startup stderr.

Bead: cave-9vzp4


BunsDev:
Fixed in 72ff91903: generic inline HTML tags are now stripped (keeping their text) right after hidden-block stripping and before either redaction pass, so a secret split like `sk-<b>proj</b>-<suffix>` is rejoined before the token-pattern check runs. Markdown URL/email autolinks are unaffected — added regression coverage in cave-chat-titles.test.ts.

PR #5362, opened 2026-09-11
Require Board coverage and wire the continuity contract

BunsDev:
## Summary
Complete the review follow-up to #5361 under cave-2i0zj. Beads-only absence cannot clear a request that may already exist as a Board-only task: missing relevant Board/task coverage is partial or unknown, and no positive match means relationship unknown rather than permission to create competing work. Exact authorized positive matches remain useful without exhaustive discovery.

Add a built-in Node contract covering the synthetic corpus, required safety language, matching entrypoints, and runner wiring. Run it in both app tests and the docs CI lane. The contract explicitly does not evaluate model decisions; a separate read-only rehearsal covers all 15 cases. Clarify authorization in the stale-owner fixture.

## Evidence
Targeted continuity, docs, Beads, CI-path, scanner, and slash-resolver tests pass; all 2009 test files are wired. The 15-case rehearsal found no rubric mismatch or forbidden action.

No runtime automation, identity changes, or changes to Cody-owned cave-o1aw3. Runtime follow-up remains cave-7qn9r. User authorized the protected-main merge and scoped cleanup.

### DIFF
--- a/.agents/skills/work-continuity/SKILL.md
+++ b/.agents/skills/work-continuity/SKILL.md
@@ -38,6 +38,10 @@ one-turn answer.
 3. Read details and recent owner comments for candidates. When supported,
    search titles and descriptions separately. Record query limits and errors.
    A result at its limit is potentially truncated, not a negative result.
+   Beads-only absence does not cover Cave Board/task records. Include relevant
+   authorized Board/task sources or report coverage `partial` or `unknown`.
+   Without a positive match, missing relevant Board/task coverage means
+   relationship `unknown`, not `no-match-in-scope`; do not start competing work.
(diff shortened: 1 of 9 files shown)


BunsDev:
Addressed the final review notes in 80c74a96c1400e6ed819de385b92fda3d52b0d4d: standalone work-continuity skill files and root AGENTS.md/CLAUDE.md now select the lightweight documentation contracts, with per-path regression cases. Unrelated skill and nested-guide paths remain unchanged. The plan explicitly distinguishes its original 14-case rehearsal from the later 15-case follow-up. Targeted continuity, path-selection, and docs-index contracts pass.

PR #5367, opened 2026-09-12
Retire unused session initiator exports

BunsDev:
## Summary
- Make the internally used agent-label helper private and remove the unused label formatter and OpenClaw transcript helpers. Keep the live session-key API unchanged.
- Confirm abandonment: PR #1858 (7f2beed2e) removed the Calls surface and the production transcript ingestion caller. Current session-list merging preserves supplied provenance or uses a familiar/harness fallback; no current ingestion consumer remains.
- Replace implementation-detail tests with live API coverage for human/system channels, normalization, invalid channels, keyed/fallback agents, and labels. Correct the historical Telegram plan while preserving its future compatibility requirements.

Bead: cave-ep9fb

## Verification
- `node --experimental-strip-types --import ./scripts/test-alias-register.mjs --test src/lib/session-initiator.test.ts src/lib/session-list-merge.test.ts` (2 files passed)
- `pnpm exec eslint src/lib/session-initiator.ts src/lib/session-initiator.test.ts`
- `pnpm typecheck`
- `git diff --check`

### DIFF
--- a/docs/specs/2026-06-29-telegram-channel-migration-plan.md
+++ b/docs/specs/2026-06-29-telegram-channel-migration-plan.md
@@ -6,7 +6,7 @@
 
 **Architecture:** Treat OpenClaw's Telegram implementation as the compatibility reference, not the long-term runtime owner. First land an OpenCoven Telegram connector for safe outbound delivery; then add Cave docs/settings and a migration inventory; then move inbound/session handling only after outbound, auth, allowlists, and transcript compatibility are verified.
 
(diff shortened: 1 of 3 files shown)


BunsDev:
Addressed the review summary documentation nit in commit 8ba61c79a0f65ce8ce5f1e0f33d36504ae739783: the migration plan now distinguishes the daemon session-key fallback from the human Cave-user default for local conversations. No runtime behavior changed.

PR #5400, opened 2026-09-14
Refuse when the device-access policy cannot be read, instead of opening

### LINKED ISSUE #5398: Windows ACL probe returns empty stdout, killing packaged server boot
`v0.4.2-rc.2` (`62e0a4606`, [run 34839946087](https://github.com/OpenCoven/coven-cave/actions/runs/34839946087), job `103962493874`) failed the Windows runtime leg with a **different** error than the four candidates before it:

```
Error: packaged server exited before readiness
```

The previous four said `packaged server did not answer within 90000 ms`. The server now **dies** rather than hanging.

## #5395 worked — and that is what exposed this

The SID-native `GetAccessRules` path removed the account-name translation that was stalling the probe past the 90s readiness window. The probe no longer stalls. It now **fails fast** instead.

`client-v1-compatibility-control.integration.test.mjs` embeds the server's own output into the failure (`output().slice(-4_000)`), and inside it is the spawn error object from the ACL probe:

```
stdout: '',
stderr: ''
```

## Chain

1. ACL probe returns **empty** stdout and stderr.
2. `parseClientV1WindowsAclReport(stdout)` cannot parse it, so it throws.
3. `assertExclusivePathOwnership` propagates.
4. `server.ts` module-scope `await createDeviceAccessStore()` rejects, unhandled.
5. Process exits → `!isRunning(child)` → "exited before readiness".

## What #5396 does and does not do

PR #5396 decouples boot from this: the server survives and device access fails closed. That **unblocks the release**.

It does **not** fix this bug. With #5396 alone, Windows boots with device access **permanently disabled** and a warning — pairing, approvals and device credentials do not work on that platform.

## Unknown, and the reason it is unknown

Whether the probe **timed out** or failed some other way is **not established**. Empty stdout+stderr is the signature of a killed child, so `WINDOWS_ACL_PROBE_TIMEOUT_MS` (12s, `MAX_ATTEMPTS` 2) firing twice is a strong candidate — but it is a candidate, not a finding.

The error object's leading fields (`status`, `signal`, `killed`) were **cut off by the test's own truncation**. `output().slice(-4_000)` keeps the *last* 4000 characters, which preserved the tail of the embedded PowerShell script and discarded the diagnosis. The visible text starts mid-token (`ier('S-1-3-4')` — a severed `SecurityIdentifier`).

**Suggested first step:** stop the truncation destroying the evidence, so the next run says which failure this is.

## Acceptance

The Windows ACL probe either succeeds on the runner, or fails with a diagnosable reason that reaches the log untruncated; device access is functional on Windows rather than permanently disabled.

Related: #5395 (fixed the stall), #5396 (decouples boot), #5391 (the 12s bound), `cave-9jt60`.

---
Filed as a GitHub issue rather than a bead: the local `bd` binary is at schema v53 against a v66 database (13 migrations behind) and refuses every command, including reads. Migrate to a bead when `bd` is usable again.

BunsDev:
Fixes a security regression **I introduced in #5396** and described as safe. Found by the Copilot review on that PR, confirmed in the gateway source.

## The defect

#5396 made a failed device-access store answer `policy()` with `{ enabled: false }`, and called that fail-closed. It is not.

The gateway reads `enabled: false` as *"device access is **configured off**"* and answers it with **legacy mode** — for a non-loopback request it adds the response to `legacy` and returns `false`, passing the request **through to the app**, with no pairing and no allowed-tailnet check:

```ts
if (direct) return false;
if (!policy.enabled) {
  legacy.add(res);
  res.once("close", () => { legacy.delete(res); });
  return false;            // <- remote request proceeds, un-paired
}
```

So a store that failed to open did not refuse. It turned a failed security check into an open door, on exactly the remote path pairing exists to gate. Before #5396 the same failure killed the process — dead, but not open. I converted "dead" into "potentially open" and wrote "it grants nothing" in the PR body.

**Exposure:** requires the server reachable off-loopback (the Tailscale path) *and* device-access initialization failing. That second half is not hypothetical — it is happening on every Windows CI runner right now (#5398).

## The shape of it

One value carried two meanings. *"Configured off"* is a decision somebody made. *"Could not be read"* is the absence of an answer. They cannot share a representation, and `enabled: false` was being asked to mean both.

`DeviceAccessPolicy` now carries `unavailable`, and all three sites that read `enabled` handle it:

| site | before | after |
| --- | --- | --- |
| legacy fallthrough | passes remote request through | **503, refused** — the hole |
| `POST /requests` | `409 disabled` | `503 unavailable` — "not enabled" was a false account of why |
| `blocksUpgrade()` | `false` → upgrade allowed | `true` → blocked |

`DeviceAccessError` gains an `unavailable` code (503), distinct from `forbidden`, which is a check that ran and said no.

## Also: the policy wait is now bounded

The gateway awaits `policy()` on **every** request, *before* the direct-loopback bypass. So an unbounded wait there is the same outage as blocking boot, moved one layer down — the server listens and answers nothing. This was the review's other finding, and it means #5396 did not actually fix the original *stall*; it fixed the *fast failure* that #5395 turned the stall into.

A policy read now waits up to 5s for initialization, then answers `unavailable`: remote traffic refused, loopback unaffected.

## Why #5396's tests missed it

They asserted `policy()` returns `{ enabled: false }` and called it fail-closed. They pinned **my belief about** the gateway rather than checking the gateway, so they confirmed the defect instead of finding it.

The replacement drives a **real gateway** over a real HTTP server and asserts a remote request is refused **and never reaches the app**. Verified it fails against the pre-fix gateway and passes against this one — I neutralised the guard and watched it go red, rather than trusting that it would.

## Verification

- `gateway.test.ts` + `deferred.test.ts`: 13 tests, 13 pass.
- Guard-removal check: the new test fails without the fix.
- `tsc --noEmit` clean; `server.mjs` regenerated via `build:server`.

Does **not** fix #5398 (why the Windows probe returns empty). That remains open, and with it the fact that Windows currently boots with device access disabled.


BunsDev:
## This looks superseded — please don't merge it without a second look

I picked this up to resolve its conflicts against `main` and stopped, because resolving them would land a **weaker** design than `main` already has.

### The hole this PR describes is closed on `main`

The PR's account of the defect is right: `enabled: false` meant both *"configured off"* and *"could not be read"*, and the gateway answered the first with legacy fallthrough — `legacy.add(res); return false;` — passing a remote request through un-paired.

`main` fixed that, by a different route. `deferred.ts` no longer returns a policy it cannot vouch for; it **throws**:

```ts
const live = (): DeviceAccessStore => {
  if (ready) return ready;
  throw new DeviceAccessInitializationError(…);
};

async policy() {
  await settled;
  return live().policy();
}
```

So on a failed store, `await currentPolicy()` throws inside `handle()`'s `try`, and the catch answers `503 unavailable`. Execution never reaches `if (!policy.enabled)`, so the legacy fallthrough cannot be hit. `main`'s own docstring states the invariant:

> Policy reads also refuse: an unavailable policy is not permission to use legacy remote credentials. Remote access fails CLOSED while local features stay up.

### And the second half is fixed too

This PR also notes the gateway awaits `policy()` on *every* request, ahead of the direct-loopback bypass, so an unbounded wait there is the same outage moved one layer down. On `main` the bypass now runs **before** the await:

```ts
const direct = options.isDirectLoopback(req);
…
if (direct && !isApi && !isHandoff) return false;
try {
  const policy = await currentPolicy();
```

### Why merging anyway would be a regression, not a no-op

The two designs disagree about representation, and that is the whole substance of this PR.

- **This PR:** add `unavailable?: boolean` to `DeviceAccessPolicy`. Every reader must remember to check it. A future site that reads `enabled` and forgets `unavailable` silently reopens the exact hole.
- **`main`:** make the unanswerable case *unrepresentable* as a policy. A thrown `DeviceAccessInitializationError` cannot be mistaken for "configured off", by any reader, present or future.

The second is strictly stronger, and it is the one shipping. Merging this would reintroduce the sentinel field alongside it and hand the next reader two mechanisms for one invariant.

### Suggested disposition

Close as superseded, and keep the write-up — the analysis of how `enabled: false` came to carry two meanings is the most useful record of this bug anywhere, and it is worth linking from whatever landed the exception-based fix.

If any part is genuinely still missing on `main` I could not find it, and I would rather be corrected than have quietly resolved 13 conflicting paths across `server.mjs`, `deferred.ts`, `store.ts` and their tests to land it.

I have not modified the branch. The merge was aborted; its head is still `5a76dea97`.

BunsDev:
Closing as superseded, per the analysis in the comment above.

The security hole this PR describes — `enabled: false` meaning both "configured off" and "could not be read", with the gateway answering the first by passing a remote request through un-paired — is closed on `main` by the exception-based design in `deferred.ts`: `policy()` throws `DeviceAccessInitializationError` rather than returning a policy it cannot vouch for, so `handle()` refuses with 503 and the legacy fallthrough is unreachable. The bounded-wait concern is covered too; the direct-loopback bypass now runs ahead of the policy await.

Merging this would not have been a no-op. It would have reintroduced an `unavailable?: boolean` sentinel alongside a design that makes the unanswerable case unrepresentable, leaving two mechanisms guarding one invariant — and a sentinel is only as good as every future reader remembering to check it.

**The write-up stays worth reading.** The description here is the clearest account anywhere of how one boolean came to carry two meanings, and of the specific gateway path that turned a failed security check into an open door. Worth linking from any future change in this area.

The branch was not modified: the conflict resolution was aborted and its head remains `5a76dea97`.

PR #5411, opened 2026-09-14
Answer what is stopped on you, not what is running

BunsDev:
Design handoff `Coven Cave Prototype.dc.html` **frame 2c**, unblocking `cave-21rp`.

## Why this frame, and why now

`cave-21rp` has been blocked since 2026-07-20 with `⚠️ NO APPROVED DESIGN EXISTS` on its DESIGN field — the spec it cited was unrecoverable. This frame is that design, approved by the owner on 2026-09-14.

The `cave-ltl38.9` tension the ledger recorded **resolves rather than persists**: that decision was that background missions should stay *quiet*, and this inbox is the one surface that agrees with it. It drops `running` entirely.

## What changes

The retired control answered *"what is running?"* with a number that was never zero, so its badge stopped carrying information and a 156-row list stopped being a popover. This one answers *"what is stopped on me?"* — blocked → failed → awaiting, oldest wait first, and usually short.

It is built **on** phase 1 (#5407) rather than beside it: rows compose through `sessionLifecycle()` and order by `compareNeedsYou`, so the inbox cannot disagree with the thread rail about what a session's state is. No backend state is invented — every field comes from data `SessionRow` already carries.

## Three spec rules, pinned by tests

Each could be undone silently by a later edit, so each has a test:

- **The panel is opaque.** `.ui-popover` carries a backdrop blur *and* a translucent glass fill, so overriding `background` alone leaves the blur compositing the transcript underneath — which is precisely the handoff's third P0 finding. It is cancelled explicitly, and the rule is qualified with `.ui-popover` so it wins regardless of sheet order (this sheet is deliberately outside the globals facade).
- **Row tint is spent on awaiting and blocked.** Failed keeps a badge and an edge. A row-wide red field across every failure is the alarm wall #5407 retired.
- **The badge counts actionable work only, and is absent at zero.** The bell itself stays mounted — that is how the empty state and ⇧⌘A are reached.

## Deliberate calls

- **Running is still counted, across sources.** The one thing the retired popover knew that sessions alone do not is its cross-source running total, so that still feeds the footer (fetched only while the panel is open). Replacing the list did not quietly narrow *"12 running"* to *"running chats"*.
- **"Mark all seen" is keyed on session id + the instant the ask was raised**, so a new ask re-surfaces through an old dismissal. Per-device and deliberately unsynced: "I have looked at this" is a fact about one reader at one screen.
- **The wait hairline saturates at a week** and is drawn as a comparative cue, never a measurement — the exact wait is always written beside it.
- **Ritual and board work reaches the inbox only through a backing session**, because attention evidence is server-authored per session and nothing else carries it.
- **A generated run with no saved transcript is dropped**, because its row would open an empty reader.

## Gates

- `pnpm lint` (design ESLint gate, codemod check, token-reference scan) — clean. The two `banked at` lines the token scan prints are **pre-existing on clean `main`**, verified by running the scan there: identical output, same 32 undefined / 0 theme-scoped-only.
- `pnpm typecheck` — clean, including the new Playwright spec.
- `offScaleSpacingPx` is **unchanged**: the sheet's three off-scale literals were snapped to `--space-1`/`-2` rather than banked.
- `inlineTsxStyles` goes **+1**, justified in the baseline comment: the per-row wait hairline's width is that item's own wait as a fraction of a week — a per-item measurement that changes every minute tick, the same category the chat run-rail and Review Cockpit entries banked.

## Test changes worth a look

Two source-text tests pinned the old wiring; both **move their assertions rather than lose them**. `shell-chrome-revamp.test.ts` also splits its spanning regex in two — its `[\s\S]{0,N}` gap had to be widened whenever the prose between the two lines grew, which is a test that fails for the wrong reason.

## Not in this PR

The rest of frame 2b. The prototype's run rail is largely **already built** (`ChatActivityMap`, from the earlier `Coven Cave - Chat Session` handoff), and its remaining delta is almost exactly the PLAN / CONTEXT / COST panels a prior session *deliberately* declined because nothing measures them. Building those would be a surface that lies.


BunsDev:
Both review findings addressed in `80e3a6b38`.

**⇧⌘A (unresolved thread)** — the reviewer was right: the tooltip advertised it and nothing was bound. It is now bound in the workspace's global keydown handler (a shortcut has to work from surfaces this component is not mounted on), asked for by a window event so the popover keeps its own open state, and catalogued in `keyboard-shortcuts.ts` — whose header requires every entry to correspond to a real binding. This is the same defect the `⌘,` block two cases above it was written to fix.

**The row `aria-label`** — also right, and I had this backwards. An `aria-label` replaces a button's descendant accessible name, so labelling the row suppressed the relative wait from assistive tech while leaving it visible on screen — and the wait is the field the list is *ordered by*. The explicit label is removed; the name now composes from visible content as title · metadata · state · "4d waiting".

That restores @BunsDev's `9bf57ccb4`, which I had partly reverted after reading it as an accidental edit. It was not — it was a correct response to this review, which I had not read at that point. The same applies to the other two commits on this branch: `7c5308375` (`showFamiliarChatList`) and `d9e45841a` (resetting the cross-source total on open) both came from these threads, and both are kept.

Tests pin all of it: the shortcut must be advertised **and** bound **and** catalogued, and the row must carry no explicit label while still containing its wait text. E2E covers the shortcut opening the panel and the row announcing its wait.

PR #5465, opened 2026-09-17
Fill TERM for the Tailscale CLI and stop reading a CLI fault as a stopped tunnel

### LINKED ISSUE #5438: Packaged app never reconciles Tailscale Serve: CLI probe fails when TERM is unset
## Symptom

`POST /api/mobile-handoff {"action":"app-start"}` against the packaged app answers:

```json
{"ok":false,"error":"Tailscale is not running","stderr":"","steps":[…{"id":"tailscale","state":"fail","detail":"Open Tailscale and connect, then retry."}…]}
```

while `tailscale status` in a terminal on the same machine reports `BackendState: Running`.

Because the probe fails, the app never reconciles its Tailscale Serve route. Observed
2026-09-15: Serve was still publishing `/ -> 127.0.0.1:3000` with nothing listening there,
so every request from the paired iPhone got a 502. The app cannot self-heal, and the
checklist blames Tailscale and the phone for a fault in neither.

## Root cause

`/Applications/Tailscale.app/Contents/MacOS/Tailscale` requires a **non-empty `TERM`** in its
environment. Without one it writes to **stdout** and exits **0**:

```
The Tailscale GUI failed to start: The operation couldn’t be completed. (Tailscale.CLIError error 3.)
```

Reproduce:

```bash
env -i          /Applications/Tailscale.app/Contents/MacOS/Tailscale status --self --json   # GUI-failed line, exit 0
env -i TERM=dumb /Applications/Tailscale.app/Contents/MacOS/Tailscale status --self --json   # JSON, exit 0
```

`TERM` is the entire minimal requirement — bisected by removing one key at a time from a
working environment. `HOME`, `TMPDIR`, `PATH`, `XPC_*`, `SECURITYSESSIONID` are irrelevant.

A **GUI-launched** process has no `TERM`. The packaged sidecar
(`CovenCave.app/Contents/Resources/resources/server/server.mjs`, parent
`CovenCave.app/Contents/MacOS/app`, ppid 1) therefore fails **every** `runTailscaleCommand`
call. `tailscaleSpawnEnv()` (`src/lib/mobile-handoff.ts`) repairs `PATH` for exactly this
GUI-launch case but not `TERM`.

The exit code is 0, so `classifyTailscaleSelf` takes the parse path, fails to parse, and
falls through to `not-running` with **empty stderr** — which is why the response carries no
usable evidence and the fault reads as a Tailscale problem.

## Proposed fix

1. `tailscaleSpawnEnv()`: default `TERM` to `dumb` when unset or empty, alongside the existing
   `PATH` repair.
2. `classifyTailscaleSelf`: distinguish exit-0-with-unparseable-stdout from a genuinely stopped
   backend, and surface the CLI's stdout as the detail instead of the generic
   "Open Tailscale and connect, then retry."

Both are covered by unit tests in `src/lib/mobile-handoff.test.ts`.

## Notes

- Only reachable from a GUI launch, so no CI job hits it (no macOS lane).
- Workaround applied by hand on the affected machine: `tailscale serve --bg http://127.0.0.1:3020`.

BunsDev:
## Why

#5438: the packaged app never reconciled Tailscale Serve. A GUI-launched sidecar has no `TERM`, and the macOS Tailscale.app CLI then answers every probe with "The Tailscale GUI failed to start" on stdout and exit 0. The classifier read that as a stopped tunnel, so the phone got 502s while the same command worked from a terminal.

## What

- `tailscaleSpawnEnv` guarantees `TERM` (a real terminal keeps its own value; a missing or blank one becomes `dumb`).
- `classifyTailscaleSelf` returns a new `cli-unusable` kind for exit-0 output that is not status JSON, carrying the CLI's own first line in the detail instead of collapsing it to `not-running`.
- The mobile-handoff route reports "Tailscale CLI could not report status" for that kind.

## Validation

- `node --test src/lib/mobile-handoff.test.ts`: pass
- `pnpm typecheck`, `pnpm lint:source`: clean

The uncommitted change was found preserved in `.worktrees/5438-tailscale-cli-term-env` with no live owner and no PR; this publishes it unchanged.

Closes #5438


BunsDev:
## Verified the two commits pushed onto this branch, 2026-09-17

`ac92a47f1` is a real improvement over what I wrote. My `classifyTailscaleSelf` only returned `cli-unusable` when `JSON.parse` **threw**. Valid JSON with no string `BackendState` (`{}`, `[]`, `{"BackendState":123}`) still fell through to `not-running` — the same wrong answer this PR exists to remove, just via a narrower door. The commit closes that and adds `cli-unusable` to `TailscaleFailureKind` so client recovery classifies it explicitly and keeps it non-recoverable.

Checked at `ac92a47f1`, all green locally:

- `mobile-handoff.test.ts`, `tailscale-failure.test.ts` (5), `tailscale-recovery.test.ts` (6)
- `pnpm typecheck`, `pnpm lint:source`
- `server-pty-ws.test.ts` (the bundled-artifact check)

**On the bundled artifact:** `server.mjs` is already current at this head and needs no regeneration. I rebuilt with `pnpm build:server` and compared — byte-identical. `classifyTailscaleSelf` is tree-shaken out of the server bundle, unlike the `tailscaleSpawnEnv` change in `6ff54a104`, which is why that one needed `609372361` and this one does not.

## Why this PR is BLOCKED

Not a failure. The `CI` run on this head is `completed/action_required` — parked behind the manual-approval gate because the push came from the Copilot bot. Only CodeQL default setup ran, so the required `Frontend build` context is absent and will stay absent until the run is approved or a fresh one is dispatched. `pnpm ci:recovery` agrees: `#5465 approval_gated_run eligible`.

That needs operator authorization, so I have not done it.

PR #5483, opened 2026-09-20
test(research): reproduce cross-process lost update deterministically

### LINKED ISSUE #5443: Lost update in cross-process research-links save/save, seen once in CI
## What happened

`src/lib/server/research-links-compatibility-crash.test.ts` failed in CI on PR #5417:

```
✖ child processes serialize save/save without losing either update (444.482137ms)
  AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
  + actual - expected
    [
  -   'https://example.com/save-a',
        'https://example.com/save-b'
    ]
```

Run: https://github.com/OpenCoven/coven-cave/actions/runs/34931923752 (job 104261807961), head `fa291017c`, 2026-09-15.

`save-a` was lost. The test spawns two child processes that save concurrently and asserts neither update is dropped, so this is the exact failure the test exists to catch.

## Why it is filed rather than fixed

It has not reproduced. A clean re-run at a fresh head passed, and the file passes 3/3 locally. PR #5417 touched only `scripts/ci-recovery.*` and cannot have caused it.

**It is filed because "passed on re-run" is not a diagnosis.** A lost update under contention is precisely the class of bug that a green re-run hides: the serialization either holds under load or it does not, and one observed loss is evidence it does not, on some scheduling. Treating it as a flake and moving on is how this becomes a data-loss report later.

Note the failure mode is also silent in production terms — nothing errors, one write simply disappears.

## Where to look

The legacy store writes via the standard create-exclusive-then-rename pattern in `src/lib/server/research-links-legacy-store.ts`:

- `open(..., O_WRONLY | O_CREAT | O_EXCL | noFollowFlag())` around line 543
- `rename(temporary, target)` around line 557

`rename` is atomic, so the file is never torn. That does not make the *read-modify-write* atomic: if both children read the same base state before either renames, the second rename replaces the first writer's copy wholesale and its entry is gone. Whether a lock is held across the whole read-modify-write cycle, rather than just the write, is the first thing to check.

A CI runner under load widens exactly that window, which fits the evidence: it failed on a shared runner and has never failed on an idle developer machine.

## Reproducing

Loading the machine while looping the single test is more likely to surface it than repeating it on an idle one:

```bash
node --experimental-strip-types --import ./scripts/test-alias-register.mjs \
  src/lib/server/research-links-compatibility-crash.test.ts
```

## Acceptance

Either the read-modify-write cycle is shown to be serialized across processes and the CI failure explained some other way, or the cycle is made atomic. Closing this because it stopped reproducing is not an outcome.

BunsDev:
## Summary
Add a controlled two-process regression for the research-link lost update in #5443. It exercises the real compatibility transactions and checks the raw legacy projection before a reconciliation read can repair and mask the failure.

## Why
The missing choosing phase fixed by #5452 allowed a process to publish an earlier ticket after another process had entered. The new test reproduces a lost row with the pre-fix lock and passes with the merged implementation. Closes #5443.

## Changes
- Pause A between ticket selection and exclusive intent publication; let B either enter or demonstrate a second admission scan.
- Coordinate transaction reads/writes over IPC without a fixed sleep deciding the schedule.
- Assert both raw projected rows survive and B observes A's committed input.
- Keep interception inside disposable child processes and kill/reap them on test timeout or failure.

## Verification
- Regression against pre-#5452 lock: failed with save-b missing from raw projection; production file restored immediately afterward.
- Current implementation: all 47 lock/compatibility/crash tests passed; after cleanup review, all 13 crash tests passed again.
- Focused ESLint, TypeScript check, 2,038-file test wiring, diff check and pre-commit hook passed.
- Read-only review identified timeout cleanup; corrected and re-reviewed with no remaining findings.

## Risk + rollout notes
Test-only; no production changes. This reproduces the historical failure class under a controlled schedule, not the exact unrecorded scheduler trace of CI run 34931923752. A compatibility read can recover both rows from manifests, so the regression deliberately asserts the raw projection first. #5452 provides the production fix; #5478 separately hardened publication-failure cleanup.

Worktree intentionally retained by Cody at .worktrees/issue-5398-windows-device-diagnostics for the remaining issue queue. Branch is pushed; no foreign units changed.


BunsDev:
Reviewed the IPC waiter concern against d40465c6. Parent messages are appended to a persistent buffer before notify(); until() checks that buffer and installs notify synchronously inside the Promise executor, with no await or event-loop yield between those operations. Each child has only one active until() caller. Child commands likewise enter a pending Set if their waiter is not installed yet. Thus an early event is retained, and the proposed lost-notification interleaving cannot occur in this single-threaded path. No code change needed for this finding. All applicable exact-head CI checks passed.

PR #5484, opened 2026-09-20
fix(arcade): target visible wisps in the browser regression

### LINKED ISSUE #5413: Flaky: glitter-crypt-chromium.test.ts fails intermittently in the app suite
`src/lib/arcade/glitter-crypt-chromium.test.ts` fails intermittently in `pnpm test:app` and passes on re-run.

## Evidence

Observed on `feat/cave-21rp-needs-you-inbox` (now merged as `2ed1ed67d`) during an otherwise-clean run:

```
AssertionError [ERR_ASSERTION]: aiming at a wisp and zapping it raises the hexed count
    at src/lib/arcade/glitter-crypt-chromium.test.ts:139:12
  actual: false
  expected: true

✗ FAILED: src/lib/arcade/glitter-crypt-chromium.test.ts  (810 passed before it)
```

Re-run **3/3 PASS** in isolation immediately afterwards, with no code change. An earlier full run of the same suite on the same branch also passed it.

## Why it is a flake and not a regression

- The branch touched none of its dependencies — the file references neither `workspace` nor anything in the changed set.
- It passed in an earlier full run and in three consecutive isolated runs.
- It failed once, mid-suite, after 810 files.

The mid-suite-only failure suggests cross-test interference or a timing assumption rather than a defect in the assertion itself — worth checking whether the wisp/zap state is shared or clock-dependent across the chromium-backed suites.

## Cost

A flake here fails the whole `Frontend validation (app tests)` job, because the runner stops at the first failure. That makes it indistinguishable at a glance from a real regression in whatever branch happens to be running, which is how it cost time on this one.

BunsDev:
## Summary
Fix the arcade browser controller targeting a nearer wisp behind a wall while a shootable wisp is available. Closes #5413.

The read-only snapshot accepts optional visible-only selection using the same wall-occlusion predicate as fire(). Default snapshot behavior and gameplay are unchanged. The browser test uses visible targets and includes failure diagnostics.

## Verification
- Chromium: normal random gameplay and seeded gameplay with 1,500 ms RAF delay during aiming both passed, including the unchanged short-tap firing assertion.
- Red control: removing the visibility filter fails the aiming assertion with score 0 and an occluded target; restoring it passes both scenarios.
- Existing 60-second aiming deadline retained; no simulated hits or collision bypass.
- Pure arcade test, focused ESLint, TypeScript check, diff check and pre-commit hook passed.
- Independent read-only review: no actionable findings.

The controlled schedule reproduces the failure mechanism, not the exact unrecorded historical CI schedule. The delayed scenario restores normal RAF and drains its outstanding callback before testing short taps.

## Worktree disposition
Cody intentionally retains .worktrees/issue-5398-windows-device-diagnostics for this PR and the remaining issue queue. Branch pushed; no foreign worktree changed.


BunsDev:
Addressed both review-summary points in 0c1c05e9e. The old seed prefix was confirmed to produce document.compatMode=BackCompat. The seed now sits inside <head>, preserving the initial doctype, and both normal/delayed scenarios assert CSS1Compat. Added same-evaluation assertions for omitted versus false snapshot option, nearest-target ordering, and unchanged living-enemy count. Both full Chromium scenarios, focused ESLint, diff check and pre-commit hook passed. The 60-second aiming deadline and real firing path remain unchanged.

Issue #5256, opened 2026-08-31
P1: Add a real-daemon Threads Playwright acceptance project


BunsDev:
Cross-repository epic: OpenCoven/coven-threads#31  
Daemon prerequisite: OpenCoven/coven#884

## Outcome

Add a focused `threads-live-daemon` Playwright project that launches/attaches to the real isolated Coven daemon fixture and proves Cave renders and forwards authority state without becoming a policy engine.

Existing mocked Threads UI tests remain useful component contracts; they are not end-to-end authority evidence.

## Required topology

```text
Playwright browser
  -> Cave application
  -> real isolated Coven daemon
  -> current Threads validator
  -> ephemeral pending / SQLite / familiar workspace
```

Cave must consume daemon-emitted variants, labels, deadlines, tension, stale/degraded state, and terminal outcomes. It must not synthesize approval authority, infer an optimistic terminal state, or reinterpret unknown variants.

### 1. Pending window -> applied

- submit a proposal through the supported UI path;
- verify pending state remains visible for the minimum interval;
- verify approval controls reflect the daemon's typed path;
- advance deterministic daemon time;
- verify the UI displays the applied terminal result only after daemon confirmation.

### 2. Veto and evidence divergence

- open a real veto window;
- exercise principal veto through the UI;
- verify exactly one `vetoed` terminal state;
- separately change committed evidence before deadline;
- verify Cave renders `evidence_diverged`, blocked/non-applied, and no stale approval affordance.

### 3. Disconnect/restart/stale authority

- open a pending proposal;
- disconnect or stop the daemon;
- verify Cave marks state stale/degraded and disables mutation/approval actions;
- restart the daemon and recover the proposal;
- verify the UI reconciles to the single authoritative terminal result without duplicate toast/action/state.

## Rendering invariants

- unknown or stale authority state renders blocked;
- no optimistic apply, queued approval, or client-generated terminal outcome;
- human approval/rationale requirements come from the daemon wire contract;
- protected values are not echoed in failure surfaces;
- reconnect never replays an approval without the daemon's current authority binding;
- accessibility names expose status, deadline, reason, and disabled cause.

## Test isolation

- synthetic familiar/principal/content fixtures only;
- unique `COVEN_HOME`, socket/port, DB, and workspace per worker;
- deterministic daemon clock exposed by the Coven test fixture;
- Chromium first; Firefox/WebKit scheduled until stability threshold is met;
[...]

PR #5498, opened 2026-09-20
Add guarded daemon edit submission to Proposals

BunsDev:
## Summary
Add a dedicated “Send an edit” form in Proposals that forwards one complete file replacement to the supported daemon familiar-edits endpoint. The UI discloses immediate application and renders only confirmed applied, staged, held or refused outcomes; uncertain responses remain blocked across reloads until explicit reconciliation.

## Why
Prerequisite for the supported UI submission journey in #5256. The full real-daemon journey suite and 30-day stability criteria remain outstanding.

## Changes
- Bound and validate the three-field payload behind the existing local-origin guard; forbid fixture forwarding and transport replay.
- Persist only a content-free uncertainty marker before sending, with single-flight submission and explicit recovery.
- Use shared modal, buttons and announcements, gated on fresh verified daemon queue state; wrap header actions in narrow panes.
- Classify documented size/quota refusals only with explicit daemon no-write evidence.
- Wire 15 new boundary/controller/route tests and remove a stale unused token allowance to restore the token-reference ratchet.

## Verification
- Focused boundary/controller/route, proposal-flow, and design-token tests passed. All 330 API route contracts pass after registering the guarded submission route.
- Typecheck, focused ESLint, defined-token scan and test wiring passed.
- Synthetic rendered browser checks passed for duplicate submission, pending Escape, unknown result across reload, explicit recovery, applied confirmation, stale blocking, focus trap/return and 390px layout.
- Inspected dark, light and Tide themes; reproduced and fixed narrow-header overlap, with geometry checks at 320/390/768/1280px. Independent review found no blockers.

## Risk + rollout notes
This is a real file mutation path: the daemon may apply immediately or stage/hold under its existing policy. No production daemon edits were issued during verification. Browser checks used synthetic responses and do not establish real-daemon acceptance or human accessibility acceptance. Ordinary memory-editor saves retain their existing semantics.


BunsDev:
Addressed the review overview concerns in 802baa373fba2d0eb797e1d9bcd488beb7726510. The narrow-header overlap was reproduced at 390px and corrected with container-based wrapping; geometry checks pass at 320, 390, 768 and 1280px. The documented ward_apply_too_large and proposal_quota_exceeded responses map to refused only with HTTP 413 and explicit details.writeApplied:false; missing or contradictory evidence remains unknown. Five outcome tests and independent review cover this boundary. The API contract inventory now includes the guarded route (330 contracts pass). Updated-head CI remains in progress.

